PR #15, opened 2018-07-31
KVM Remote desktop working in ES6

ricardopolo:
This PR to use the scripts for KVM, using CIRA, to work in ES6 so you cant import them easily in another projects using
````js
import { CreateAmtRemoteDesktop } from '../../assets/scripts/amt/amt-desktop-0.0.2'
import { CreateAmtRedirect } from '../../assets/scripts/amt/amt-redir-ws-0.1.0'

CreateAmtRedirect(CreateAmtRemoteDesktop('desktop-canvas'))
````

### DIFF
--- a/public/scripts/amt-0.2.0.js
+++ b/public/scripts/amt-0.2.0.js
@@ -1,3 +1,5 @@
+/* eslint-disable */
+
 /**
 * @fileoverview Intel(r) AMT Communication StackXX
 * @author Ylian Saint-Hilaire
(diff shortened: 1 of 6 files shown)


ricardopolo:
@jsastriawan `node meshcentral.js` runs whitout any error.

And because the exposed methods continue working whitout any modifications, I think this should not broke anything. I think the purposed `export` will only be used if need using and `import` from Vue or React, otherwise it will be ignored.

Can you help me making futhers tests?

ricardopolo:
@Ylianst any comment?

Ylianst:
Well, I tried adding the exports/imports and browser don't like it. I see how it can be useful in a different context, but I can't fix it here without the browsers breaking. If there is a way about this, let me know.

luckydonald:
Probably a transpiler like babel or webpack (using babel) could get rid (or even translate) incompatible code.

Ylianst:
I am going to close this request as I can't see how this can work with the existing code base. However, if this something that is really desirable, I am open to suggestions.

ricardopolo:
Would be great that the kvm and the amtstack be npm modules tha can be used easily from multiple projects

Issue #126, opened 2019-03-11
Google Chrome complains about certificates saying NET::ERR_CERT_COMMON_NAME_INVALID


rzteux:
Hi,

I'm currently evaluating meshcentral2 which seems very promising to replace cost effective tools like teamviewer or anydesk solutions :)

I'm currently able to install it and run it behind my personal set top box :)

Unfortunately, this is not completely perfect :p

Under Google Chrome, generated certificate seems to not fulfill latest X.509 RFC that stipulate that SAN field (Subject Alternate Name) should be set with at least content of CN ( Common Name). Actually, this field seems to be empty.

Settings it during certificate generation should fix the annoying "confidentiality security warning in google chrome".

Regards,
Stéphane RZETELNY.

rzteux:
Well,

I have take a look in your code and I guess I have found the problem :)

Apparently you have tried to setup subjectAlternateName but it is not sufficient to fulfill Google Chrome security checks.
After reading RFC, I think you should, in addition to URI type SAN like "http://", add also equivalent DNS ones without http:// 
  https://github.com/Ylianst/MeshCentral/blob/fd523ae75ba0d9e5d95cf6141cbda81cd15c2278/certoperations.js#L157

So that in final certificate you should have following subjectAltNames:
- URL=http://<FQDN or IP@>
- URL=http://localhost
- DNS=<FQDN or IP@>
- DNS=localhost

What do you think about that ?

Regards,
Stéphaen RZETELNY.

Ylianst:
Hi. Sure, I can add that it a release later today. If it solves that error, it would be great.

Ylianst:
Made this change in MeshCentral v0.3.0-j. Let me know if it changes anything for Chrome.

rzteux:
Hi,
I have just finished to test latest version and unfortunately problem is still here with Google Chrome.
I had tested it also with Edge and there is same behavior.

So, I decided to fork your repository to fix it by myself :)

So I have added a new pull request #133 to let you review and certainly merge my changes about https certificate generation ;)

Hope to see new release soon :p

Regards,
Stéphane.

Ylianst:
Wow, thanks. I just took a quick look at the code and it looks perfect. I also did not know about the IsIP() method, that could be handy. As soon as I get back to the office, I will accept the code change.

Ylianst:
Published MeshCentral v0.3.0-o with these certificate generation changes. Please test and confirm it works. Thanks.

[...]

PR #133, opened 2019-03-16
Fix annoying Google Chrome warning about invalid certificate

rzteux:
add subjectAltName with DNS and IP entry type when necessary according to passed commonName in "IssueWebServerCertificate" function


rzteux:
ping back the original issue #126

PR #159, opened 2019-03-30
Moving style definition from login.handlebars to styles.css

elastalink:
We started with moving styles from login.handlebars to CSS:
- Style definitions on HTML elements are moved to styles.css
- Added classes on body element and CSS selector to address states (IE fullscreen)
- Removed JS code which was changing styles on state changes (except display: none style used to hide/show) elements
- Added few new ID's to elements with style definition and without ID's, before moving to CSS
- Footer is defined using div (like a header) instead of using HTML table
- Improved page responsiveness on fullscreen state (hide the welcome image, center login form, etc.)

We proposed changing page title based on {{title}} instead of using static "Meshcentral" title. This is useful when working with multiple "Meshcentral" sites at the same time.


Ylianst:
Wow! This looks great. I am on vacation until Wednesday, but just doing a quick look at the code, it looks really nice. I like the new "QC" function!

elastalink:
> Wow! This looks great. I am on vacation until Wednesday, but just doing a quick look at the code, it looks really nice. I like the new "QC" function!

Thanks :) Do you want us to continue working on other templates (while you are enjoying well-deserved vacation)?

Ylianst:
Please continue, this is wonderful! I feel a little bad being on vacation just as you submit a pull request. Anyone helping out must get top priority. I like any replacement of JavaScript with CSS, that is the right way to do it and it works so much better and a lot smoother. I did try to replace the default devices view with CSS, but should not get it to work right. In that screen, re-sizing is not smooth. Anyway, given how sensitive the project is to security, I need to do a security review on any pull request, but it will not be very difficult in this case. I expect to accept as-is. I am in Los Angeles now, I get back to Portland Wednesday during the day. Will be in the office Thursday.

elastalink:
Great, thanks. We are almost done with default.handlebars. Sending PR shortly...

Ylianst:
After testing, I had to make a few fixes to make the changes work. The left bar was hidden in the main site, hidden welcome image and gap at the bottom of the login page. Doing the same changes for the mobile site would be nice at some point. Thanks for doing this.

PR #245, opened 2019-05-24
MeshCMD - Added provisioning to ACM flow and unconfiguring from ACM flows

matt-primrose:
Added provisioning to ACM flow and unconfiguring from ACM flows to MeshCMD.  Provisioning to ACM requires amt-rcs service

### DIFF
--- a/agents/meshcmd.js
+++ b/agents/meshcmd.js
@@ -95,7 +95,7 @@ function run(argv) {
     //console.log('addedModules = ' + JSON.stringify(addedModules));
     var actionpath = 'meshaction.txt';
     if (args.actionfile != null) { actionpath = args.actionfile; }
-    var actions = ['HELP', 'ROUTE', 'MICROLMS', 'AMTLOADWEBAPP', 'AMTLOADSMALLWEBAPP', 'AMTLOADLARGEWEBAPP', 'AMTCLEARWEBAPP', 'AMTSTORAGESTATE', 'AMTINFO', 'AMTINFODEBUG', 'AMTVERSIONS', 'AMTHASHES', 'AMTSAVESTATE', 'AMTSCRIPT', 'AMTUUID', 'AMTCCM', 'AMTDEACTIVATE', 'SMBIOS', 'RAWSMBIOS', 'MESHCOMMANDER', 'AMTAUDITLOG', 'AMTPRESENCE'];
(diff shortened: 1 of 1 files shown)


matt-primrose:
Here is a link to the sample RCS server: https://github.com/matt-primrose/amt-rcs

Ylianst:
Excellent improvement. Adding support for Intel AMT Admin Control Mode (ACM) to MeshCMD is going to be really useful.

Ylianst:
Question: The help for the AMTACM command reads:

Intel AMT must be in "pre-provisioning" state for this command to work and a administrator password and provisioning certificate must be provided.

Should I read this correctly that the administrator password and provisioning certificate are provided by the activation server?

Also, the option --wss takes in a servername:port, however, it would be better if it accepted a URL. For example: wss://servername:port/customer1/activateamt. This is because we may have a server with many domains (co-hosting different customers).

FYI. I am going to fix the help a little, replace "amtconfig.js" with "meshcmd".

Thanks,
Ylian

matt-primrose:
Regarding the where the administrator password and provisioning certificate are located, yes, these are specified as part of the activation server configuration (currently).  Once the .mescript running is implemented, the administrator password will be part of the .mescript used to configure AMT after ACM activation.

Regarding -wss input, good point.  I can add that in the next update.

Regarding the little items.  Thanks, thought I caught all of those, but may have missed a few.

Thanks,
Matt

matt-primrose:
Followup on the --wss input.  Shouldn't the URL be dependent on the web server implementation?  So, if MeshCentral.com supports wss://servername:port/customer1/activateamt then the websocket server would handle routing to the correct tenant?  My simple websocket server I created for amt-rcs just passes all connections directly through, however, I could see the desire to have multiple amt-rcs instances running for each customer for security reasons.  A more intelligent websocket server could then understand how to route the connection to the correct amt-rcs instance.

spyram:
HI,

I'm trying to use AMTACM command but it ends up with a CCM activation instead :(

"Intel AMT CCM Activation success."

I'm using a custom generated certificate for server side AMT service.

The clients are only setup with a custom admin password,

The group is setup like "Simple Admin Control Mode (ACM)"
I running the command like Meshcentral "Activation" button instructs 
meshcmd amtacm --url wss://<myserver.fqdn>/amtactivate?id=<theserverId> --serverhttpshash <theserverhash>

but I'm getting this on amtactivation.log...
{"time":"2019-09-30T16:21:20.518Z","action":"**ccm**activate","domain":"","amtUuid":"<hwguid>","amtRealm":"Digest:2B300000000000000000000000000000","user":"admin","password":"<Ourpass>","ipport":"<clientIP>:62718","meshid":"mesh//<base64Hash>"}
{"time":"2019-09-30T16:21:20.999Z","action":"**ccm**activate-success","domain":"","amtUuid":"<hwguid>","ipport":"<ClientIP>:62718","meshid":"mesh//<base64Hash>"}

Any ideas?

BTW, you speak of amt-rcs. I'm running the latest MeshCentral2 (0.4.1-m). Is it builtin or do I need to setup Intel RCS or any other sw?

MailYouLater:
@spyram I recommend you open a [new issue](https://github.com/Ylianst/MeshCentral/issues/new) instead of trying to continue a conversation that hasn't had a comment in 4 months on a closed one. You're more likely to get the attention of the people who might be able to help that way.

spyram:
Thanks for the advise. I will :)

Please just let me know if amt-rcs is something we have to install or if its builtin Meshcentral.

Thanks.

PR #255, opened 2019-05-29
Update to meshcmd.js to add RCS message protocol

matt-primrose:
Version 1 of RCS message protocol implemented.  Dependent on RCS from: https://github.com/matt-primrose/amt-rcs


Ylianst:
Nice. When it's ready, I want to take a look at adding the RCS server to MeshCentral.

PR #338, opened 2019-07-05
Rename readme.txt to readme.md and improve markdown

MailYouLater:
Github doesn't render markdown in txt files, so readme.txt should be renamed to readme.md so that it will.

This PR also makes some improvements to the markdown present in the readme.


Ylianst:
Wow! Thank you! This is an excellent improvement.

PR #461, opened 2019-08-29
Inital commit for APF over secure websocket

jsastriawan:
Ylian,

This is initial version of APF over Websocket support. I will send test script separately via email.


Ylianst:
Merged. Yes, please send the test scripts. In taking a look at this, I may take a look at making the "mpsserver.js" to handle websocket connections instead of having a separate apfserver.js. This said, you feedback on this is welcome.

PR #491, opened 2019-09-17
Adding console apf command

jsastriawan:
This patch is to add apfclient at meshagent and to add console command to enable apf. Currently apf command is to check the status of apf tunnel, apf on to activate apf tunnel and apf off to terminate apf tunnel.

Please review.

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -306,6 +306,7 @@ function createMeshCore(agent)
     var amtscanner = null;
     var nextTunnelIndex = 1;
     var amtPolicy = null;
+    var apftunnel = null;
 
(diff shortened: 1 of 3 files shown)


MailYouLater:
What is "apf"?

jsastriawan:
APF: AMT Port Forwarding protocol. This is a tunnelling protocol to send Intel AMT command to the system remotely. The same protocol that power AMT CIRA (see mpsserver.js) but this version runs on top of websocket.

I still have to add server-side AMT JavaScript stack to demonstrate the capabilities.

MailYouLater:
Oh, that's not what came up when I searched.  Interesting.

PR #530, opened 2019-10-04
Added MQTT support over WSS and multiplexed with MPS

jsastriawan:
MQTT over MPS port and WSS. Test code will be sent via email. It can also be debugged with MQTT Exporer too http://mqtt-explorer.com/


Ylianst:
Thanks. I will be making a bunch of change to make this feature optional. However, this offers interesting possibilities going forward.

PR #601, opened 2019-10-24
Expand pluginHandler support for plugins

ryanblenis:
Expand pluginHandler support to communicate with browser<->server over the control channel. (Aids in gathering server-side plugin specific data).
Expand plugin functionality on agents to get plugin commands as their own action. (Aids in sending data to an agent from the server without abusing existing console msg's).


Ylianst:
Accepted, it will be in the new published NPM version tomorrow.

PR #610, opened 2019-10-27
Expand PluginHandler to Support Periodic Updates

ryanblenis:
Update to enable plugins to engage with agents via the sendPeriodicServerUpdate call and participate in periodic data exchanges.


Ylianst:
Looks good. Thanks.

PR #612, opened 2019-10-28
Revert "Expand PluginHandler to Support Periodic Updates"

Ylianst:
Reverts Ylianst/MeshCentral#610


Ylianst:
After looking into this, I can't accept this request. The contract between the agent and server is that the websocket stays connected and only if state the changes on one side is a message sent. Agent can't be polling for server changes. There are some people using MeshCentral over 4G/5G and added idle traffic is really bad, also adds hosting costs and reduces server scaling. On top of all this, people that don't use plugin's will have to deal with the added polling traffic which is really bad.

ryanblenis:
Ah, sorry about that. I thought the ~5kb/day/agent would be negligible. I'll come up with another way. Thank you.

PR #614, opened 2019-10-28
Expand Plugin Interactions

ryanblenis:
- Introduce plugin hook to verified agents with non-recovery cores
- Introduce plugin hook to agent message data, should the plugin need to interact (without any additional bandwidth used for installs without plugins enabled)


Ylianst:
Looks good.

Ylianst:
Hey! If you can comment on your thinking for #616, that would be wonderful. Thanks.

PR #636, opened 2019-11-06
Clean up module install code

### LINKED ISSUE #631: Greenlock install broken
I get the following message:
image
It seems like a timeout of 10 seconds is set, which is WAYYYY too short for npm, especially on low power devices like Raspberry PIs  (I suggest using a timeout of 120 seconds). https://github.com/Ylianst/MeshCentral/blob/338e8eb99e2e027a70bb9843bf9149d4f4872a01/meshcentral.js#L1956
Additionally, MeshCentral doesn't seem to care about the fact that I manually installed the specified version already. (Seems to be buggy because of usage of @ in the package name indicating the version. Works fine for other packages.)

TotallyNotElite:
Did some clean up on the module install code. 

(Merging this PR will auto close #631)

(psst @Ylianst, I wouldn't mind you marking some issues as `good first issue` :+1: )

### DIFF
--- a/meshcentral.js
+++ b/meshcentral.js
@@ -1932,8 +1932,10 @@ function InstallModules(modules, func) {
     var missingModules = [];
     if (modules.length > 0) {
         for (var i in modules) {
+            // Modules may contain a version tag (foobar@1.0.0), remove it so the module can be found using require
+            var moduleName = modules[i].split("@", 1)[0];
             try {
-                var xxmodule = require(modules[i]);
+                require(moduleName);
             } catch (e) {
                 if (previouslyInstalledModules[modules[i]] !== true) { missingModules.push(modules[i]); }
             }
@@ -1943,7 +1945,6 @@ function InstallModules(modules, func) {
 }
 
 // Check if a module is present and install it if missing
-var InstallModuleChildProcess = null;
 function InstallModule(modulename, func, tag1, tag2) {
     console.log('Installing ' + modulename + '...');
     var child_process = require('child_process');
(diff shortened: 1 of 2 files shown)


MailYouLater:
Careful, currently this PR changes the version number back to one that has already been used.  Please add another commit that fixes this, then use an interactive rebase to squash the fixup into the commit revert commit, and force-push back to your fork to fix this Pull Request.

Also, it's best practice to make your changes on a separate branch on your fork (not `master`) and make your Pull Request from there, because it's easier to keep track of multiple separate patches that way, and it makes it easier to update your fork without messing up a PR.

TotallyNotElite:
Thanks for the advice, but I know how to use git :+1:

Ylianst:
Thanks for this! Your right, my original change was a hack until I get back. FYI. I am in South Australia right now, pretty much as far form home as I could be. I really appreciate the help with this pull request.

MailYouLater:
> Thanks for the advice, but I know how to use git 👍

But you amended the wrong commit. ☹️ If you'd done what I said, then the git history would show reverting the code changes without changing the version number, but now the main repo has two commits that change the version number back to a version that was already used (first to v, then to w).  This is very bad form, some projects automatically publish an update based on a new version number, and changing the version number to a previously used version can cause them major issues.

TotallyNotElite:
Some projects do do that, but it's important to note that the pipeline wouldn't even run on the first commit in the first place. Also, I'm not a big fan of amending revert/merge commits in general.

MailYouLater:
Depends on what they're using for CI, and how it's set up.
Whatever, what's done is done, the important part is that the fix is in. 👍

Issue #616, opened 2019-10-29
Plugin update checker / status page


MailYouLater:
Now that MeshCentral supports plugins, it would be nice if the plugin standard included an update URL the server could use to check for updates for a particular plugin. This would be necessary to make my request possible.

I'd like the MeshCentral web UI to have a "Plugins" page in the "Server" section, which can show an admin which plugins are installed on the server, which of those plugins are enabled, and which ones have new versions available.  Including the ability to enable/disable plugins and a one-click update function would be awesome, but (assuming people will actually make & publish MC plugins) even just being able to easily check on one page to see which plugins are out of date will (eventually) be extremely helpful to anyone who uses MeshCentral with plugins.

Ylianst:
Probably one for @ryanblenis to answer. I have not played with plugin's myself and have just made sure that any plugin code does not run or impact anything if a server is deployed without them. I do think that if plugin's start being important, this request is more than reasonable.

ryanblenis:
I've been weighing the options here for the better part of a week in preparation for this and I'm still in the planning phase of everything that needs to be implemented for this interface to be done correctly.

In addition to what you've mentioned, we also need to consider:
- Version compatibility between the plugin/MeshCentral
- The plugin having its own administrative interface (to manage plugin settings and / or provide its own functionality)
- User permissions delegation to the plugin and / or parts thereof

I'm still making my way through seeing how all the parts of MeshCentral all work together to make this happen, but rest assured, it's coming as fast as I can!

MailYouLater:
I'm not too worried about how quickly this feature gets added, I just felt like it needed to be on the roadmap. Glad to hear you're thinking about how all the parts of a plug-in system come together.

Also, as a side-now for the future. If a one-click update function does get added at some point, it would be really nice to be able to bring up a changelog first.

Ylianst:
I do like where this is going. Looking forward to where this goes.

geanferrani123:
Guys, Good Evening!
I would like to know the features that these system plugins bring, where I view the list of plugins, I made a topic questioning this but unfortunately was closed, could anyone explain me?

MailYouLater:
[...]

PR #676, opened 2019-11-23
Plugin GUI

ryanblenis:
Ok- kind of a large one here!

Features / Notes: 
- Adds a core "plugin" MongoDB collection/NeDB file to track and manage installed plugins
- A new GUI administration interface ("My Plugins" tab)
- Add plugin configurations by referencing a "Plugin Configuration URL" from your friendly neighborhood plugin developer! (Plugin developers specify plugin name, description, download URLs, MeshCentral compatibility, etc.)
- Installation of new plugins without requiring a server reboot
- Disable / Delete plugins
- Ability to check and install updates (checks an update URL for a file containing the JSON configuration)
- Links to the project home page
- Links to the project changelog (no interactive viewing yet)
- Checks for plugin version / MeshCentral version considerations
- Allows plugins to offer their own administrative interface for managing settings/options
- Plugin installs / disables dispatch an event to update the pluginHandler for all logged in users
- After installing / removing a plugin the user is offered to push new meshcore's to all nodes they have access to (useful if the plugin adds functionality to nodes)
- If a plugin offers an administrative interface, the name of the plugin is clickable
- - To handle a possible large amount of plugins, tabs are created as needed and can be removed via right-click context menu
- Downgrading plugins is currently hidden pending actual plugins / testing data to be available

Enabling / Installing:

Simply add (or replace your existing) config.json plugins option with the following and restart your server:
>"plugins": {
>         "enabled": true
>    },

Note: if the `list` attribute exists in the `plugins` object, it should be removed unless you are developing and testing locally.

Screenshot / Preview:
Screen Shot 2019-11-22 at 9 30 46 PM

Looking forward to your feedback!

Satisfies #616


Ylianst:
Wow, this is great! I am going to accept this and make some changes. For one, if plugins are not used, I want to make sure no additional code paths are used or made available for use. I will take a look at your new tab, I may discuss having it "My Server" since that tab is administration only. In any case, let me take a look at this a bit.

ryanblenis:
Thanks! FYI I used the top level tab to take advantage of the existing navigational elements, as well as if the need/want should arise in the future to make parts of the Plugin Admin interface available to non-admin users.

I've implemented it as securely as I possibly could and attempted to not load/expose any code paths that don't need to be (e.g. plugins disabled does not expose any plugin URLs/functionality).
Any questions on the implementation feel free to let me know.

Ylianst:
Just published MeshCentral 0.4.4-t with the new plug-in system. I did a bunch of user interface fixes and moved the tab to "My Server". I did not understand the "goPlugin()" and "noGoPlugin()" methods and so, commented then out for now. Can you explain the usage and/or have a example plugin that uses them? So far, I have only used you sample plugin (https://github.com/ryanblenis/MeshCentral-Sample).

MC2-PlugIns

Ylianst:
Also, I added proxy support for plugin management by doing this in the settings section:

```
    "plugins": {
      "enabled": true,
      "proxy": "http://proxy.domain.com:123"
    }
```

This way, I can downloads the plugins behind an HTTP proxy.

ryanblenis:
Hi Ylian, lots of changes to the code! My apologies for making you do the extra work. I'm still getting the hang of the preferred coding style of the project and trying to match it! Great work integrating it though!

There's another plugin you can load that utilizes the goPlugin functionality: [DevTools](https://github.com/ryanblenis/MeshCentral-DevTools) that (while it only contains plain text with a few links) made a few development tasks easier for me.

When a plugin indicates it has an admin section (hasAdminPanel: true) the name of the plugin gets an A tag with a link. Clicking it invokes goPlugin().
The goPlugin function adds (or if it already exists, switches to) a tab for the plugin, and calls an iframe in the content area that is handled by the pluginHandler->plugin get/post methods. The iframe is not reloaded on subsequent calls unless the tab was closed (to preserve changed data).

Not knowing how the plugin within MeshCentral might grow, I thought it important to keep the area dynamic so users can open/close the tabs they need to fit their viewport.

You can get a feel for the functionality by rolling back to the initial merge and installing [DevTools](https://github.com/ryanblenis/MeshCentral-DevTools), [EventLog](https://github.com/ryanblenis/MeshCentral-EventLog), or both, as they both utilize the admin panel.

A few screenshots of the intended functionality:

DevTools was open, but we're on the main tab:
Screen Shot 2019-11-25 at 6 13 35 PM

Event Log was clicked (tab + content added and switched to)
Screen Shot 2019-11-25 at 6 06 06 PM

DevTools tab was selected (or clicked again from the main plugin list)
Screen Shot 2019-11-25 at 6 05 52 PM

Right-clicking to close the Event Log tab:
Screen Shot 2019-11-25 at 6 10 28 PM

The noGoPlugin() function is a cleanup function that, when the user clicks into another area of MeshCentral, hides the div's containing the iframe'd plugin content.

Ylianst:
Oh perfect. Thanks for the info. No worries about making me work, I appreciated your pull request. I will work on getting the plugin tabs integrated. Can each plugin have at most one tab? I imagine yes, but want to make sure. I am working on a different feature now, but once done, I will get back to plugins.

ryanblenis:
Yes, it was designed as one tab per plugin. Plugin Devs can add their own tabs within their iframe to separate their content / management interface.

Thank you! I'm hoping some others jump on board with developing plugins with the new system!

Ylianst:
Once the new plugin system is doing well, I will blog and add it to Twitter, etc. I want to credit you for the work. This is probably a week or more away, but if you have a blog/page you want me to promote, let me know. As with all these new features, there are 1000x more users of the plugins than developers for them.

ryanblenis:
That's great, thank you! I don't really have a blog/page for myself. I have an MSP and just happen to enjoy coding. When I came across MeshCentral it seemed unique and it has a lot of potential in the RMM space (as it seems many people already use it for remote access and some management/configuration purposes). I know with some work and additional features it could ultimately be an invaluable tool for MSP's and in-house IT to escape some traditional, pricey RMM solutions- and I've always loved open source! I'm just happy to be contributing what I can.

You have created an excellent framework with an intuitive codebase. I already see great things from the project and see only better things down the road! Thank you!

jsastriawan:
Amazing!.. this is the kind of impact we would like to have to the ecosystem. Thank you for the contribution and this is definitely encouraging us to do better.

Ylianst:
Published MeshCentral v0.4.4-w with support for the plugin web page :)

MC2-PlugIns2

Ylianst:
At some point, I would like an "Upload Plugin" button to upload a plugin directly from a local zip file. It would be better for people that don't want to download the plugin using a URL.

ryanblenis:
Wow, looks good! You're much quicker on UI stuff than I am. One potential issue. Not sure if it matters much (at least this early in its inception) but the original PR left the iframe available for each plugin until closed. This update overwrites the iframe. My thoughts are regarding an admin potentially needing to switch between two plugin windows to check data before completing a task in one of them. Currently this destroys any partial updates/changes to the iframe. Not a big deal now with no real plugins out there, but it's something I thought would come into play down the road and thought I'd mention it.

I thought about local zip files as well (hint: if you add the "list" parameter back into the config it will load only those plugins and ignore the plugin database). The URL based approach I ultimately took was 4 fold: 

- Easy copy/paste installation for the user
- Encourages an open source plugin economy
- Encourages users to keep plugins up to date (security issues / MeshCentral compatibility / etc.)
- Eases the ability for MeshCentral to integrate a "Store" to search / one-click-install in the future with a list of plugin URL's

I took pause on implementing the local zip upload because of the aforementioned and because I didn't know if a local zip upload should:
- exclude the update checker / update from web functionality
- not enforcing any of the plugin config.json attributes besides name and shortName
- how these above distinctions should be displayed to the user

Thoughts welcome!

Ylianst:
I see what you mean by only having iframe, I can fix that to have one for each plugin name.

I really like being able to dynamically add/remove/configure plugins, I think it's wonderful. It's much easier to "play" with plugins this way.

It's probably correct that if you use a local zip file, you don't want to perform checking and auto-update. At that point, your doing fully manual plugin install. It's also good because some are very concerned about the server doing any checks of getting any update of any kind without administrator present. Security is rightfully really important, especially for a RMM server.

MailYouLater:
I don't think that fully automatic updates (checking for and installing updates without any administrator action) should ever be a default, opting instead for update notifications where the admin still has to initiate the update, but fully automatic updates may be useful in some situations (e.g. if a company is developing their own plugins, they may want to have them auto-update as soon as they make the update available) but should probably be enabled on a per-plugin basis.

Also, just because you're uploading a plugin directly as a zip doesn't mean you don't want to be informed when an update is available, it may, but not always.  I feel like this should either be something that is asked during the install, or perhaps checking for updates could be switched on/off per plugin at will after a plugin is installed, and the default for plugins installed via the built-in download system can be 'receive update notifications' and plugins installed by directly uploading a zip could have the default be 'do not check for updates'.  This fits nicely with the first half of this post, as the fully automatic updates could be enabled in the same place by choosing 'install updates automatically' for a particular plugin

ryanblenis:
Just to note, as it may be unclear from the conversation line. Currently:

- Plugins do NOT auto-update
- Plugins DO check if a new version is available

It looks like what you're requesting to be default is:
- Plugins do NOT auto-update
- Plugins do NOT check if a new version is available

and that admins can opt-in to either:
- auto check for updates
- auto-update (implies previous)

I'm really iffy on the full auto-update, as plugins can ultimately do many things, and could cause users to get disconnected from machines or interrupt a user currently working within a plugin flow. Changing the default auto-check for updates is possible, but for the small amount of data to check for a new version versus admins not being notified at all of updates, I thought it was a fair tradeoff.

MailYouLater:
Yeah, I agree on auto-update being iffy.  I really only mentioned how I would suggest implementing it because @Ylianst had mentioned it in the previous post in a way that made it sound like it was likely to be available.  Conceptually auto-update is a great idea, but it's too easy for it to cause problems, so if you want to make it available, that's OK, but it should never be a default for any situation.

I was mainly trying to point out that people may want update notifications for some plugins, and not others, but which ones will be getting update notifications should be a visible setting that administrators can choose to enable or not, not something that just happens to be enabled or not based on which way you performed the install.

ryanblenis:
Thinking more about it, I think an auto-update opt-in option would be appropriate if we could implement idle user detection or a case where no users are logged on to the server at all. That way it could happen without the possibility of a user experience issue.

MailYouLater:
If implemented, it might be a good idea to mark it in the UI as an 'experimental' feature.

OutbackMatt:
Is there a way that I can remove the DB entries for plugins?

Deleting the plugins from the plugins folder doesn't do that

ryanblenis:
@OutbackMatt ,

If you are attempting to remove plugins, you can disable then uninstall the plugins from the interface.

If you're looking to delete the data the plugins saved, that is plugin dependent, and varies on your DB (NeDB / MongoDB / etc.)

I should write a "data removal" mechanism for each that is accessible from the UI, but for now, if you're using NeDB, a plugin-pluginName.db is created in the meshcentral-data folder for each of my plugins that can be deleted to clear all data. For MongoDB, a collection is made under plugin_pluginName that can be deleted with db.collectionName.drop()

PR #677, opened 2019-11-24
Use PTY if python is available

TotallyNotElite:
Currently, using the terminal functionality will only make the current shell interactive. This means that sub-shells will not have the usual "$" or other values for $PS1, CTRL+C will kill any sub-shells created and backspace and other special key combinations won't work correctly.

This issue occurs whenever a sub-shell is started, including switching users (which is very bad because of the above mentioned problems). Using a PTY fixes this problem completely, by making programs (like bash) think they're running in an interactive terminal.

(PS: I spent a lot of time figuring this out because its REALLY frustrating and DESTROYS the user experience)

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -1100,7 +1100,7 @@ function createMeshCore(agent) {
                         this.prependListener('end', function () { this.httprequest._term.end(function () { console.log('Terminal was closed'); }); });
                     } else {
                         if (fs.existsSync("/usr/bin/python") && fs.existsSync("/bin/bash")) {
-                            this.httprequest.process = childProcess.execFile("/usr/bin/python", [ "python", "-c", "import pty; pty.spawn([\"/bin/bash\"])" ], { type: childProcess.SpawnTypes.TERM });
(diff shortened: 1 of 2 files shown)


TotallyNotElite:
Before: image
After: image

TotallyNotElite:
Looks like the `.min.js` has to be updated too, but I don't understand the difference between the files. There are only minor changes related to "wallpaper" in there. I guess I'll just copy the file over?

krayon007:
The .min one was run thru a minifier to reduce the size of the file. I usually only update the non minified one because ylian runs a script when he publishes that minifies the thing automatically or something like that.

TotallyNotElite:
> The .min one was run thru a minifier to reduce the size of the file. I usually only update the non minified one because ylian runs a script when he publishes that minifies the thing automatically or something like that.

Ah! Thanks for the clarification! If that's the case, it doesn't really serve any purpose. Maybe removing it is a good idea to stop confusion in the future :thinking:

krayon007:
Yeah, when I do testing, I just delete the minified version from my test server. Then the server will just serve the regular version. Same with the agents. I delete the signed ones from the server, so it'll serve the non signed test version.

Ylianst:
Nice work! I am certainly accepting this. If someone can figure out a way to do this without Python that would be even better, but this great.

PR #681, opened 2019-11-26
Fix available plugin actions

ryanblenis:
Looks like something in the pull request / rework caused the denied plugin actions to check based on the index rather than the actual values (so no cant_actions are actually accounted for). This compares the correct values and populates the options appropriately.


Ylianst:
Thanks. I had to change the syntax of that area because Visual Studio 2015 did not understand the for loop.

PR #690, opened 2019-11-27
Fix MeshCentral version comparisons. 

ryanblenis:
Currently breaking versions comparisons for versions -a through -c (ASCII char val is 97, 98, 99) which appears to be a lower version than intended for version comparisons.

E.g. 0.4.4-z is evaluated as 0.4.4122 (z ASCII is 122) so a comparison with 0.4.5-a is 0.4.597, which appears to be lower in version. This PR pads the conversion from -alpha_char to 3 numbers so version 0.4.5-a would become 0.4.5097, thus satisfying an upgrade comparison correctly.

### DIFF
--- a/pluginHandler.js
+++ b/pluginHandler.js
@@ -285,8 +285,8 @@ module.exports.pluginHandler = function (parent) {
                             var s = require('semver');
                             // MeshCentral doesn't adhere to semantic versioning (due to the -<alpha_char> at the end of the version)
                             // Convert the letter to ASCII for a "true" version number comparison
-                            var mcCurVer = parent.currentVer.replace(/-(.)$/, (m, p1) => { return p1.charCodeAt(0); });
-                            var piCompatVer = newconf.meshCentralCompat.replace(/-(.)\b/g, (m, p1) => { return p1.charCodeAt(0); });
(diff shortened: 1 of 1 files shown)


Ylianst:
Excellent. I had no idea this was an issue.

ryanblenis:
Me neither until I saw plugin version upgrade checks failing where they shouldn't be today! Thank you!

ryanblenis:
Just some background if anyone sees this and is interested. The semver module considers the format 0.0.0-letter that the "letter" is a prerelease or tag, so a regular comparison drops it. Hence plugins would only be able to note compatibility with the 0.0.0 version scheme since the letter is dropped.

To fix this and let plugins define a specific version (including the letter) I convert the letter to its ASCII value, making a version number of 0.0.0000 that it can logically compare.

MailYouLater:
Wouldn't it be better to put the letter as a fourth part?  i.e. `0.4.4-z` is evaluated as `0.4.4.122` and `0.4.5-a` is `0.4.5.97`?  This may not be possible with the semver package, but [compare-versions](https://www.npmjs.com/package/compare-versions) supports it.

ryanblenis:
It was my first thought, yes, however semver doesn't seem to support it, it only takes the major, minor, and patch from the looks of it. For the small amount of work vs gain it didn't seem worth it to suggest a dependency to the project just for that.

Ylianst:
Published in MeshCentral v0.4.5-b, hopefully this works :)

PR #698, opened 2019-12-02
Plugin Tab OnStates / Better MeshCore processing

ryanblenis:
A couple small updates:

- OnStates for current plugin tab within the device page
- Move p19 styles to stylesheet from the default.handlebars
- Wait to update meshcore until after all plugins loaded, rather than after each plugin is loaded
- More intuitive registerPluginTab functionality / flexibility (with backward compat)

### DIFF
--- a/pluginHandler.js
+++ b/pluginHandler.js
@@ -51,8 +51,8 @@ module.exports.pluginHandler = function (parent) {
                         parent.db.updatePlugin(plugin._id, plugin_config);
                     } catch (e) { console.log("Plugin config file for " + plugin.name + " could not be parsed."); }
                 }
-                obj.parent.updateMeshCore(); // db calls are delayed, lets inject here once we're ready
             });
+            obj.parent.updateMeshCore(); // db calls are async, lets inject here once we're ready
         });
     } else {
         obj.loadList.forEach(function (plugin, index) {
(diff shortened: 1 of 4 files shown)


MailYouLater:
Sorry if this is too far off topic, but it would be nice to have a persistent 'refresh agent cores' button, maybe in the [DevTools](https://github.com/ryanblenis/MeshCentral-DevTools) plugin.  Also, "[Sample Plugin](https://github.com/ryanblenis/MeshCentral-Sample)" seems like a poor name choice for a plugin that someone requested and will probably actually use it in a production environment.

ryanblenis:
I can see how a "refresh agent cores" could fit into the DevTools plugin, though I don't know why most (non-dev) people would need it other than when updating / installing / changing plugins.

The Sample Plugin was just that- a sample to show how easy it can be to change the functionality to help fill a need/want. Ideally someone would just fork/copy the sample and change the name/path and (likely) some specifics of the formatting I used. I know I copy it now when I'm starting a new plugin!

Ylianst:
This pull request is now in MeshCentral v0.4.5-e that I just published.

For the mesh core, right now I go to the device's "console" tab and you can hit the "Agent Action" on the upper right and use that to clear or update the core. I also "ALT + "Agent Action" will clean the core and "SHIFT" + "Agent Action" with update the core. We use that a lot to push and test new cores on a specific agent. As dev's, I don't really have a use to push the core to all agents, I just restart the server and it does it.

Ylianst:
Also, I want to point out [@MailYouLater request](https://github.com/Ylianst/MeshCentral/issues/699) in case it's not getting attention on the issue page.

MailYouLater:
@Ylianst: It would be really nice to be able to see somehow what these 'secret keys' as you call them are.  Could they either be indicated in the UI somehow, or at least documented in a list somewhere?

Issue #699, opened 2019-12-02
Device's "Plugins" tab should not be visible if no plugins want to display something in it


MailYouLater:
Not all plugins have an interface that you access from the interface of a single device.  If you don't have any plugins that want to show something in the "Plugins" page of an individual device, the tab should be hidden from the devices menu.

![image](https://user-images.githubusercontent.com/1279475/69998917-4fc01280-150c-11ea-95bc-493ee5b6f5e1.png)
![image](https://user-images.githubusercontent.com/1279475/69998989-741bef00-150c-11ea-88ab-5a4700d60b7c.png)
In these images, I don't even have any actual plugins enabled at all (I just have the plugin system enabled but haven't downloaded any plugins), so the Plugin tab will definitely not have any content (yet) and should be hidden until a plugin actually puts something there.

Ylianst:
This is one for @ryanblenis. I agree, that tab should not show unless there is a use for it.

ryanblenis:
Submitted PR (#700) to address this.

Ylianst:
Excellent! Nice and simple fix, I like it. I will have a published version with this tomorrow.

MailYouLater:
I'm surprised at how small the changes necessary to implement this were.  Thanks for getting to this so quick!

MailYouLater:
ACK!  There might be an issue with it.  Will post comments on #700.

MailYouLater:
I can verify that `0.4.5-g` still shows the Plugins tab when needed, but no longer shows it when no plugins use it.

PR #700, opened 2019-12-02
Show/Hide Plugins tab in device view

ryanblenis:
Show plugins tab in device tab only if plugins exists that use that functionality.

Note that while testing I was receiving "updateMapMarkers is not defined" in the console. Seems 
 @Ylianst is working on some GeoLocation.

Satisfies #699

### DIFF
--- a/pluginHandler.js
+++ b/pluginHandler.js
@@ -100,6 +100,7 @@ module.exports.pluginHandler = function (parent) {
                 QA('p19headers', '<span ' + defaultOn + ' onclick="return pluginHandler.callPluginPage(\\''+d.tabId+'\\', this);">'+d.tabTitle+'</span>');
                 QA('p19pages', '<div id="' + d.tabId + '"></div>');
             }
+            QV('MainDevPlugins', true);
         };
         obj.callPluginPage = function(id, el) {
             var pages = Q('p19pages').querySelectorAll("#p19pages>div"); 
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -4528,7 +4528,7 @@
                 QV('MainDevFiles', ((mesh.mtype == 2) && ((node.agent == null) || (node.agent.caps == null) || ((node.agent.caps & 4) != 0))) && (meshrights & 8) && fileAccess);
                 QV('MainDevAmt', (node.intelamt != null) && ((node.intelamt.state == 2) || (node.conn & 2)) && (meshrights & 8) && amtAccess);
                 QV('MainDevConsole', (consoleRights && (mesh.mtype == 2) && ((node.agent == null) || (node.agent.caps == null) || ((node.agent.caps & 8) != 0))) && (meshrights & 8));
-                QV('MainDevPlugins', pluginHandler != null);
+                QV('MainDevPlugins', false);
                 QV('p15uploadCore', (node.agent != null) && (node.agent.caps != null) && ((node.agent.caps & 16) != 0));
                 QH('p15coreName', ((node.agent != null) && (node.agent.core != null))?node.agent.core:'');
 


MailYouLater:
LGTM, I copied the changes into my local install and it seems to work exactly as I thought it should. 👍

ryanblenis:
0.4.4-y in the log is baffling if that isn't your version, but the checkConfigAuth is something I use in my EventLog plugin. Are you using that with NeDB (the default) database type?

MailYouLater:
Yes, but I had disabled it at the time. 😕

ryanblenis:
Hmm, between the 0.4.4-y version and the plugin being disabled something is definitely fishy. That function only exists in that plugin- no core MeshCentral files- so it must have been running.

That said, I'm testing a fix for that in the EventLog project as well as a new issue I found from the IE 11 fixes that broke the plugin page in the MeshCentral project. Give me a few to submit a pull request and update EventLog.

Ylianst:
I published MeshCentral v0.4.5-f with this pull request, however, feel free to make more pull requests.

ryanblenis:
A new PR (#703) has been submitted to fix an issue entering plugin admin interfaces.

EventLog 0.0.17 updated and published to fix some NeDB issues as well. Feel free to let me know if you have any other issues in that project's issue tracker. Thank you.

@Ylianst - any comments on the 0.4.4-y version being output in the mesherrors.txt when he says he had 0.4.5-e installed?

Ylianst:
Thanks for the PR. I will merge that tomorrow. For IE11, yes, one has to take core not to use modern JavaScript in the web application as it breaks older browsers.

For the mesherror.txt file, are you sure the error was not in fact written by an older version of MeshCentral? The errors are appended to the file and never removed. So, I have plenty of errors from previous versions still in my mesherrors.txt. I have never seen this, but I think I get the MeshCentral terminal from the package.json file, so maybe you have one that was not updated? Still, very weird.

MailYouLater:
Yes, I'm certain.  From looking at the existing messages in the Command Prompt window, I can see that I was on `0.4.4-y` when I last launched MeshCentral manually, but since then I had updated it to `0.4.5-a` via the built in updater (click "Check Server Version", Check box, Click "OK"), then I updated again to `0.4.5-e` (also via the built in updater), then I edited the `.js` files, restarted the server via the `resetserver` command.  At that point, I didn't have a `MeshErrors.txt` file.  I tried to install the "Sample Plugin" (Note: I had previously entered the URL to the `config.json` in the "Download Plugin" input, but hadn't clicked "Install", and I can verify that the files had not been downloaded yet.) and that's when it crashed.  After the crash, the only error in the `MeshErrors.txt` file was the one I posted here.

Those two updates from `0.4.4-y` to `0.4.5-a`, then to `0.4.5-e` were both done via the web UI so it wasn't on `0.4.4-y` at all on the 3rd.

Note: I left it last night with the EventLog plugin installed, and my Command Prompt was full of `Error loading plugin handler (TypeError: obj.db.checkConfigAuth(...).then is not a function)` lines in the morning.

I just updated both, and so far I haven't seen any error messages, so hopefully it's fixed.  If not, I'll open a new issue.

Regarding the incorrect version being in the MeshErrors.txt file, I'll make sure I can reliably duplicate it, then I'll open a new issue that's specifically for that.

MailYouLater:
I'm getting `Error loading plugin handler (TypeError: Cannot read property 'serveraction' of undefined)` in the Command Prompt output again with MC `0.4.5-g` and EventLog `0.0.17`, but it happens at seemingly random times, and only when the EventLog plugin is installed, but disabled.  I'll open a new issue for it later, but I feel like it's an MC issue, not an EventLog issue, as I don't feel MeshCentral should be doing anything other than reading the basic name/version/etc. with a plugin that's not enabled.

ryanblenis:
Ah, that description tells me that the issue is that the endpoint cores were not updated once the plugin was disabled.

E.g. 
- Server starts with EventLog enabled. 
- Endpoints check in to server and get cores with EventLog enabled.
- Event log gets disabled on server
- Clients still attempt checking in to server with new event logs. These fail with your error because the server component no longer exists for them to "talk to."


Given the constant stream of events from endpoints, the endpoints get their own periodic timer to collect and report back new logs (instead of being instantiated by the server every time).

The simple fix would be to click the update agent cores button when changing the state of EventLog (or any plugin that modifies agent cores). The only reason this is not automated is because this could effect thousands of machines, and if doing multiple installs or uninstalls, you would not want to flood the server/endpoints with this traffic, but only when you've completed changing the plugins you intend to change.

MailYouLater:
But it happens after I've clicked on the Refresh Agent Cores button, and even happens after I restart the server with EventLog already disabled. 😕

ryanblenis:
Ok, I see the issue when disabled and you've refreshed cores (see new PR #707), but it should never load if disabled when the server starts- so meshCores should be fresh and not include any plugin data that is disabled. I could not replicate that.

Given the new pull, the problem is that the Update Mesh Cores call wouldn't do anything if you'd only disabled the plugin- so the clients would still send data. Thanks for catching that!

MailYouLater:
Hmmm.  I disabled and removed the EventLog plugin and then I reset the server yesterday, and today I have a few lines of `Error loading plugin handler (TypeError: Cannot read property 'serveraction' of undefined)`, but not many.  If your diagnosis is correct, the only thing I can think of is that maybe a few computers that weren't online last night (so their cores didn't get updated) connected and tried to send EventLog data.  IDK... but once PR #707 is included in a new version I'll update and test to see if I still get these messages.  Till then, I'm taking a little break from this issue.

PR #703, opened 2019-12-04
Plugin Admin Links Broken

ryanblenis:
Hi @Ylianst ,

It looks like you refactored some plugin code using inline javascript variables for templating in the "IE11 Fixes" commit. I didn't realize that broke IE11, sorry!

While you were fixing it, quotes were removed from the plugin admin links, meaning javascript functions were called with the name of the plugin as variables instead of text. (e.g. goPlugin("eventlog", "Event Log"); became goPlugin(eventlog, Event Log);)

This fixes that, as well as the opportunity that a plugin may contain a single quote that would break it in the same fashion.

### DIFF
--- a/pluginHandler.js
+++ b/pluginHandler.js
@@ -412,12 +412,16 @@ module.exports.pluginHandler = function (parent) {
                                         } else {
                                             parent.db.updatePlugin(id, { status: 1, version: version_only.name }, func);
                                         }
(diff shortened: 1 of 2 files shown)


ryanblenis:
Also added new try/catch during initial plugin instantiation to catch any potentially malformed plugins from crashing the server.

PR #722, opened 2019-12-12
Remember the last plugin tab and jump back to it.

ryanblenis:
Small update to remember a last plugin tab selected, if any, and jump back to it.


Ylianst:
Thanks for this, it's in MeshCentral v0.4.5-s

PR #735, opened 2019-12-15
Safety checks, cleanup, and add eventing capability to plugins

ryanblenis:
- Clear plugin headers / pages on device change
- Safety check existence of plugin on device change
- Add plugin eventing capability


Ylianst:
This is published in MeshCentral v0.4.5-v.

PR #761, opened 2019-12-25
update NL in translate.json

petervanv:


Ylianst:
Thank you! This was included in MeshCentral v0.4.6-p. I did have to fix some issues with the JSON being not quite right, but nothing major.

PR #762, opened 2019-12-26
Additional hooks, better passing of arguments

ryanblenis:
Fix argument passing for internal pluginHandler calls.
Add two new hooks for plugin devs.


Ylianst:
Thank you. Included in MeshCentral v0.4.6-p.

PR #769, opened 2019-12-28
Add 2 new hooks for detecting page changes in the webUI

ryanblenis:
Allows plugins to take action on page changes


Ylianst:
Published MeshCentral v0.4.6-r with these included. Thank you!

PR #777, opened 2019-12-31
Tweaked NL translation including latest additions (0.4.6-u)

petervanv:
The latest tweak of the language file, includes the user group settings


Ylianst:
Thank you! I see you already translated the strings for the new user groups feature, this is wonderful! I will include this next time I get to work.

PR #783, opened 2020-01-03
Add call to onWebUIStartupEnd when pluginHandler is refreshed, so new…

ryanblenis:
… plugins can run their startups, if they have one.

Add plugins option (disabled by default) to the sample config


Ylianst:
Published with MeshCentral v0.4.6-w :) Thanks.

PR #788, opened 2020-01-04
Add semver module for plugin version and compatibility checks

ryanblenis:
Hi @Ylianst,

It [turns out](https://github.com/ryanblenis/MeshCentral-ScriptTask/issues/1) that if NODE_ENV is production, semver is never installed. However in NODE_ENV=development it is a devDependency of several other deps. This causes new plugin version checks to fail on production systems without semver installed (as well as other odd plugin behavior). This checks and adds the semver module if plugins are enabled.


Ylianst:
Nice! It will be in MeshCentral v0.4.6-x when I next publish it.

Ylianst:
Just published MeshCentral v0.4.6-x with this included.

Issue #765, opened 2019-12-27
German Translation for MeshCentral


veitw:
Hi,

I thought I could spend some time over Christmas and New Year to contribute a German translation, but I am unable to get the Ressource Translation Tool up and running:

On Windows 10 Pro x64 1903 with .NET 3.5 and all updates installed, I see the following error when I click on a cell in the "Translated Text" column.
![image](https://user-images.githubusercontent.com/16285895/71516929-fbf6de00-28ab-11ea-9cad-4e98773b20f9.png)
The error messages translates to: The object of type "System.Collections.Generic.Dictionary`2[System.String,System.Object]" cannot be converted into type "System.String".

Also I cannot start the application on Linux using Mono/Xamarin, because WPF for .NET 3.0 is not implemented.

My third attempt was using Wine, but the win32 architecture target required to install either dotnet35sp1, dotnet35, dotnet30sp1 or dotnet30 winetricks targets seems to be b0rken in my 64bit Wine and all of them fail with one or another failure.

Is there a version of Ressource Translation Tool available that is built against a less WinXP-ish version of .NET and runs on either Windows 10 or Mono/Xamarin or Wine (e.g. dotnet40 installed ootb without problems)?

Best regards,
// Veit

Ylianst:
Just posted a fixed [ResourceTranslator here](http://info.meshcentral.com/downloads/MeshCentral2/ResourceTranslator.zip). The problem was that the .json file I posted was not correct, it has an invalid entry at the top of the file in the first entry. I fixed the translate.json, but also fixed the translator tool to ignore any entries that are of invalid type.

Should work now. Let me know if it fixes it.

veitw:
Seems to work, thank you. :)

veitw:
Two questions:

1. Is there a short cut to jump to and select the next row that has no translation so far? This would speed-up the process quite much.

2. Where are singular/plural for entries such as " - Reset in {0} day{1}." configured? Because simply appending "s" for plural does not work for German. Or shall I remove the {1} placeholder and include the plural into the string, e.g. "{0} Tag(e)", "{0} Stunde(n)", "{0} Minute(n)"?

veitw:
The tool seems to work fine now, but the JSON output is quite compact and thus not diffable against the original translation.json.
Which tool/commands do you recommend to bring it into the same format (and order) as the input file?

Ylianst:
Hi.
1. There is no short cut to move to the next one, but I can work to add that.
[...]

PR #824, opened 2020-01-13
Updates German translation

veitw:
- fixed pronoun due to resolved ambiguity

### DIFF
--- a/translate/translate.json
+++ b/translate/translate.json
@@ -4398,7 +4398,7 @@
     },
     {
       "cs": "Vytvořit",
-      "de": "Erstelle eine",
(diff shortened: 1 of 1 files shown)


veitw:
The following string are still untranslated due to ambiguity:
 * "Mesh Relay" -- is this a (new) software tool? I am not sure whether and how I shall translate this.
 * "PluginHandler could not event message: " -- there is a verb missing, thus I am not sure what it means.

Any hints welcome.

veitw:
Refers to  issue #765 .

Ylianst:
Thanks for the fix, this will be in the next published release.

Issue #793, opened 2020-01-06
Bug: My Devices -> Group Actions no longer work with some language translations [0.4.6-y]


veitw:
Hi,

I have noticed that with some language translations, the Group Actions in My Devices tab no loger work, e.g. selecting a device and choosing "wake up" or any other action, nothing happens.

I could observe the problem in current 0.4.6-y and back to at least 0.4.6-v, but I think older versions are affected as well.

So far I verified that it still works for:

 - ?lang=en
 - ?lang=fr

... and that it does not work for:

 - ?lang=de
 - ?lang=nl

Other languages have not been tried.

I suspect that in some languages, strings have been translated that are actually used as identifiers and not interface text, and thus shall not be translated at all.

Best regards,
// Veit

Ylianst:
Thanks for the report, I will take a look at this.

vitko-bg:
That is interesting fact ....
@Ylianst if you can update what should not be translated it will be amazing.
BTW, Russian almost done, once it is ready - Bulgarian will be next :)

Ylianst:
@vitko-bg Oh my!! For the translation updates, the translate.json that is included in the /node_modules/translate should always be current with the latest strings. Each time I publish a new MeshCentral server on NPM, I re-generate this file.

veitw:
Hi @Ylianst,

I identified one more string that I suspect that they may break something if translated, but I was unable to find proper test case:

* "DataChannel"
Used as an identifier for WebRTC, seems to work for the German translation "Datenkanal" but I suspect it could invite encoding-based trouble with e.g. Japanese or strings that contain accents or umlauts.

As it seems to be used nowhere except once for naming the data channel itself without any display, I suggest to remove this string.

Best regards,
// Veit

veitw:
Problem solved after update to 0.4.7-l.

PR #825, opened 2020-01-13
Reenable broken group actions in My Devices overview screen

veitw:
The group actions in My Devices overview screen were broken in the following language translations due to translation of an identifier as reported in #793 :

 * de
 * nl
 * pt

The translations of the identifier have been reverted to the English original to mitigate the problem.
Final solution should be the removal of the identifier "DeviceCheckbox" from translatable strings.

### DIFF
--- a/translate/translate.json
+++ b/translate/translate.json
@@ -5379,11 +5379,11 @@
     },
     {
       "cs": "DeviceCheckbox",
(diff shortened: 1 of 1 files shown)


veitw:
This mitigates issue #793 .

Ylianst:
Thanks for the fix, this will be in the next published release.

PR #827, opened 2020-01-13
Fix #826

### LINKED ISSUE #826: Install error - Linux (Debian 10)
When attempting to install meshcentral as a service, I get the following response:

```
Installing MeshCentral as background Service...
Writing config file...
Enabling service...
ERROR: Unable to enable MeshCentral as a service: Error: Command failed: sudo systemctl enable meshcentral.service
Failed to enable unit: File meshcentral.service: Invalid argument
```

I'm running `node ./node_modules/meshcentral/meshcentral --install` while logged in as root. I attempted to add sudo, just to be sure it's not permissions related, no dice.

Node version: 12.14.1
MeshCentral version: v0.4.7-j

egallup02360:
Service installation process on Linux. Now confirmed to work for Debian 10, Ubuntu 18.04, CentOS 7.6 and Fedora 31.


Ylianst:
Thank you for this pull request! Much appreciated. I do see this:

config.split("\n").join("\n")

That looks like it does nothing. Should it not be .join("\\n") as we need the escape the new lines? I am traveling this week, but will try to make a release or two during my travels, this would be a great fix.

egallup02360:
You're absolutely right, weird. It appears the new lines don't need to be escaped or transformed at all. I removed `.split("\n").join("\n")` and confirmed this still works on debian and red hat based systems.

Ylianst:
Perfect! Thanks for making that fix. I am back from travels this weekend and will publish this ASAP. Apologies for the delay.

Ylianst:
Published MeshCentral v0.4.7-k with this included.

PR #834, opened 2020-01-16
German translation update (#765, #793)

veitw:
Reverted translation of ID "DataChannel", corrected ambiguity of "SharedProcess" and corrected pure case of "Soft disconnect agent".


Ylianst:
Thank you!! This will be in the next published version.

Ylianst:
Published MeshCentral v0.4.7-k with this included.

PR #868, opened 2020-01-24
[WIP] Add dynamic resizing of xterm tty

### LINKED ISSUE #859: [xterm.js 0.4.7-s] scaling problem
Step to reproduce:

- use Windows as client (firefox ESR 68)
- upgrade to v 0.4.7-s
- connect to any linux host by xterm.js
- just see screens

изображение

изображение

THE GOOD NEWS: MC **WORKS** in xterm.js. YES! YES YES

TotallyNotElite:
(Feel free to push to this branch btw.)

Basically just sending the SIGWINCH signal and reacting to it in the shell. Ideally data would be sent along with the signal somehow so the shell can just resize without calculating the terminal size.

Why I marked this as WIP:
1. I didn't test this on anything other than script+bash
2. Resizing the window multiple times in a short time frame for multiple seconds causes the terminal to "lag out" temporarily, can be cancelled by pressing ctrl+c.

fixes #859

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -39,8 +39,7 @@ var MESHRIGHT_LIMITEDINPUT = 4096;
 
 function createMeshCore(agent) {
     var obj = {};
-    if (process.platform == 'win32' && require('user-sessions').isRoot())
-    {
+    if (process.platform == 'win32' && require('user-sessions').isRoot()) {
         // Check the Agent Uninstall MetaData for correctness, as the installer may have written an incorrect value
         try {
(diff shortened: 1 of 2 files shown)


TotallyNotElite:
@Ylianst I didn't intend to get this merged yet. See the [WIP] tag in the name.

image

Ylianst:
Got it. Bryan and I just tested it and it does not work on my Raspberry Pi. Bryan can improve his kill to accept a different signal value, but invoking the kill command like this is a good temporary solution.

In any case, I get the gist of this... thank you so much, we where not making much progress.

PR #870, opened 2020-01-25
Add dynamic resizing support to xterm.js terminal

TotallyNotElite:
Some notes:

- Only works in PTYs. I couldn't get xterm.js to not do [this](https://user-images.githubusercontent.com/38938720/73110406-90357e80-3f06-11ea-8ee8-8d919883e5fb.png), even without dynamic resizing enabled.
- I removed the following line because it shouldn't even be called when not in terminal mode, + it didn't work for user terminal sessions: `                if (ws.httprequest.protocol == 1) { // Terminal`

Would be great if you could test this first. I tested this on my Manjaro Linux install, but you never know.

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -39,8 +39,7 @@ var MESHRIGHT_LIMITEDINPUT = 4096;
 
 function createMeshCore(agent) {
     var obj = {};
-    if (process.platform == 'win32' && require('user-sessions').isRoot())
-    {
+    if (process.platform == 'win32' && require('user-sessions').isRoot()) {
         // Check the Agent Uninstall MetaData for correctness, as the installer may have written an incorrect value
         try {
             var writtenSize = 0, actualSize = Math.floor(require('fs').statSync(process.execPath).size / 1024);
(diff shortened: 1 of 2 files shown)


krayon007:
I have 30 VMs of different Linux distros that I can test it on :)

I also added support to child-process.kill to support passing in a signal too, so I can test that at the same time.

Ylianst:
This change is published in MeshCentral v0.4.7-x. I tested it and it does not work on my Raspberry Pi, but does on my old Linux box. However, on my old Linux, if I am in "pico" and resize, it does not work, I have to quit and come back to "pico" and I get the new size.

TotallyNotElite:
Hmm, maybe there's no way around it after all. @Ylianst would it be acceptable to temporarily store the sizes in a file in /tmp? Name would be something like /tmp/mcterminal${pid}

krayon007:
Funny you mention that, as that's actually how I implemented it in my testing.

Ylianst:
Note on this topic. This dynamic resizing code was removed since we are going with a better approach using Kernel support for PTY. It will be in the next agent. Right now, re-size is not supported but this is temporary.

PR #871, opened 2020-01-25
Improve the way dynamic dependencies work

TotallyNotElite:
Meshcentral can now specify specific versions of dynamic dependencies and the module system will try to match said version automatically. This removes the need for the weird greenlock hacks and allows meshcentral to control dynamic modules similarly to package.json.

I also added otplib as a dependency, since it's required by default.

### DIFF
--- a/meshcentral.js
+++ b/meshcentral.js
@@ -2168,20 +2168,27 @@ function getConfig(createSampleConfig) {
 function InstallModules(modules, func) {
     var missingModules = [];
     if (modules.length > 0) {
+        var dependencies = require("./package.json").dependencies;
         for (var i in modules) {
             // Modules may contain a version tag (foobar@1.0.0), remove it so the module can be found using require
-            var moduleName = modules[i].split("@", 1)[0];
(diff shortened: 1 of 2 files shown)


Ylianst:
Technicality, you don't need otplib.js if the server is setup to do domain authentication (SSPI), that is why it was not present. I do agree that it's *almost* always used.

Ylianst:
It's published in MeshCentral v0.4.7-x, hoping there are no unintended consequences. If someone updates an old MeshCentral server with GreenLock v2, this new code will not update GreenLock and will cause the serve to fail. Probably not common. I removed the v3.1.5 version for GreenLock as now MeshCentral works with v4.

TotallyNotElite:
@Ylianst why will this code not update Greenock? I mean now that you removed the version it won't, so I suggest that you add the v4 version to the module so npm can force the update.

TotallyNotElite:
@Ylianst just so you know https://github.com/Ylianst/MeshCentral/commit/b382a8fcc07a66beef583db8b675d20082861593#r36967548

PR #889, opened 2020-01-29
Dutch translation update 0.4.8-e

petervanv:


Ylianst:
Quick note to say that MeshCentral 0.4.8-f includes these updates. Thank you!

PR #916, opened 2020-02-05
Fix issues with linux/macos terminal

TotallyNotElite:
- Fixed #858 again, added comments ([regression found here](https://github.com/Ylianst/MeshCentral/commit/4c70ee0d0b72cda171d12dcb155a413f3bfdb21e)) (not included in PR anymore, committed directly by @krayon007)
- Removed now useless commands from sh setup (pty makes these redundant) (not included in PR anymore, committed directly by @krayon007)
- Fixed non xterm.js terminal not starting at correct size ([regression found here](https://github.com/Ylianst/MeshCentral/commit/9ed44cef5f1e87ed18a290f674e64673b4093213#diff-4db0233255dcb9c9b229513a61f80f58R165))
- Other minor changes

PR #919, opened 2020-02-05
Fix issues with linux/macos terminal

TotallyNotElite:
Literally the same pull request because the other one got closed for [no reason](https://github.com/Ylianst/MeshCentral/pull/916#issuecomment-582647871)...


Ylianst:
Ok, apologies. I miss communicated with Bryan, accepted it. Will be in the published version.

PR #936, opened 2020-02-10
Add existing CIRA connection exception from periodic AMT scanning.

jsastriawan:
This to alleviate issue when AMT CIRA status unintentionally reset due to TCP check failure. We are seeing multiple instances where AMT connectivity state unintentionally set to offline because TCP check returned false because AMT device is in CIRA mode. 

We may need to add additional exception for checking for hybrid mode because amtscanner attempted to perform scan on device connected over the internet (behind NAT). We should only perform AMT TCP check if the source IP of MeshAgent connection is one of the IP of the interfaces reported by AMT (via MeshAgent).

CC: @matt-primrose @Ganeshr93 for review


Ylianst:
Excellent! Thank you.

PR #937, opened 2020-02-10
Fix missing sendText on amt-redir-ws.

jsastriawan:
This fix AMT terminal not working because of missing obj.parent.sendText on amt-terminal. The other alternative fix is to replace occurence of obj.parent..sendText at amt-terminal with obj.parent.xxSend. but I think add alias for sendText is clearer.


Ylianst:
Thanks for this. I was working this morning on making Intel AMT SOL work with the legacy terminal, and the MeshAgent using XTerm.js, I hit this issue by random.

PR #938, opened 2020-02-10
Dutch language update 0.4.8-w

petervanv:


Ylianst:
Thank you! This will be included in MeshCentral v0.4.8-y when it gets published later today.

Ylianst:
Published!

PR #980, opened 2020-02-27
U2F fix in second place

ryanblenis:


Ylianst:
Thanks!

Ylianst:
Just made the same change on the mobile site 👍 Will be checked in shortly.

PR #1000, opened 2020-03-07
language update 0.4.9-z

petervanv:
the screenshot shows (circled) text not to be found
verified and not verified in the translate.json
untranslated-strings


Ylianst:
Perfect! Thank you!

PR #1011, opened 2020-03-12
meshCentralCompat version comparison changes and updated docs for cla…

ryanblenis:
…rity

Update the documentation to more accurately reflect the new version comparison.

Make versionCompare and versionToNumber member objects (in the event a developer wants to maintain compatibility with older versions, but "feature block" some aspects with checks in their own code based on the version of MC currently in use).

I also saw the changes to rimraf you made. I believe rimraf gets installed as devDeps of other dependent packages, so it shouldn't have a real change in most environments. However it does present the issue that without rimraf installed the plugin folder will not be removed from the file system. Aside of storage space being used that would otherwise not be, it won't affect anything. However I'm open to your thoughts on whether we should add rimraf as a dependency if plugins are enabled. It looks like Node's native recursive directory removal is still in an "Experimental" state as of v12 and v13.

Thank you!

### DIFF
--- a/pluginHandler.js
+++ b/pluginHandler.js
@@ -276,12 +276,12 @@ module.exports.pluginHandler = function (parent) {
 
     // MeshCentral doesn't adhere to semantic versioning (due to the -<alpha_char> at the end of the version)
     // Convert 1.2.3-a to 1.2.3-3 where the letter is converted to a number.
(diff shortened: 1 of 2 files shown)


Ylianst:
MeshCentral v0.5.0-f published with this fix. Thank you. For "ramraf", I don't mind having it. I was just using a really old version of NodeJS (I try to support 6.x and 8.x if possible) and "ramraf" was not working for me. I will remove the exception catching on the next version.

PR #1021, opened 2020-03-15
Additional Config Options

MordyT:
Couple of options are missing from the sample config that I had to track down...


Ylianst:
Makes sense, thank you!

Ylianst:
Published MeshCentral v0.5.0-h that includes this. Thanks.

PR #1029, opened 2020-03-16
update untranslatable strings !!

petervanv:


Ylianst:
You are crazy fast! Thank you. This will be in the next published release later today.

PR #1030, opened 2020-03-17
Adding trustedproxy to sample config

MordyT:


Ylianst:
Good one :)

Ylianst:
Note that "trustedproxy" must be set to the IP address of the proxy, not "true". "tlsoffload" can be set to true is you want to indicate the TLS is offloaded but no trusted proxies are set.

PR #1035, opened 2020-03-17
Updated with new options

MordyT:


MordyT:
Added Invite Code Config Setting

PR #1081, opened 2020-03-27
Dutch language update 0.5.0-W

petervanv:


Ylianst:
You are so fast! Thanks you. You are translating strings that are not features that are not released yet :)

PR #1113, opened 2020-04-04
Dutch language update 0.5.1-D

petervanv:


petervanv:
The files translated, are open again to translate, do i need to? is it possible to remerge this file again ?

PR #1152, opened 2020-04-13
Add metadata to the relay connection for user -> device tracking

ryanblenis:
Update server console command 'relays' to display friendly information

A small quality of life update for users who utilize relays so that the Server Console -> "relays" command outputs the user and peer device utilizing the connection.

image

I often use the "relays" command to see if my server is being used for relays (Work From Home) people before upgrading / restarting the service. Rather than just numbers (and backtracking from IP's to devices), this adds the user id and name of the node they are connected to.

### DIFF
--- a/meshrelay.js
+++ b/meshrelay.js
@@ -20,6 +20,7 @@ module.exports.CreateMeshRelay = function (parent, ws, req, domain, user, cookie
     obj.user = user;
     obj.ruserid = null;
     obj.req = req; // Used in multi-server.js
+    obj.metadata = {};
 
     // Check relay authentication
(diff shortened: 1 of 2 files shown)


Ylianst:
This one caused an immediate server crash because it's build to only handle the case where the relay is setup using a routing cookie. Many times, relay connections are not created like that and so would cause a server exception. I changed it to only display the user name, but not the remote machine for now.

I may make some big change to the relay code in the coming weeks, so I may keep it like this until the big changes are made.

ryanblenis:
Oi, sorry about that! I'll keep an eye out for the relay changes. Thank you!

PR #1155, opened 2020-04-14
1. Add latency timer and callback opportunity to CreateAgentRedirect

ryanblenis:
2. Add latency to Remote Desktop view

Another quality of life update. I added a latency timer to the CreateAgentRedirect object that, once started, sends a control channel message every ~3 seconds to measure latency between the MC user and the client machine.

As an example (and probably one of the more useful places for it) I added the output to the Remote Desktop device tab- but it can be easily added to anywhere a CreateAgentRedirect is used to measure latency via `varName.latency.onUpdate([callbackFunc])`

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -1912,6 +1912,10 @@ function createMeshCore(agent) {
                 if (sdp != null) { ws.write({ type: 'answer', ctrlChannel: '102938', sdp: sdp }); }
                 break;
             }
+            case 'latency': {
+                ws.write({ type: 'latency', ctrlChannel: '102938', time: obj.time });
+                break;
+            }
         }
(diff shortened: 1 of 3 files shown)


Ylianst:
I accepted it, but made a bunch of changes to it. It's a little odd that the onUpdate(0 method was both used to subscribe to events and to fire the events, but no worries. I removed the use of a timer since, I only use times if absolutely necessary and in this case it was not needed. I just mesure the time since the last RTT check. I also moved the displayed value on the lower right near the timer. Lastly, the way this is built, if WebRTC is used, it's going to measure the RTT thru the server, not the RTT of the WebRTC channel, so I just disabled it when WebRTC is in use.

PR #1204, opened 2020-04-20
Fix plugin version compare for new semver numbers

ryanblenis:
Caught some plugin version checks failing with the new semver versions (which I love, btw). Thank you!


Ylianst:
Nice, will be in MeshCentral v0.5.10 later today! Thanks.

Ylianst:
MeshCentral v0.5.10 is published!

Issue #1336, opened 2020-05-12
BUG: Faulty minifying renders v0.5.22+ unusable if "Minify" is set to true (v0.5.31)


veitw:
Hi!

A faulty minifying (if enabled using "Minify": true for the settings or a domain) renders MC unusable for any version newer than 0.5.21, including the current 0.5.31:

When the placeholder `{{{pluginHandler}}}` (which also remains unminified even if minifying is active) is expanded, the substitute ends with ";", resulting for default-min.handlebars in this construct:

        ... var ...,pluginHandlerBuilder=function() {
        ...
        return obj; };,pluginHandler=null; ...

The `;` appended to `function{}` within the declaration list causes a syntax error. As a result, anything beyond the login page appears just as a blank white page.

Best regards,
// Veit

veitw:
PR #1337 fixes this.

Noice number :D

Ylianst:
Nice. Thanks for the PR!! It will be in MeshCentral v0.5.32 when next published later today. Note that I don't do any testing with plug-in's and leave it up to @ryanblenis.

veitw:
Great, will update tomorrow. :)

I need to add that I also tried with all plugins in MC being disabled, resulting in the same problem. But the plugins were not completely removed, so they still showed up in the Plugins tab.

Unfortunately I did not have time to try with a completely fresh instance with empty meshcentral-data/ dir.

ryanblenis:
I hadn't usually tried with minification in my dev env, I'll have to add that to my test list. As for testing with plugins disabled, you can just change "plugins" to "_plugins" in the config and it will disable them all in one swoop while testing. Thanks for the find & PR!

veitw:
v0.5.32 runs fine ootb with plugins and minifying enabled. Thanks!

PR #1337, opened 2020-05-12
Fix minifying renders MC 0.5.22+ unusable if enabled #1336

veitw:
* pluginHandler.js: Removed superfluous ; to fix faulty substitution in minifying mode


Ylianst:
Thanks, this will be in MeshCentral v0.5.32 when next published later today.

PR #1397, opened 2020-05-22
Meshcmd Updates

ryanblenis:
1. Running from the command line does not currently appear to support the serverUrl variable (thus requiring a meshaction.txt file). This fixes that.

2. Allow a generated authCookie to serve as a credentialed logon via meshcmd, versus using plaintext username/password for the routing function.

Note: Apologies in advance. For some reason GitHub thinks I rewrote the entire file on the second commit. Viewing in a local client appears to produce only the ~12 changes lines. 🤷‍♂️

### DIFF
--- a/agents/meshcmd.js
+++ b/agents/meshcmd.js
@@ -126,6 +126,7 @@ function run(argv) {
 
     // Set the arguments
     if ((typeof args.action) == 'string') { settings.action = args.action; }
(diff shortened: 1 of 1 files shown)


Ylianst:
Thanks you! Looks great and excellent additions. This will be in MeshCentral v0.5.40 when I next publish it.

Issue #1227, opened 2020-04-23
Unknown Agents Added to Server


PetieM:
I'm not sure how much detail should be provided in public for this particular issue but I did think it warranted reporting regardless. This morning, two agents were added to my server that I do not recognize. Each was online briefly and has not been seen since so it's possible the owners of those machines saw the issue as well and uninstalled the agent themselves. According to IP addresses, both agents were added from different countries but that's about all I can see. Does anyone have any idea what's going on here?

Ylianst:
That is weird. Obviously, not sure about the specifics here but agents don't get at rights to the server or to other agents, so there should be no benefits since agents grant rights to the server, but get to rights back. Maybe if an agent is used behind a privacy VPN, you may see them show up from other countries.

ryanblenis:
Perhaps adding a default robots.txt to the project with a content of:
```
User-agent: *
Disallow: *
```
would prevent some of these reports? Running a search on Google I can find a few MC installs that should probably not be indexed.

PetieM:
That definitely wouldn't hurt but just landing on the login page shouldn't be enough to get you an agent installer, should it? I'm not concerned about security in this case since really, the only one who is at risk is the stranger installing remote management software on their machine (correct? - I imagine a computer with the agent software installed should have no access back to the server or any other agents) but I'd still like to get to the bottom of what happened here.

My server has only been up for a week or so and the group these two agents were added to was only created a few days ago so it seems unlikely this is just an accidental VPN mixup.

Ylianst:
@ryanblenis I did not think of the robot.txt file, that is a good idea.
@PetieM Correct, just accessing the portal is not sufficient. If one of the machines you manage has been hacked and the meshagent.exe/meshagent.msh where obtained, they could run it and connect back to your server. In any case, they would not gain any access, but would grant you access to their system.

ryanblenis:
if you go to https://yoururl.com/meshagents you'll get the agent download versions for your server. I'd imagine the case here is that your server is available via IP directly (vs hostname based directive) and you simply got scanned for port 80/443 and they found the login page and ended up at the meshagents page. Perhaps a honeypot that downloads executables and runs them in a sandbox environment- which is why they were only online briefly.

You're correct that they're the ones at risk since you technically control them at that point.

PetieM:
[...]

PR #1398, opened 2020-05-22
Add robots.txt to block search engines

ryanblenis:
As much as I enjoy googling to find everyone running MeshCentral servers and see how outdated some of them are... they probably shouldn't be listed!

Also referenced [here](https://github.com/Ylianst/MeshCentral/issues/1227#issuecomment-618667608), but figured I'd just issue a PR for it in case it was forgotten about!

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /


Ylianst:
Excellent, this is long overdue! Thanks you.

PR #1405, opened 2020-05-25
Fix plugin refresh failure due to PR #1337

ryanblenis:
Looks like PR #1337 caused an error in the refresh javascript when adding/removing plugins. This fixes that change while keeping the minification fixes.


Ylianst:
Thanks, this will be in MeshCentral v0.5.45.

PR #1414, opened 2020-05-26
Intial ADFS Quick Doc

MordyT:


MordyT:
Yes, I need to clean this up a lot, but wanted to get the initial doc created before I return to work tomorrow and forget about it. There's still another issue that I'm debugging as well.

PR #1593, opened 2020-07-09
Fixed: Testing for a valid filename on a folder path causes plugins w…

ryanblenis:
…ith views to fail to load associated views

Looks like commit 7f75ae419cee8fc7357d96ab93360a3b7826e38d added some validation checks, including this folder path with a file name validity check (causing a failure since it is a folder).

Not sure if there's a new built-in folder validation you'd want to use instead or if IsFilenameValid() should be updated for folders as well, but I saw [this](https://www.reddit.com/r/MeshCentral/comments/hn5fs0/plugin_unauthorized/) post with the issue and verified it.

Thank you as always!


Ylianst:
Oh, I see the problem. Thanks for the pull request. I will need to fix this again because for security, you can't allow ".." in paths, otherwise you can get access to things outside the expected folder.

PR #1624, opened 2020-07-19
Use bind address (if present) when checking redir port

bartnv:
The CheckListenPort() function of redirserver.js was not using the optional bind address ("redirPortBind" in config.json), causing it to change the port number when unable to bind on all interfaces. Correct this by passing in 'addr' as the second parameter to createServer.listen().


Ylianst:
Thanks. This will be in MeshCentral v0.5.89.

PR #1672, opened 2020-08-01
Dutch language update 0.6.0

petervanv:


Ylianst:
Thank you! I removed the `"</i></div>"` string that should not have been marked for translation. Thanks.

PR #1697, opened 2020-08-11
Pull Request On Final edits

Re9iNee:
modified DOM Element to prompt the user by default
removed Terminal Section from consents
add 2 other Desktop consents (show toolbar connection & notify user) to the consent section


Ylianst:
Can you explain the background of why this is a good pull request? These seem like personal customization of the web application that would not apply to everyone else. Am I missing something?

Re9iNee:
> 
> 
> Can you explain the background of why this is a good pull request? These seem like personal customization of the web application that would not apply to everyone else. Am I missing something?

I've accidentally sent P.R. sorry!! You're right I'm Customizing this for myself.

PR #1707, opened 2020-08-13
Add check node.agent!=null to fix AMT only broken view.

jsastriawan:
This small patch fix broken view on AMT only device. On AMT only device (CIRA or Direct) node.agent property does not exists thus it will throws exception which in turn break view rendering.


Ylianst:
Outstanding fix, thank you so much!

PR #1735, opened 2020-08-20
Relaxing check for HW Connect button for AMT only device group.

jsastriawan:
Currently the node.intelamt metadata collected by MeshAgent contain ver, sku and other attributes. HW Connect buttons' visibility is currently determined by those attributes. However, node.intelamt metadata for AMT only device group does not have that attributes. This prevents the button which rightfully should be visible to be invisible for the device. This patch relaxes checking on AMT only devices and maintains existing checking for MeshAgent-managed device group.


Ylianst:
Nice thank you! This is great!

PR #1887, opened 2020-10-10
added matrix to social link

tiny6996:
added matrix to social link


Ylianst:
Arg. It's completely ok with anyone creating a chat forum for MeshCentral on any platform, but this pull request is basically asking me to make this specific one "official". I don't know much about matrix and permissions granted, etc. So, I can't really approve this. I just looked into it a little and created a MeshCentral community, etc. I will work on it a bit more and may send out a different link.

Issue #1776, opened 2020-09-03
LDAP Binding fields to users in MeshCentral


LFM8787:
I love MeshCentral, since in my organization we use Active Directory as the authentication method, it would be great to be able to use the data currently uploaded to Active Directory to fill in the fields for each user automatically.

This is the case for GLPI, each field is bind to an actual Active Directory (ldap) field:
![image](https://user-images.githubusercontent.com/13755139/92158650-9ed00680-ee02-11ea-90c2-cce528a63d9e.png)

Since a lot of companies uses ldap for their users it could be populated automatically:
![image](https://user-images.githubusercontent.com/13755139/92159235-75fc4100-ee03-11ea-964c-fae8c81db6a8.png)

Ylianst:
This is a very good request. I don't have much knowledge of LDAP, but I think I can give this request a try.

Ylianst:
Looking at the code, can you try this line in your config.json at the same level at "auth":"ldap":

```
      "auth": "ldap",
      "LDAPUserEmail": "mail"          <--------
```

This should indicate to MeshCentral to grab the email address from "mail". let me know if this works, I will add values for "RealName" and "Phone" since I don't have these yet. The "Phone" is just in case you are using a SMS provider and want to support SMS 2FA.

Thanks,
Ylian

LFM8787:
> Looking at the code, can you try this line in your config.json at the same level at "auth":"ldap":
> 
> ```
>       "auth": "ldap",
>       "LDAPUserEmail": "mail"          <--------
> ```
> 
> This should indicate to MeshCentral to grab the email address from "mail". let me know if this works, If it does, I will add values for "RealName" and "Phone" since I don't have these yet. "Phone" is just in case you are using a SMS provider and want to support SMS 2FA.
> 
> Thanks,
> Ylian

Thanks for the fast response, I've tested some combinations but none of them seem to work.

**Test 1:**
```
         "Auth": "sspi",
         "Auth": "ldap",
         "LDAPUserEmail": "mail",
```
Since I use sspi as auth method, try to keep this in config file, server service didn't start.

**Test 2:**
```
         "Auth": "sspi",
         "Auth": "_ldap",
         "LDAPUserEmail": "mail",
```
Server service starts but didn't fill the email field, tested with existing and new accounts.

**Test 3:**
```
         "Auth": "_sspi",
         "Auth": "ldap",
         "LDAPUserEmail": "mail",
```
[...]

Issue #336, opened 2019-07-03
Not able to assign users to groups


treicadani:
Using meshcentral2 with ldap (ad) authentication, when i try to assign a user to a grup it fails with the message "Device Group: User --username-- not found";

I debugged the code and it seems in meshuser.js at line 1522 builds a newuserid with user/domain/username, then assign it to newuser. but parent.users has user//longbinarycode as identifier.

So it tries to bind user//username to user//longbinarycode which will never match

Ylianst:
Arg. This is a problem I need to work on. yes, the mapping is not right.

treicadani:
I have digged a bit into this and i found in webserver.js at line 356 is
var userid = 'user/' + domain.id + '/' + shortname;

Where shortname is the long binary code, if i change it into
var userid = 'user/' + domain.id + '/' + xxuser.sAMAccountName;

It works as intended, i can assign user to groups etc... however, i don't know if anything else is affected by changing the identifier, but i haven't find anything that breaks yet;

Maybe a global setting in config to set what type of identifier to use will fix it, but this will need to be implemented across the app where is required.

Ylianst:
My intention is to convert the user name to the SAMAccoutName when in LDAP mode. I need to look into this, however I am swamped with feature requests and so, apologies that I can't get to it quickly.

routerino:
Hi @Ylianst 

I've also encountered this problem. For now, an easy fix will be to remove the character length cap on the "add user to group" section. I suspect you can add a user by their ID, but you get cut short if you try to paste in the whole ID.

treicadani:
@routerino 

An easy fix to add users to groups is to edit to make it friendly for "Non technical peope" is to edit the source code in webserver.js and edit the two lines containing 

    var userid = 'user/' + domain.id + '/' + shortname;

and change it to 

    var userid = 'user/' + domain.id + '/' + xxuser.sAMAccountName;

As for the length when @Ylianst  sees it, he may be able to default it to a higher limit and also can use the code stated above to make it as a temporary fix to clear the bug and be upgraded/modified with the right fix when he has more time for this.

Ylianst:
Your right, this is something I need to fix ASAP. I will look at this wen I get back next week.

routerino:
Hi Ylianst, is there any traction on this change? As far as I can tell I still need to enter the 56 digit ID and manually change the HTML input box to add a user.

routerino:
[...]

PR #1931, opened 2020-10-30
Default userid to sAMAccountName on AD based LDAP

routerino:
On an active directory based LDAP system, sAMAccountName is the unique identifier for active directory. This will always be the preferred ID to use when using Active Directory.

### DIFF
--- a/webserver.js
+++ b/webserver.js
@@ -345,7 +345,10 @@ module.exports.CreateWebServer = function (parent, db, args, certificates) {
                     var username = xxuser['displayName'];
                     if (domain.ldapusername) { username = xxuser[domain.ldapusername]; }
                     var shortname = null;
-                    if (domain.ldapuserbinarykey) {
+                    // if we are binding to Active Directory, set our userid to the sAMAccountName
+                    if(xxuser.sAMAccountName) {
+                        shortname = xxuser.sAMAccountName
+                    } else if (domain.ldapuserbinarykey) {
                         // Use a binary key as the userid
                         if (xxuser[domain.ldapuserbinarykey]) { shortname = Buffer.from(xxuser[domain.ldapuserbinarykey], 'binary').toString('hex'); }
                     } else if (domain.ldapuserkey) {


routerino:
This is a fix for https://github.com/Ylianst/MeshCentral/issues/336. Also in reference to https://www.reddit.com/r/MeshCentral/comments/jj5f2i/share_device_groups_with_ad_logins/.

Needs testing, will test myself once merged (can't easily recreate my docker infrastructure to use my fork).

routerino:
Also #1776  which seems to be a repeat of #336

routerino:
I should also mention that this will be a breaking change for active directory users, IE: their user ID is going to change. This may or may not be a problem as it's not really in a super functional place to begin with.

Ylianst:
Hi. Thanks for the pull request. So, you should be able to test this by making the changes to your webserver.js file in your docker container. Obviously, I don't want to take breaking untested changes and put them into production so that someone can test it.

The other issue here is that you made a change in the LDAP testing section, this code section is only run when `domain.ldapoptions.url == 'test'`, so, I am not sure your changing the correct place.

Lastly, you can go in the config.json and set "ldapuserkey" to "sAMAccountName" and it would have the same effect as the change you made. Since you put the change as the first if in the chain, your saying that if xxuser contain "sAMAccountName" use that and ignore what the administrator configured. I would want administrator configuration to come first at a minimum.

Hope that makes sense?

routerino:
Thanks @Ylianst for the feedback. In hindsight that was a fairly low effort pull request, I can do better (or at least set up a separate docker instance for testing).

I've been previously resolving this with a sed replace, which makes sense why I didn't notice that there was a testing section (since it would have just replaced all instances indiscriminately). I'm not sure when `ldapuserkey` got introduced, but I would have to imagine it was after me resorting to sed replaces (or maybe I just wasn't being observant). Either way, introducing it as an option is certainly a superior choice.

I've tested the ldapuserkey setting and it's functional. Possibly we can update and close the two issues mentioned?

dinger1986:
> Thanks @Ylianst for the feedback. In hindsight that was a fairly low effort pull request, I can do better (or at least set up a separate docker instance for testing).
> 
> I've been previously resolving this with a sed replace, which makes sense why I didn't notice that there was a testing section (since it would have just replaced all instances indiscriminately). I'm not sure when `ldapuserkey` got introduced, but I would have to imagine it was after me resorting to sed replaces (or maybe I just wasn't being observant). Either way, introducing it as an option is certainly a superior choice.
> 
> I've tested the ldapuserkey setting and it's functional. Possibly we can update and close the two issues mentioned?
> 
> EDIT: took a look at the source code history, it definitely did get added sometime after my last update on the issue. I feel a little less silly :). Just wasn't aware the problem was resolved.

are you going to have a look at this again?

si458:
@dinger1986 i agree with @Ylianst above

> you can go in the config.json and set "ldapuserkey" to "sAMAccountName" and it would have the same effect as the change you made.

Issue #1901, opened 2020-10-19
Feature Request - curl instructions


map7:
Some of the computers I've added to MeshCentral don't have wget and cannot be installed, but they do have 'curl'. I got meshagent installed with the following;

`curl -k -o meshagent "<https address>"`

My request is that the curl instructions be displayed as an alternative to wget when adding/inviting an agent.

b8two:
@map7, I agree and posted a suggestion [965#issuecomment-708971952](https://github.com/Ylianst/MeshCentral/issues/965#issuecomment-708971952) to resolve the inconsistancies.

si458:
closing as stale

Issue #965, opened 2020-02-21
Minimum Linux environment requirement for Mesh client?


b8two:
I have some Linux based modem/routers, I was wondering if it was possible to install the Mesh client on to them. 
I've tried the linux install script on one but it failed with a Bash error.

krayon007:
I'm actually working on updating the installer, to make it uniform across platforms, so that you can do a -fullinstall on any platform.

The install script is a little convoluted because it tries to determine what platform, OS, and cpu you are using, so it can fetch the appropriate binary. But if you already know which binary you need, you'll be able to just run the binary directly to install it. (When I'm done with the update that is)

krayon007:
Do you know what soc your router uses? Was the bash error related to not knowing which soc you have?

b8two:
Hi Kryon007,

While you are working on the "meshinstall.sh", the Script is called with wget parameter --no-check-certificate, however internal to the script, this parameter is not used. When I was installing on our Proxmox Linux package (Debian based), I needed to modify the script to add in --no-check-certificate for each wget command.

The two devices we use are NTC-140 and now RUT240. 

But I only tried the NTC-140 so far which results with;

2020-02-22 13:49:35 (58.2 KB/s) - './meshinstall.sh' saved [9596/9596]
-sh: sudo: not found
env: can't execute 'bash': No such file or directory
root:~#

Hence the script didn't start to run at all.

Ylianst:
You can get a list of the available MeshAgents by going to "/meshagents" on your server. For example you can get the list here: [https://meshcentral.com/meshagents](https://meshcentral.com/meshagents). You can see a list of binaries and download links.

Basically, you could try to download the agent that matches the architecture and OS or your device and see if it runs.

If there is no agent that run on your device, the agent will need to be cross-compiled on it.

b8two:
Hi Ylianst,
Thanks for the URL for the MeshAgents, it has moved me a step further. 

Hi Kryon007,

I have found that I don't have BASH shell scripting, however SH shell scripting is avaliable. This is using buzybox.
I have an ARMv7 CPU, hence I started testing with the  [ARMv5 ](https://stackoverflow.com/questions/6771941/how-to-find-out-if-cpu-is-arm-v5-cpu-instructions-compatible) binarary.
After downloading this binrary I was able to launch it, after which I loose control of the shell. On another instance of the shell I check with ps and can see that it is still launched. Disconnecting the first shell also closes the mesh process.

[...]

PR #1975, opened 2020-11-11
Update meshinstall-linux.sh

b8two:
Feature Request  #1901  & part of #965

wget & curl usage ignoring certificates.
This is less secure but a typical road block to installing on Linux if certificates are self signed or if certificate has an incomplete authority chain.


Ylianst:
So, not sure I understand this. You added `--no-check-certificate` to the script, but the way MeshCentral currently works, this option is already added in {{{wgetoptionshttps}}} when MeshCentral uses it's default certificate. If you use a Let's Encrypt or different certificate, this option is removed as it's assumed your going to be using a trusted certificate.

The change you made will show '--no-check-certificate --no-check-certificate', that is this option will be shown twice when built-in cert is used and once when a different cert is used.

Let me know if I got that right?

PR #2107, opened 2020-12-24
Spelling fix

dev-id:
Fixed "avaialble" to "available" in 3 user (admin?) facing areas.


Ylianst:
Thanks! Will be in MeshCentral v0.7.32 when it comes out.

PR #2163, opened 2021-01-12
Prevent reverse proxy 5xx codes from triggering a page reload

TotallyNotElite:
Prior to this fix, the MeshCentral page would automatically reload itself after 10 seconds even if MeshCentral is still down when behind a reverse proxy like NGINX.


Ylianst:
Interesting! Looks good. Thanks.

PR #2335, opened 2021-03-02
Updated German translation

veitw:
* fixed mistranslations introduced by automatic translation
* fixed out-of-context misinterpretations


Ylianst:
Thanks you! This will be in MeshCentral v0.7.80 when next published.

veitw:
Thanks, looks good in 0.7.80. :)

I already found two other contextual mistranslations. I will collect some and create another PR.

PR #2404, opened 2021-03-20
Dutch language update 0.7.91

petervanv:
question: what are "file files" can't translate it because the meaning is unknown


Ylianst:
Thank you!

PR #2426, opened 2021-03-26
Fix MeshCmd AMTPower poweroff action to use supported power state.

jsastriawan:
Minor update to fix poweroff NOT READY because of unsupported power off state.


Ylianst:
Excellent! Thank you.

Issue #2486, opened 2021-04-07
Fails to pass backupCapability check if mongodb is on remote server


nzalev:
When hosting MongoDB on a remote server, meshcentral will fail to pass the `checkBackupCapability` test in `db.js`.

The following is the offending section (starting at line#1593 in db.js)
```javascript
            var mongoDumpPath = 'mongodump';
            if (parent.config.settings.autobackup && parent.config.settings.autobackup.mongodumppath) { mongoDumpPath = parent.config.settings.autobackup.mongodumppath; }
            const child_process = require('child_process');
            child_process.exec('"' + mongoDumpPath + '"', { cwd: backupPath }, function (error, stdout, stderr) {
                try {
                    if ((error != null) && (error != '')) {
                        if (parent.platform == 'win32') {
                            func(1, "Unable to find mongodump.exe, MongoDB database auto-backup will not be performed.");
                        } else {
                            func(1, "Unable to find mongodump, MongoDB database auto-backup will not be performed.");
                        }
```
By default, mongodump will eventually time out. If one concatenates error to the end of the error message, the following would be printed: 
```
WARNING: Unable to find mongodump, MongoDB database auto-backup will not be performed.Error: Command failed: "mongodump"
2021-04-07T17:22:07.148+0000    Failed: can't create session: could not connect to server: server selection error: server selection timeout, current topology: { Type: Single, Servers: [{ Addr:
 localhost:27017, Type: Unknown, State: Connected, Average RTT: 0, Last error: connection() : dial tcp [::1]:27017: connect: connection refused }, ] }
```

One *can* include the --uri option in the `mongoDumpPath` string under the `autoBackup` settings. However, doing so would conflict with the command string when actually performing the backup under `performBackup`, as it already includes the --uri option.
```javascript
                var mongoDumpPath = 'mongodump';
                if (parent.config.settings.autobackup && parent.config.settings.autobackup.mongodumppath) { mongoDumpPath = parent.config.settings.autobackup.mongodumppath; }
                const child_process = require('child_process');
                var cmd = '\"' + mongoDumpPath + '\" --db=\"' + dbname + '\" --archive=\"' + newBackupPath + '.archive\"';
                if (dburl) { cmd = '\"' + mongoDumpPath + '\" --uri=\"' + dburl.replace('?', '/?') + '\" --archive=\"' + newBackupPath + '.archive\"'; }
```

<br><br>

[...]

PR #2488, opened 2021-04-07
Fix for #2486

nzalev:
if dburl then use --uri option


Ylianst:
Excellent, thank you. This will be in MeshCentral v0.8.5 later today.

PR #2503, opened 2021-04-11
Add mariadb/mysql db dumps to autobackups

nzalev:
I have added mysql/mariadb database dumps to the automatic backups.

There are three main areas added:

1. Added a check for mysqldump to checkBackupCapability
2. Added db dump + archive for mysql and mariadb databases in performBackup
3. Added mysqldump path to config schema

I have tested MariaDB on linux and windows. MySQL has the same syntax.

<br>
<br>
PS:
The logic of performing db backups is very similar between MongoDB and MariaDB/MySQL. The overall structure is the same:

1. set up the dump command
2. run child process
3. perform archiving in callback function

If you want, I can refactor the db backups in a future commit (both mongodb and mariadb/mysql) since the only real difference is in setting up the dump command.

### DIFF
--- a/db.js
+++ b/db.js
@@ -1808,6 +1808,48 @@ module.exports.CreateDB = function (parent, func) {
                         archive.finalize();
                     } catch (ex) { console.log(ex); }
                 });
+            } else if ((obj.databaseType == 4) || (obj.databaseType == 5)) {
+                // Perform a MySqlDump backup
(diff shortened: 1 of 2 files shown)


Ylianst:
OMG! Thank you! That is amazing. I think you just make MySQL/MariaDB on the same level as MongoDB which is great! This will be in the next version of MeshCentral.

PR #2527, opened 2021-04-19
Fixed log.txt permission bits

nzalev:
The permission bits must be expressed in octal.

Old behaviour
<pre>--w--wx--T 1 noah_z noah_z   99 Apr 19 12:55 <font color="#4CE64C"><b>log.txt</b></font></pre>

New behaviour
<pre>-rw-rw-r-- 1 noah_z noah_z   99 Apr 19 12:59 log.txt</pre>


Ylianst:
Thank you!

PR #2538, opened 2021-04-23
Support different db names

nzalev:
I did not have time last night, but today I have added alternate mysql/mariadb db name support to autobackups as well


Ylianst:
Excellent! Thank you!

PR #2555, opened 2021-04-28
Replacing public with private.key filename on page 5 in user guide.

uhofemeier:
I noticed the filename in the user guide for integrating legitimate trusted certs isn't quite right. the current documentation wants one to replace "webserver-cert-public.key" and this file doesn't exist in the meshcentral-data directory. The file that has to be replaced is "webserver-cert-private.key." I updated the documentation accordingly. 
Signed-off-by: Ulf Hofemeier <ulf.hofemeier@intel.com>


Ylianst:
Thank you!

Ylianst:
Just updated the MeshCentral.com downloads section with this fix.

Issue #2557, opened 2021-04-28
Connecting to remote MySQL database with enforced TLS


D4V3M0NK:
Can MC2 connect to a remote database with enforced TLS? 

(When I connect using the mysql client, I can use the `--defaults-extra-file` option to contain my connection parameters, including the path to the certificate I have to present, along with the SSL mode, which I've provided for context in the hope that it's useful:

<details>
<summary>mysql.params</summary>

```
[client]
host = name-of-my-remote-mysql-host
user = mc2
password = "mc2_user_password"
ssl-ca = /location/of/certificate.bundle.pem
ssl-mode = VERIFY_IDENTITY
```
</details>

I then invoke `mysql --defaults-extra-file=mysql.params` to connect.)

From the complete `config.json` [options available](https://raw.githubusercontent.com/Ylianst/MeshCentral/master/meshcentral-config-schema.json) the MySQL section doesn't contain anything regarding connecting securely, but I seem to remember it being mentioned in prior discussions here, that it could be posisble?

D4V3M0NK:
Update : found the complete JSON options available can be found [here](https://raw.githubusercontent.com/Ylianst/MeshCentral/master/meshcentral-config-schema.json)

nzalev:
> Can MC2 connect to a remote database with enforced TLS?
> 
> (When I connect using the mysql client, I can use the `--defaults-extra-file` option to contain my connection parameters, including the path to the certificate I have to present, along with the SSL mode, which I've provided for context in the hope that it's useful:
> 
> mysql.params
> ```
> [client]
> host = name-of-my-remote-mysql-host
> user = mc2
> password = "mc2_user_password"
> ssl-ca = /location/of/certificate.bundle.pem
> ssl-mode = VERIFY_IDENTITY
> ```
> 
> I then invoke `mysql --defaults-extra-file=mysql.params` to connect.)
> 
> From the complete `config.json` [options available](https://raw.githubusercontent.com/Ylianst/MeshCentral/master/meshcentral-config-schema.json) the MySQL section doesn't contain anything regarding connecting securely, but I seem to remember it being mentioned in prior discussions here, that it could be posisble?

The MySql and MariaDB options were passed in directly to the sql connectors. In a recent change, I began creating a new options object because including something like a `database` (name) option would cause the connection to fail during the automatic db creation feature.

[...]

PR #2561, opened 2021-04-30
Improved connectionObject creation

nzalev:
Step 1 in adding good mariadb/mysql ssl support
this cleans up creating the connection object used by the mariadb and mysql connectors

This is enough to use trusted ca signed certificates.

I will be able to finish the rest (support for self-signed certs) by this weekend for sure, I think.


Ylianst:
Thank you. Will be in MeshCentral v0.8.25.

Ylianst:
MeshCentral v0.8.25 was published. Fix for #2557.

PR #2572, opened 2021-05-01
Added selfsigned cert support to mysql/mariadb ssl

nzalev:
Part 2 of mysql/mariadb connector SSL support

This adds support for using self-signed certificates at the database server. The needed CA certificate can be loaded off disk via the `caCertPath` ssl option (added to config schema). Currently, this uses an absolute path.

This also adds support for two-way ssl authentication, by giving the option to load a client certificate + key off the disk as well.


Ylianst:
Nice! I like that you updated the schema. Very well done. Will be in the next version.

PR #2575, opened 2021-05-01
Mysql/mariadb ssl support

nzalev:
Part 3 of SSL support and other bugfixes

This pull introduces a number of things:

1. dedicated function to build the sqldump command (with more and more options, this becomes complex enough to warrant its own function)
2. adds ssl support to mysqldump
3. fixes inappropriately checking properties of `mysql.ssl` / `mysql.ssl` when `ssl` does not exist (cannot access a property of undefined - my bad)
4. don't rely on `use database` - although this worked on all the versions I tested, it does not work on *all* versions

What's going on with 4)
If the `database` parameter is set in the connection object, the connection will fail if the db does not exist - this makes the automatic db creation feature unusable. To keep this feature working, the `database` must be removed from the connObj. 

`USE database` gets around this for most db versions - however, while expanding the number of test servers to experiment with ssl, I found `USE database` does not work on all versions.

The next best thing I could think of is to connect to the server without a db specified, create db if not exists, and disconnect. Then re-add the database parameter to the connection object, and continue normally. This is obviously a lot more complex, and what I was trying to avoid the first time.

### DIFF
--- a/db.js
+++ b/db.js
@@ -438,37 +438,8 @@ module.exports.CreateDB = function (parent, func) {
         obj.dbRecordsDecryptKey = parent.crypto.createHash('sha384').update(parent.args.dbrecordsdecryptkey).digest('raw').slice(0, 32);
     }
 
-    if (parent.args.mariadb || parent.args.mysql) {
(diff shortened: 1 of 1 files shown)


Ylianst:
Thanks is wonderful. Thank you so much for making these improvements.

PR #2577, opened 2021-05-02
Mysql/mariadb ssl support - additional options

nzalev:
Added last needed options to mysqldump to support 2-way authentication
Added `dontCheckServerIdentity` option to disable hostname checking during certificate verification. Useful for self-signed certs where the hostname on the cert may not match.


Ylianst:
This is wonderful. Thank you.

Ylianst:
@nzalev Could you take a look at #2659 and see if it makes sense? Thanks.

nzalev:
> @nzalev Could you take a look at #2659 and see if it makes sense? Thanks.

Sure - I will comment on that pull request though

PR #2578, opened 2021-05-03
refactor: rename function getCurrentVerion => getCurrentVersion

nicolasduteil:
PR just renames local function from `getCurrentVerion` to `getCurrentVersion` for consistency


Ylianst:
How did I not notice that! Thanks.

PR #2598, opened 2021-05-07
Bugfix: MySQL/MariaDB crash when deleting groups

nzalev:
Deleting a device group while using MySQL/MariaDB causes Meshcentral to crash. This bug goes back to the original implementation - surprised it hasn't come up before.

The callback function used in `CreateDB` gets passed around throughout db.js, until it is eventually re-called in `RemoveMeshDocuments`.
bug

This results in the hilariously useless stack trace:
<pre>SQLERR1 TypeError: Cannot read property &apos;SetupDatabase&apos; of null
    at /home/noah_zalev/Documents/meshc_testing/node_modules/meshcentral/meshcentral.js:717:24
    at /home/noah_zalev/Documents/meshc_testing/node_modules/meshcentral/db.js:898:45
    at process._tickCallback (internal/process/next_tick.js:68:7)
</pre>

It was very hard to find, but the fix is very simple :).


Ylianst:
Oh my! I don't use MySQL or MariaDB much. I to am surprised that was never reported. Thank you!

PR #2604, opened 2021-05-08
Create haproxy-with-client-certs.cfg.example

Bobberty:
This is a suggested addition to the Doc folder.  It provides an examples directory as well as a haproxy config that adds some features without modifying the core of MeshCentral.

In my specific usage case, I have multiple clients with multiple systems managed. Usernames and passwords have been distributed to key site users.  I have further distributed Client Certs.  This provides a restriction to who can access the webserver with less worry of weak passwords.  Further, the agent urls don't require a user login and are easy to remember.

Example file for Haproxy that adds agent direct downloads, splitting of IPv4/IPv6 and validated client certificates to MeshCentral.  This does not override the need for a username/password.  It simply provides a layer for authorization.


Ylianst:
This is excellent! Great idea. Thank you.

PR #2659, opened 2021-05-19
Update db.js

narwahls:
The Slash before the additional parameters inside the mongoDB connectionstring gave us great headaches.
After the AutoBackup initiated it immediately fails as it couldn't find the database with a trailing slash.
Our connectionstring contains information for the local MongoDB User and the authentication DB.
The String looks as follows (user and password have been omitted):
mongodb://{{user}}:{{password}}@localhost:27017/meshcentral?authSource=admin&compressors=zlib&gssapiServiceName=mongodb

### DIFF
--- a/db.js
+++ b/db.js
@@ -1656,7 +1656,7 @@ module.exports.CreateDB = function (parent, func) {
             var mongoDumpPath = 'mongodump';
             if (parent.config.settings.autobackup && parent.config.settings.autobackup.mongodumppath) { mongoDumpPath = parent.config.settings.autobackup.mongodumppath; }
             var cmd = '"' + mongoDumpPath + '"';
-            if (dburl) { cmd = '\"' + mongoDumpPath + '\" --uri=\"' + dburl.replace('?', '/?') + '\"'; }
+            if (dburl) { cmd = '\"' + mongoDumpPath + '\" --uri=\"' + dburl + '\"'; }
             cmd += (parent.platform == 'win32') ? ' --archive=\"nul\"' : ' --archive=\"/dev/null\"';
             const child_process = require('child_process');
             child_process.exec(cmd, { cwd: backupPath }, function (error, stdout, stderr) {
@@ -1834,7 +1834,7 @@ module.exports.CreateDB = function (parent, func) {
                 if (parent.config.settings.autobackup && parent.config.settings.autobackup.mongodumppath) { mongoDumpPath = parent.config.settings.autobackup.mongodumppath; }
                 const child_process = require('child_process');
                 var cmd = '\"' + mongoDumpPath + '\" --db=\"' + dbname + '\" --archive=\"' + newBackupPath + '.archive\"';
-                if (dburl) { cmd = '\"' + mongoDumpPath + '\" --uri=\"' + dburl.replace('?', '/?') + '\" --archive=\"' + newBackupPath + '.archive\"'; }
+                if (dburl) { cmd = '\"' + mongoDumpPath + '\" --uri=\"' + dburl + '\" --archive=\"' + newBackupPath + '.archive\"'; }
                 var backupProcess = child_process.exec(cmd, { cwd: backupPath }, function (error, stdout, stderr) {
                     try {
                         var mongoDumpSuccess = true;


Ylianst:
What operating system are you using to run the MeshCentral server? I have not written this code, but I suspect that the problem may be OS specific and don't want to break anyone else with the change. Thanks.

nzalev:
> What operating system are you using to run the MeshCentral server? I have not written this code, but I suspect that the problem may be OS specific and don't want to break anyone else with the change. Thanks.
> 
> Would like to have @nzalev command on this pull request.

When I wrote this code, I did it based on the existing mongodump command that was there. I last modified this section in #2488, which allowed the mongodump check to use the uri option - however, I took and re-used the existing command.

Coincidentally, I know exactly when this .replace() was added because I noticed it the last time I performed an upgrade on our production system: https://github.com/Ylianst/MeshCentral/commit/cce68922d798b892a517407e0139724ab7e9e03b

Ylianst:
Thanks done.

Ylianst:
On Windows, I am now getting this error:

```
ERROR: Unable to perform MongoDB backup: Error: Command failed: "C:\Program Files\MongoDB\Tools\100\bin\mongodump.exe" --uri="mongodb://127.0.0.1:27017?replicaSet=rs0" --archive="C:\Users\Default.DESKTOP-9CGK2DI\Desktop\AmtWebApp\meshcentral-backups\mongodump-2021-05-20-13-17.archive"
2021-05-20T13:17:41.717-0700    error parsing command line options: error parsing uri: must have a / before the query ?
2021-05-20T13:17:41.717-0700    try 'mongodump --help' for more information
```

nzalev:
> On Windows, I am now getting this error:
> 
> ```
> ERROR: Unable to perform MongoDB backup: Error: Command failed: "C:\Program Files\MongoDB\Tools\100\bin\mongodump.exe" --uri="mongodb://127.0.0.1:27017?replicaSet=rs0" --archive="C:\Users\Default.DESKTOP-9CGK2DI\Desktop\AmtWebApp\meshcentral-backups\mongodump-2021-05-20-13-17.archive"
> 2021-05-20T13:17:41.717-0700    error parsing command line options: error parsing uri: must have a / before the query ?
> 2021-05-20T13:17:41.717-0700    try 'mongodump --help' for more information
> ```

Ahh, I see now why you originally added the .replace().  But I think your connection string has an incorrect format: it should be `mongodb://127.0.0.1:27017***/***?replicaSet...`

How do you have it set in your config.json? I don't believe the connection string gets modified between the config and its use here in db.js.

https://docs.mongodb.com/manual/reference/connection-string/

nzalev:
> Just put in another fix: [626364b](https://github.com/Ylianst/MeshCentral/commit/626364bab81fbac745ee9ec67d4f00631120744a)

There is an error in this update if the connection string includes the name of the database:
```mongodb://127.0.0.1:27017/meshcentral?compressors=zlib```
is returned as
```mongodb://127.0.0.1:27017/meshcentral/?compressors=zlib```
which contains a trailing / at the end of the database name.

I can fix this if you want, but I think the better option is to not modify the connection string given by the user. I think that will save some headaches.

**Comment by narwahls** [comment]
I just came around to answer your question.
Sorry for the delay.
We are using Ubuntu Server 18.04.
MongoDB is at 4.2.14.

The fix #626364b just made the original version a little bit fancier.
It seems that a trailing slash is just expected, when replicasets are used.
So how about something like this (db.js, line 1650)
```javascript
// MongoDB mongodump requires that the URL have a / at the end of the path. If not present, this will add it.
  function terminateUrlPathWithSlash(str) {
    const u = require('url').parse(str);
   return u.protocol + '//' + u.host + (u.pathname ? u.pathname : '') + (u.search.includes('replicaSet') ? '/' : '' ) + (u.search ? u.search : '');
}
```

nzalev:
> I just came around to answer your question.
> Sorry for the delay.
> We are using Ubuntu Server 18.04.
> MongoDB is at 4.2.14.
> 
> The fix [626364b](https://github.com/Ylianst/MeshCentral/commit/626364bab81fbac745ee9ec67d4f00631120744a) just made the original version a little bit fancier.
> It seems that a trailing slash is just expected, when replicasets are used.
> So how about something like this (db.js, line 1650)
> 
> ```js
> // MongoDB mongodump requires that the URL have a / at the end of the path. If not present, this will add it.
>   function terminateUrlPathWithSlash(str) {
>     const u = require('url').parse(str);
>    return u.protocol + '//' + u.host + (u.pathname ? u.pathname : '') + (u.search.includes('replicaSet') ? '/' : '' ) + (u.search ? u.search : '');
> }
> ```

This will not work in general either, even if it works properly if the connection string contains the replica set option. It will break just the same for any set of options which do not contain replica set, for sure.

The format of the connection string is

```
mongodb://[username:password@]host1[:port1][,...hostN[:portN]][/[defaultauthdb][?options]]
```

The important part is
```
[/[defaultauthdb][?options]]
```
The whole section is optional, and within it, the db name and options are both optional as well. However, if either or both are included, they must be preceeded with the /. So only then should we include a /.

Again, I think the best option is to not modify the connection string and to just make sure the user sets it properly. That would mean reverting https://github.com/Ylianst/MeshCentral/commit/626364bab81fbac745ee9ec67d4f00631120744a . I can also submit a fix by tonight, whichever @Ylianst prefers.

narwahls:
Ok.
It was just a thought.

As using MongoDB or another non standard Backend involves work by an user, leaving the validation of a connection string to them might be the best idea.
An user could just add the slash as needed.

Another thought, could it be OS specific? Ylianst seems to use Windows, as can be seen above (https://github.com/Ylianst/MeshCentral/pull/2659#issuecomment-845456225). I am using Ubuntu on my server for Meshcentral?

Ylianst:
Wow. Thank you for this work. I will not do any corrections, but will also display the MongoDB output if there is an error so users can see what is wrong and add the missing slash if needed. Let me check that is now.

PR #2808, opened 2021-06-23
Bugfixes

nzalev:
Simple bugfixes: fixed typo in variable name. Found another `for (i = 0;` in a file with `use strict`


nzalev:
I have added another commit. 

I saw you recently added (in 5e7669e6d88948c873b9e9d7325932f445bf5789) a check to see if `user` is null. However, `user` is not yet defined, so it will always be null. This bypasses the notools check.

I have moved the following above the notools check:
```javascript
var user = obj.users[userid];
```

Ylianst:
OMG!! Thank you so much!!!

PR #2866, opened 2021-07-07
Refactor mongodump command

nzalev:
This is a bit of code cleanup around making the mongodump command for auto-backups


Ylianst:
Nice! Looks wonderful, excellent update.

Issue #2868, opened 2021-07-07
Refactoring meshuser


nzalev:
This is an issue purely to discuss code. I was reading #1321 last night and was encouraged to offer some reorganization. 

meshuser is a perfect starting place for this. Currently, it is a 5000 line switch statement, with a nested switch statement to handle console commands. Using some form of lookup table is a cleaner way of doing this that's used in a lot of other large projects.

I will give you an example

```javascript
    var commands = {};
    commands['intersession'] = intersessionCommand;
    commands['authcookie'] = authcookieCommand;
    ....


    function intersessionCommand( ... ) { ... }
    function authcookieCommand( ... ) { ... }

    function processWebSocketData(msg) {
        ...

        try {
            commands[cmd]( ... );
        } catch (e) {
            if (commands[cmd] == null)
                console.log('Command not found');
                // Handle command not found here
            else
                console.log(e);
                // Generalized error handling
                // Can re-throw
        }
    }
```

The main reason to structure it this way is readability and navigability (maintainability). Currently, finding the command you want can mean constantly ctrl+f'ing or scrolling through thousands of lines.

Having the table (in js's case an associative array) of all commands and their handlers at the top gives you the ability to see all of them at a glance, and most IDE's will allow you to ctrl+click the function name to jump directly to the handler. From a philosophical view, it also separates out responsibility into more suitable chunks.

I do not want to refactor the whole thing to find out the PR will not be accepted, so I am making this issue first to guage what you think.

Ylianst:
I am ok with this. Let me refactor one switch statement and see how it goes.

nzalev:
Before you start, I will put in a PR and see what you think. I did serverconsole as a test. @Ylianst

Ylianst:
Arg. I was just doing my own refactor of this very same thing. Ok, now I have to figure out what to do.

nzalev:
> Arg. I was just doing my own refactor of this very same thing. Ok, now I have to figure out what to do.

I am surprised at how quickly you finished!

Ylianst:
Ok, that was not good. We did the same thing at the same time. My version is current, I am not going to touch it anymore. Can you take a look at it and if you don't like it, re-submit your version and I will accept it.

I would like all functions used for user commands to all start with the same prefix... like "serverUserCommandxxxxx" instead of starting with the command name "xxxxServerUserCommand'. If you order the function names, they order well. Also, I used the command table for the help command.

nzalev:
[...]

PR #2871, opened 2021-07-08
Refactor serverconsole

nzalev:
Please see Files Changed. This is related to #2868

I have refactored the nested switch statement for `'serverconsole'`. I have done it exactly as described in #2868 

I have seen improvement already in readability. There were actually two instances of `heapdump` in the server console switch statement. Very difficult to notice as it was, but much easier, given the command table.


Ylianst:
Arg. I just did my own version of this exact refactor.

PR #2874, opened 2021-07-08
Refactor meshuser cleanup

nzalev:
- Removes extra command table (from double merge last night)
- Removes extra command handlers (from double merge last night)
- Fixes typo in your command table
- I have used your command table, which includes the help messages

There were some issues with your command handler functions (undeclared `r`, break statements where there should be returns, use of out of scope variables), so I replaced them with the ones from my refactoring. I modified my help command handler to match what you made.

I have kept the `cmdData` object from my refactoring. This object is used to pass the command, arguments, and result to the command handlers. Some of the handlers need access to the command (such as heapdump), and some need access to the arguments (such as help), and some handlers need both (webpush). The `cmdData` object provides a unified way for all the handlers to get what they need (and avoids issues like superfluous arguments).

Having the result available to be modified avoids having to declare and return `r` in every handler.

#2868


Ylianst:
Looks perfect! Approved. Apologies for the trashing, I will not jump the gun next time.

PR #2885, opened 2021-07-10
Fix myserver errorlog, fix trailling comma

nzalev:
- schema shows option "ShowLog" when option in code is `errorlog`
- removes trailing comma


Ylianst:
Nice! Thank you!

PR #2928, opened 2021-07-21
Allow passing absolute path to configfile

lheckemann:
Sometimes, we don't want the path to the config file to be interpreted as relative to the data path.


Ylianst:
Thanks, that makes sense.

Issue #2929, opened 2021-07-21
how to get all events, not only 1000?


q16marvin:
How can i get events from last month for example? Currently via webpage i can only download last 1000 events... Thx!

VINISHVV:
If you are using MongoDB, you can use the "mongo" command to query the events collection of the meshcentral database.

Another option is through  **MeshCtrl.js** You can to pull the event log using command "listevents" .

q16marvin:
hi, i dont use mongodb... with MeshCtrl.js the first events i can see is from 2021-07-01T12:34:12.899Z but we use mesh far longer :)

VINISHVV:
Not sure, may be its due to your configurations / default meshcentral config to reduce the Database size.

Please add below commands under settings sections to increase the DB expiry.

"dbExpire": {
    "events": 31536000,
    "powerevents": 301400,
    "statsevents": 301400
    },

Ylianst:
There is currently no way to get more than 1000 events from the web page. You can use other tricks like going directly into MongoDB, but this is a feature I would like to add. Also, better filtering on the web page would also be very good.

Ylianst:
@nzalev pushed a fix for this with #2931. I will publish a new version on Saturday that includes it.

Ylianst:
Fix will be in MeshCentral v0.8.84.

PR #2931, opened 2021-07-21
Add no limit option to events,userevents dropdowns

nzalev:
Minor (incomplete) resolution for #2929

Both user events and general events have queries which exist without limits, so it is just a matter of adding an option in the dropdowns where `limit` is null.

This at least allows users to view all the events in the UI, though I agree more needs to be added. I can look at adding filters soon.


Ylianst:
Ha, this makes sense and it's such a easy addition. Nicely done!!!

Issue #2846, opened 2021-07-01
"Last seen" when hovering computer name


rodrigolzd:
In the computer list, currently we have to go into the pc details to see when it was last seen, perhaps a tool tip hover could help or even an extra column, but it would take up screen space. It’s mostly a debugging/troubleshooting information.

nzalev:
The last seen column is a hotly request feature, seen in many other issues.

I was working on it a few weeks ago, but I have since been occupied with my day job.

Sadly, it is not as simple as just adapting the existing last seen feature from the General tab in the PC details, but I will likely start working on it again soon.

VINISHVV:
> The last seen column is a hotly request feature, seen in many other issues.
> 
> I was working on it a few weeks ago, but I have since been occupied with my day job.
> 
> Sadly, it is not as simple as just adapting the existing last seen feature from the General tab in the PC details, but I will likely start working on it again soon.

Wow its a good news !!!
We are awaiting for this features, Hope this will include in export options too.....

Ylianst:
The hover over the computer name should be reserved for cases where the computer name is very long and does not fit available space. This said, a new column type is a great idea but it's not as easy as one would think. The last connected data is some other data in the database that is not normally sent to the browser unless requested. So, it's not part of the basic device data.

If your thinking of coding this, let me know what your thinking first. I guess we could do a batch request for this data for all devices you own.

nzalev:
>This said, a new column type is a great idea but it's not as easy as one would think. The last connected data is some other data in the database that is not normally sent to the browser unless requested. So, it's not part of the basic device data.

This is what I found while I was looking into it. 

For those who are interested, the existing query to get 'last seen' from the database is only for a single record. The client (browser) must send a request to the web server and the web server's request handler is designed for a single node. The client then receives the response and sets 'last seen' for that single node.

Using what currently exists for a new column, as a result, can cause thousands of requests to be sent to the server at once and can actualy DOS it, depending on the number of agents.

@Ylianst I have been planning on writing a batched query to pull all `lastconnect` records from the database, then write new request/response handlers to process all the docs at once. I am thinking that the data should be requested on page load, then probably on a timer (maybe once every 5 minutes).

rodrigolzd:
This can be closed now.

nzalev:
@rodrigolzd You have the power to close since you opened the issue.

Issue #2329, opened 2021-02-27
Add Custom / "Last Seen" column in "My Devices" view


L422Y:
Add "Last Seen" column into xdevices table view, better yet, make the columns customizable!  #enhancement

vitko-bg:
This is a duplicate :) sorry, cannot provide number of the issue, as on the go and on the phone :)

si458:
this already exists, click the COG in the columns view and select 'Last Seen'

Issue #1693, opened 2020-08-08
Feature Request: Add Last Seen Date Column to List View


johnnyq:
Please add Last Seen Date Column to List View .This would be fantastic way to easily sort out what computers have been offline for awhile especially when you have many agents. Currently I have to click on each agent to see the last seen date. Thank you!! =] and Excellent work!!!

johnnyq:
Even OS Would be helpful in the list view and LAN IP and rename Address column to WAN IP and add MAC Address and add action column with the following actions icon for remote desktop file manger terminal, details. Just some ideas throwing them out there. Thanks again

Ylianst:
What I really need to do is allow customization of the columns, there are requests for all sorts of information on that view and so, I need to have people sure setup the columns the way they want.

vitko-bg:
+1 for the above request :)

Mfinadmin:
> What I really need to do is allow customization of the columns, there are requests for all sorts of information on that view and so, I need to have people sure setup the columns the way they want.

@Ylianst - pls do if possible

vitko-bg:
@Ylianst Please please add the column "Last seen". Recently you added IP address to columns #2580 and a column with RDP/SSH links, but this issue is open from long time. I think lots of people will appreciate the option to have this column in detail view.
Big THANKS in advance :)

VINISHVV:
> What I really need to do is allow customization of the columns, there are requests for all sorts of information on that view and so, I need to have people sure setup the columns the way they want.

Please allow customization based on all available details...

si458:
this already exists, click the COG in the columns view and select 'Last Seen'

Issue #1137, opened 2020-04-10
Enhancement - Add Last seen in List view.


vitko-bg:
Hi,

I know that you are very busy with all big and small request. Sorry for adding additional work on the top of all. However, I think it will be very useful to have Last Seen information in the List View.

In this case will be easier to diagnose network problems in certain departments or remote offices - if all computers in one group are offline and all are went offline at the same time, that means that internet or network went down. Otherwise I have to click on every single computer to see Last Seen information.

See below:
![image](https://user-images.githubusercontent.com/59362521/78978344-81c42f80-7b11-11ea-8ef6-b362c8bff8e8.png)

I added Date and Time all the same - but it is for example only :)

Thanks!

ryanblenis:
Oddly enough the other night I just started an initial test-run of modifying the device page to have a "custom columns" format, where users could choose the columns to be displayed.

<img width="1646" alt="Screen Shot 2020-04-10 at 5 55 19 PM" src="https://user-images.githubusercontent.com/1929277/79025807-d76dfb80-7b54-11ea-8a51-8d30137dbe81.png">

It's early in the inception, and I need to find a way to load the associated/requested data in a sane manner without either a) grabbing everything possible; or b) making a ridiculous amount of individual queries for separate pieces of data. So I'll definitely be watching to see if @Ylianst implements your request and his preferred method of going about it.

Along with this concept, I also want to make the headers clickable for re-ordering versus the current dropdown of pre-defined sort options.

Ylianst:
I am hesitant to implement changes that are more preferences because the real ask is really to allow more customization of the user interface. Even if I change things in the UI, it will never make everyone happy and best to allow users to setup the UI the way they want it. @ryanblenis is exactly on the right track in my view. There are a lot of things on my todo list, but I would want a "devices" page that is a lot more customizable. Different people use MeshCentral is different ways, for IoT usages, current logged in user is not important at all, but for remote support, it is.

Ylianst:
Ok, not going to do this. However, future work on better customization of the device page would be interesting.

si458:
this has already been implemented, you click the COG on right hand side and select 'Last Seen'

PR #2939, opened 2021-07-25
Added Last Seen column to device view

nzalev:
I have finally added the Last Seen column to the device list view.

The main problem in adding a last seen column was that not all the `lastconnect` data was loaded. However, agents which connect/disconnect do update `node.lastconnect`. I am loading all the last seen data for a user's nodes on page load, and only once. When agents connect/disconnect, they already update the info on their own.

On the server side, the last seen command gets a user's nodes and creates an associative array using the node IDs. Then all `lastconnect` docs are pulled from the db, `lcnode//` is converted to `node//`, and the last connect time is inserted only if the ID already exists in the associative array. This avoids leaking any information about nodes the user does not have access to. This is effectively left-joining the LC times on to the user's nodes.

The UI changes are pretty simple. The client side handler for `lastseen` update's the `lastconnect` property of nodes in the list `nodes`.

Finally, I factored out the code that gets a user's extra node IDs into its own function, since I needed to re-use it.

Screenshot from 2021-07-25 01-20-50
Screenshot from 2021-07-25 01-20-35

### DIFF
--- a/meshuser.js
+++ b/meshuser.js
@@ -5554,12 +5554,11 @@ module.exports.CreateMeshUser = function (parent, db, ws, req, args, domain, use
                 for (var j in docs) {
                     var nodeid = docs[j]._id.substring(2);
                     if (LCs[nodeid] != null) {
-                        delete docs[j]._id;
-                        LCs[nodeid] = docs[j];
+                        LCs[nodeid] = docs[j].time;
                     }
                 }
 
-                console.log(LCs);
+                try { ws.send(JSON.stringify({ action: 'lastseen', lastconnects: LCs })); } catch (ex) { }
             });
         });
     }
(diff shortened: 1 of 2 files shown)


nzalev:
This feature is requested in
- #2846
- #734
- #1693
- #1137
- #2329


---
Some additional technical notes: there may be two issues I have to address

1. If a user has a node in the database, but the node does not have an associated `lastconnect` doc, this will result in displaying `invalid date`. I'm not sure if this is possible though. In all my datasets, there are an equal number of LCs to nodes.
2. I have not had this happen in testing, but in a very high latency situation, it is possible that the `lastseen` message is received by the client before the `nodes` message. Because the client-side 'lastseen' handler will only set `node.lastconnect` if the node is not null, it is possible for the last seen handler to fail. In this case, the last seen time will simply not be displayed.

A very simple solution to 2 is to add a `setInterval()` in meshuser which will update the last seen dataset every few minutes, the very same way that 'serverstats' works. The last seen time will then be displayed on the next view update.

Ylianst:
Oh wow! Thank you. I think you will have the record for the one improvement that will close the most requests at once! I have some write paper to write today. After that, I will get this checked, approved and published ASAP. I think you will make a lot of people very happy.

Issue #2956, opened 2021-07-29
Minor Display Bug


PetieM:
In list view, the colspan set on header rows is not taking into account the newly added "Last Seen" column (which was a great addition). It's not affecting usability but it does look a bit odd when the header stops before the end of the row:

![image](https://user-images.githubusercontent.com/16453878/127425145-b8b41c27-66ee-45f5-81ee-b0d99dd5f3e6.png)

nzalev:
That's a bit strange. I remember dealing with this when I first prototyped it a while back, but I'm not getting the same results:
![Screenshot from 2021-07-29 11-11-06](https://user-images.githubusercontent.com/63608819/127517711-c3e3bad2-984e-490c-9e2d-30f66c44cfd1.png)

I tested with Firefox and Chromium. What browser are you using?

PetieM:
I'm seeing this behavior across the board but it's not actually exclusive to the new column and more seems to be a function of how many columns are selected. I have all but the Router links showing and am getting that header issue. If I disable Last Seen and enable Router Links, I see the same one column gap. If I enable both, that gap grows to two columns.

nzalev:
Thanks. I see the issue. It's column span is hard coded to 5, so it will only ever cover 5 columns. I will submit a patch.

PetieM:
Just FYI, setting the colspan to a greater number of columns than shown will technically work but breaks in table-layout: fixed display modes so it's best avoided. It would likely be better to grab the number of displayed columns on change and dynamically update the colspan value at that time (assuming that's not already your plan).

nzalev:
Yes, that's already the plan. The selected columns are stored in an array, `deviceViewSettings.devsCols`, so it is just a matter of taking length and adding 1 :)

Have to do it for each sort type though - since the header is generated separately for different view types.

Ylianst:
Oh dear. I switched `colspan` to 99 and it seemed to work before I came here and read your comments. @nzalev let me know if you are working on it, I will not touch it.

nzalev:
I'm about to push the fix

nzalev:
@Ylianst please see PR. Other device headers (power, group-tag, etc.) also need to be updated.

Ylianst:
Nicely done! The solution is very clean and elegant.

PetieM:
[...]

PR #2959, opened 2021-07-29
Dynamically size device headers colspan

nzalev:
This is an updated fix for #2956 

The hard-coded colspan affects more than just the the default view. Other sort views (power, tag, group-tag,..) also have statically set colspans.

This sets the colspan to be exactly the correct size, and works properly when changing the number of columns or changing the sort.


Ylianst:
I like it a lot, this is very elegant.

Issue #2970, opened 2021-07-30
Translator tool link is dead


OvrAp3x:
The link for the translator tool found here is dead,
https://github.com/Ylianst/MeshCentral/tree/master/translate

http://info.meshcentral.com/downloads/MeshCentral2/ResourceTranslator.zip

nzalev:
The link is not dead, but it is HTTP. The server will serve a 301 redirect to https.

This is likely being blocked for being mixed content, ie, trying to get HTTP from an HTTPS connection.

If you open the link in an incognito window, or you copy+paste it into a new window, it should download.

Ylianst:
@nzalev Thank you so much! Your fix was accepted (of course!).

@OvrAp3x Please check and close this issue if it's fixed - Thanks.

OvrAp3x:
Thank you🙂

PR #2971, opened 2021-07-30
Update translate link to https

nzalev:
The translate link will fail to download on some browsers due to Mixed Content (HTTP request from HTTPS connection)

#2970


Ylianst:
Oh wow! That is a great catch!

Issue #2972, opened 2021-07-30
Add "Last seen" data in export option


pggsadmin:
@Ylianst
Can you include device last seen details in export menu ?

Ylianst:
Good idea!

Ylianst:
Ok, working on this one now.

Ylianst:
Added last connection time and last connection address to device export. Will be in MeshCentral v0.8.92.

pggsadmin:
Thanks @Ylianst

PR #2973, opened 2021-07-30
Added Last Seen sort option

nzalev:
I think this is a fairly niche feature, but useful for users who have a very large number of deployed agents.

It is useful to find agents which have been disconnected for a long time: they may have been decomissioned and need to be removed, the agent may not be connecting and needs to be reinstalled, etc.

Up until the Last Seen column, it has been too tedious to go through disconnected agents and find how long they have been disconnected. However, when there are a very large number of agents, it can still be difficult to go through the list even with the last seen column.

Being able to sort by Last Seen solves this problem.

---

If the last seen column is not selected, sorting by last seen will automatically select it:
Screenshot from 2021-07-30 13-49-35
Screenshot from 2021-07-30 13-53-24
Screenshot from 2021-07-30 13-54-53

If the lastconnect data has not been fetched yet, it will be fetched and updated automatically.

### DIFF
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -318,6 +318,7 @@
                                     <option>Device</option>
                                     <option>Tags</option>
                                     <option>Group-Tags</option>
+                                    <option>Last Seen</option>
                                 </select>
                                 &nbsp;
                             </div>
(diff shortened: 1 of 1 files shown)


pggsadmin:
@nzalev is it possible to include the same in exporr options since i have 10k+ agents

nzalev:
> @nzalev is it possible to include the same in exporr options since i have 10k+ agents

I can look into #2972

Ylianst:
There is a problem since I made it so the agent last connected data is only pulled from the server when needed. In this view, the "lastconnects" data was not pulled yet, so it's sorting on an empty field. Also, if the last connect data is empty or identical, maybe sort by device name as a secondary sort?

image

Solution would be to pull the lastconnects data when this sort option is selected. I will work on that.

nzalev:
Ah, I see what you're saying. The last connect data is only pulled while in the list view.

I have tested it out and adding
```javascript
if (requestedLastConnects == false) { requestedLastConnects = true; meshserver.send({ action: 'lastconnects' }); }
```
updates that view in the same way as the list view.

PR #2978, opened 2021-07-31
German translation update

veitw:
Fixed translation errors and added missing translations.

### DIFF
--- a/translate/translate.json
+++ b/translate/translate.json
@@ -1082,7 +1082,7 @@
     },
     {
       "cs": ", {0} sledování",
-      "de": ", {0} zuschauen",
+      "de": ", {0} sehen zu",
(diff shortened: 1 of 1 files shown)


Ylianst:
Thanks you so much! This will be in MeshCentral v0.8.93.

veitw:
Always glad to support your amazing work! :)

PS: For the first time I translated strings that seem to belong to MeshCentral Assistant, but I was unable to verify success. Am I right to assume these changes will not be visible until MeshCentral Assistant is rebuilt?

PPS: In Assistant of all things the desktop access consent question string does not seem to be translatable as it does not show up in translate.json, though this seems to be the most important one of all to be legally safe when offering this to people. Maybe some single quote/double quote confusion?

PR #2986, opened 2021-08-04
Handle all configurable paths as potentially absolute

lheckemann:
Essentially the same as #2928, but extended to all places I was able to find where `path.join` was used indiscriminately and the right-hand side of the join was a variable; in addition, I added a function to `common.js` to factor the common logic out into one place, and replaced some handwritten absolute-path checks as well.

:warning: This is running with no issues on my meshcentral deployment, but I have not tested all the functionality I touched, so a careful review would be appreciated.


Ylianst:
Thanks. It looks like you removed some of the error handling and handling of cases where the paths are not set in the config.json. I will need to take a look at this closely and made some changes.

Ylianst:
Ok, I will accept and make changes.

Ylianst:
Actually... after second review, it looks very good! I like it.

PR #3044, opened 2021-08-20
Mobile terminal improvements

### LINKED ISSUE #2478: Feature Request: Terminal Tab in Mobile View Mode
Hello

It would be nice to have the Terminal Tab also in Mobile View. 

MeshCentral ist absolutely amazing!!!!!

### LINKED ISSUE #2814: Terminal on mobile version hangs meshagent
Whenever text is deleted on the mobile version of meshcentral, Linux agents hang completely.
I've already described this here https://github.com/Ylianst/MeshCentral/issues/2204#issuecomment-824010947 in April, but the issue was recently closed for whatever reason.
The same issue is also described [here](https://github.com/Ylianst/MeshCentral/issues/2478#issuecomment-817563865) and [here](https://github.com/Ylianst/MeshCentral/issues/2478#issuecomment-828204880).
This unfortunately makes the mobile terminal rather unusable :/
I've also noticed these hangs sometimes happen on the desktop version of MeshCentral too after entering some characters, although I can't actually reproduce the issue reliably. This makes me suspect that there's an underlying issue.
I've already done some investigative work on the agent by piping the ws output into the agents stdout, and it appears that backspace from the mobile site acts differently than backspace from the desktop site, but I don't know the exact specifics since the "data" event does not seem to act the same way as it does on nodejs.

TotallyNotElite:
Did some work to make it stop crashing agents and made it actually usable to some degree. Doesn't work *that* well in Firefox on mobile but it works fine in chrome (the Firefox issues were already there before).

Let me know what you think

Fixes #2814
Closes #2478

### DIFF
--- a/views/default-mobile.handlebars
+++ b/views/default-mobile.handlebars
@@ -1,4 +1,4 @@
-﻿<!DOCTYPE html>
+<!DOCTYPE html>
 <html lang="en" dir="ltr" xmlns="http://www.w3.org/1999/xhtml">
 <head>
     <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


TotallyNotElite:
The actual crash fix is found here:
https://github.com/Ylianst/MeshCentral/pull/3044/files#diff-944f10e74e9d7a95eaf58836982d1eee14d7737b1e9eb71ca44f142f8d5f7368R2882

Agent just freezes if it gets an empty message. Probably something that should be fixed on the agent itself but I'll let you deal with that.

Ylianst:
Oh nice! I see what you did there. I had never seen the terminal crash on mobile and did not do the same "screen fitting" on mobile as the desktop site. Totally approving this. Will be in v0.9.8.

Issue #3231, opened 2021-10-30
Files > Drive Connect > Find non-functional Windows 10 21H1


SomeGuru:
Hi Meshcentral team,

I attempted to us the Find button in Files tab and it doesn't fetch any files even the demo *.txt doesn't work. Is this expected behavior?

Thanks,
SomeGuru

nzalev:
This is a bug. I have submitted a fix for it.

Ylianst:
Outstanding fix by @nzalev. Pull request was accepted and fix will be in MeshCentral v0.9.40.

SomeGuru:
Completed! Closing issue...

-SomeGuru

PR #3232, opened 2021-10-30
Fixed reference error in file search (fix for #3231)

nzalev:
This is a fix for #3231

When attempting to search a node's files, the variable `winAgent` was written as `winagent`.


Ylianst:
Oh wow! Yes, indeed... excellent fix!

PR #3273, opened 2021-11-11
Fix bug in addusertousergroup

HuFlungDu:
Fix obscure bug when the last user added in list of addusertousergroup was an unknown user causing the server to crash.

Also updated the group update message of same to display all users added, rather than only the last one in the list.

### DIFF
--- a/meshuser.js
+++ b/meshuser.js
@@ -2006,7 +2006,7 @@ module.exports.CreateMeshUser = function (parent, db, ws, req, args, domain, use
                     if (group != null) {
                         if (group.links == null) { group.links = {}; }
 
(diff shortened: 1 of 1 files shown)


Ylianst:
Thanks. Should `knownUsers` not be initialized to an empty array? like `var knownUsers = [];`?

Otherwise, this is an excellent fix. I just want to make sure you tested it and it worked.

HuFlungDu:
Oh yeah, looks like you're right. I tested it, but when I copied over the
code from the rest environment, looks like I forgot to copy over the
initialization correctly. I'll update the pr with that fix.

Ylianst:
Thanks. This will be in MeshCentral v0.9.51.

PR #3310, opened 2021-11-29
Fix MariaDB/MySQL power events

nzalev:
For power events, when the dates/times are pulled from MySQL/MariaDB, they get pulled as strings instead of Date objects.

This causes the power events csv generation to fail, as a string has no function `getTime()`. 

I have added a check to see if time pulled from the database is a string, and if so, to convert it to a Date object.


Ylianst:
Nice! Thank you!

Ylianst:
MeshCentral v0.9.55 is out and includes this fix!

PR #3411, opened 2022-01-06
Update agent ip on device reconnect - fix #3354

### LINKED ISSUE #3354: IP Address column shows last IP instead of current.
Hi. Just recently realized that IP Address column in list view does not show correct information. One of the employees got his reinstalled in the office laptop and took it to the other one, as soon as connected to internet - it showed the local IP which was in the main office. However, not sure if that make sense here are steps to recreate:
1. Install and run agent on some laptop on some network (let say network "one" with range 192.168.5.X).
2. Address column on MC server will show let say 192.168.5.81 (MC Server is in hybrid mode)
3. power off the laptop and take it to other place, or mobile hotspot
4. As soon as it come online - Address field in list view on MC still show 192.168.5.81
5. When click on device itself and go to Details tab, IP is different :)
6. Disconnect laptop from hotspot and connect to 3rd different network
7. Address on MC server in list view will be the address from the 2nd connected network.
8. And so on :)

Outcome - looks like Address field in list view on MC shows the last connected address instead of current IP address.
In my case for WAN agents - current IP: 86.X.X.X, Address on MC is: 33.X.X.X, disconnected from 86.X.X.X and connected to 74.x.x.x on MC shows 86.x.x.x

MC server: 0.9.55 (LAN+WAN)
OS: Ubuntu Jessie
Agent OS: Windows 10 x64

nzalev:
This updates the agent IP on reconnect. As it currently stands, when an agent connects, the IP address is only updated if the agent is initally starting and sends its core.

If an agent reconnects without restarting (suppose it moves from a mobile hotspit to a WiFi connection), nothing will have changed in the core, so it will not resend it. However, the device IP is only set from `ChangeAgentCoreInfo()`. As a result, `node.ip` ends up having an incorrect result in the database when a device changes networks.

#3354


Ylianst:
Nice! Thank you!

PR #3412, opened 2022-01-06
Add KEEPALIVE_OPTIONS_REQUEST message support to mpserver

Ganeshr93:
I have seen an interesting behavior on AMT 15 Dell platform, where AMT sends a keepalive to MPS (CIRA) every 90secs. Default timeout on MPS is 90 secs, and as a result socket timeout is triggered and MPS kills (ends) CIRA socket connection. Within few seconds, AMT device connects back to MPS but again timeout is triggered (and this happens continuously forever).

All other AMT devices by default send keepalive message every 30secs (tested this behavior on AMT 11, 12, 14 and AMT 15 versions).

I've added KEEPALIVE_OPTIONS_REQUEST logic to MPS so we can synchronize the keepalive time (i.e., to 30secs) across all devices.

Specification link: https://software.intel.com/sites/manageability/AMT_Implementation_and_Reference_Guide/HTMLDocuments/MPSDocuments/Intel%20AMT%20Port%20Forwarding%20Protocol%20Reference%20Manual.pdf

### DIFF
--- a/mpsserver.js
+++ b/mpsserver.js
@@ -30,6 +30,7 @@ module.exports.CreateMpsServer = function (parent, db, args, certificates) {
     const net = require('net');
     const tls = require('tls');
     const MAX_IDLE = 90000;         // 90 seconds max idle time, higher than the typical KEEP-ALIVE periode of 60 seconds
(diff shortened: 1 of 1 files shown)


Ylianst:
Wow, interesting! Good pull request. This will be in the next version.

PR #3427, opened 2022-01-08
Updated Polish translation.

nicrame:


Ylianst:
Thanks. I took your file and only merged the "pl" language strings. It will be in MeshCentral v0.9.67 when it's next released.

PR #3429, opened 2022-01-08
Factorize meshuser

nzalev:
Added function handlers for the following:

- addmeshuser
- adddeviceuser
- agentdisconnect
- close
- confirmPhone
- removePhone
- verifyPhone

Bug fix:
In the `close` server command, `obj.req` can be undefined, so `if (obj.req.session ...` can cause
```
TypeError: Cannot read property 'session' of undefined
```

Since we are only deleting it, we do not need to meticulously test for its existence - just stick it in a try block - if it throws a type error, it doesn't exist anyway.

Other improvements:
I have removed some hidden unicode characters from a comment, which generate warnings when viewing the file on github.

### DIFF
--- a/meshuser.js
+++ b/meshuser.js
@@ -3188,13 +3188,6 @@ module.exports.CreateMeshUser = function (parent, db, ws, req, args, domain, use
                     }
                     break;
                 }
-            case 'close':
-                {
-                    // Close the web socket session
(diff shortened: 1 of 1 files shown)


Ylianst:
Nice! Thanks!

PR #3470, opened 2022-01-16
Add agentkey to meshagent.msh generation

nzalev:
This adds agentKeys to meshagent.msh generation, to avoid manually editing the .msh file for future agents.


Ylianst:
Arg! I don't think we want to do that. It would defeat exactly the purpose of this agentkey.

If we embed the key in the agent download (agent executable or agent install script), this will cause the exact problem people using this system want to avoid. They want to make it so you can't download the agent and use it to connect to the server. This system is (on purpose) an extra road block to having agents connect.

I am ok with doing this if an extra option is set in the config.json to embed the key in the .msh, but at that point, not sure what use this system would have?

nzalev:
I generally agree.


The main thing that I notice are antivirus vendors connecting to user servers, but they have access to the .msh file for the agent to connect already - I don't think they download it through the agent link.

The next that I generally see are scanners / probes connecting on the agent port, which probably don't have a download link either (and likely won't know how to find one).

The main use case I see is when MSPs publish the download link on their website for their clients. But outside of that, I think including it in the .msh file is a good compromise between increased security and usability.

If that sounds reasonable, I can add an additional domain option to enable adding to the .msh file and re-open the pull request.

PR #3478, opened 2022-01-18
Include agentKey check at agent only port

nzalev:
The agent key was not checked if using `agentPort`.

This includes the check for agent keys if using an agent port.


Ylianst:
Oh wow!! Nice catch!!!

Issue #3464, opened 2022-01-15
Switched to mySQL - Server Crash with Error


JSuenram:
Just imported Data with Meshcentral from neDB to mySQL, server is running but crashed after a few minutes...
Service restarted but unsure what has happened here?

```
-------- 15.1.2022, 21:05:58 ---- 0.9.68 --------

events.js:291
      throw er; // Unhandled 'error' event
      ^

Error: Packets out of order. Got: 0 Expected: 6
    at Parser._tryReadPacketHeader (/etc/meshcentral/node_modules/mysql/lib/protocol/Parser.js:470:15)
    at Parser.write (/etc/meshcentral/node_modules/mysql/lib/protocol/Parser.js:33:29)
    at Protocol.write (/etc/meshcentral/node_modules/mysql/lib/protocol/Protocol.js:38:16)
    at Socket.<anonymous> (/etc/meshcentral/node_modules/mysql/lib/Connection.js:88:28)
    at Socket.<anonymous> (/etc/meshcentral/node_modules/mysql/lib/Connection.js:526:10)
    at Socket.emit (events.js:314:20)
    at addChunk (_stream_readable.js:297:12)
    at readableAddChunk (_stream_readable.js:272:9)
    at Socket.Readable.push (_stream_readable.js:213:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:188:23)
Emitted 'error' event on Connection instance at:
    at Connection._handleProtocolError (/etc/meshcentral/node_modules/mysql/lib/Connection.js:423:8)
    at Protocol.emit (events.js:314:20)
    at Protocol._delegateError (/etc/meshcentral/node_modules/mysql/lib/protocol/Protocol.js:398:10)
    at Protocol.handleParserError (/etc/meshcentral/node_modules/mysql/lib/protocol/Protocol.js:380:10)
    at Parser._tryReadPacketHeader (/etc/meshcentral/node_modules/mysql/lib/protocol/Parser.js:478:10)
    at Parser.write (/etc/meshcentral/node_modules/mysql/lib/protocol/Parser.js:33:29)
    [... lines matching original stack trace ...]
    at Socket.emit (events.js:314:20) {
  code: 'PROTOCOL_PACKETS_OUT_OF_ORDER',
  fatal: true
}
```

Ylianst:
[...]

PR #3501, opened 2022-01-21
Translation updates.

JSuenram:
Nothing important. Just added 2 missing German translations and changed one.

### DIFF
--- a/translate/translate.json
+++ b/translate/translate.json
@@ -748,7 +748,7 @@
     },
     {
       "cs": "(Žádný)",
-      "de": "(Keiner)",
+      "de": "(keiner)",
       "en": "(None)",
       "es": "(Ninguno)",
       "fi": "(Ei mitään)",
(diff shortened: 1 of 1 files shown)


Ylianst:
Can you explain why you changed 'createPool' to 'createConnection'? Looks like a pool could handle many queries at once and so, would perform better. Also, did you test this?

JSuenram:
> Can you explain why you changed 'createPool' to 'createConnection'? Looks like a pool could handle many queries at once and so, would perform better. Also, did you test this?

This is the fix for https://github.com/Ylianst/MeshCentral/issues/3464 and also gives some performance boost.
At least it makes the connection much more stable between Node and mySQL even without setting max_allowed_packet=500M. So in my opinion this is the better way to connect to mySQL.

Ylianst:
Nice! Thanks for pointing out #3464. This is excellent!!

Issue #3510, opened 2022-01-22
Assitent Translate User conssent Message ?


CSW2000:
Hello
1. How can I translate user consent from wizards? ( Assistant )
        In the agent this works with "consentMessages".. Desktop … Terminal .... in config.json.
        But how do I do this with the assistant?
2. And how does it work with "Auto accept all requests for next 5 minutes" line
3. And where can I set the time 5 minutes is usually very short to really help.
Greeting from Bavaria / Germany
Roland W. 
 

![userconsens PNG](https://user-images.githubusercontent.com/56474379/150631963-b4713f5f-03d2-4d73-ba1c-1119d26ef89f.jpg)

JSuenram:
Translation is here I guess.
https://github.com/Ylianst/MeshCentral/blob/master/agents/modules_meshcore/coretranslations.json
I can add needed German translations and commit.

CSW2000:
Thanks it was a fast service now
But what happens after update of Meshcentral2, is my translation gone? – I hope not 
Since the change is made outside the config.json.

Roland W.

JSuenram:
> Thanks it was a fast service now But what happens after update of Meshcentral2, is my translation gone? – I hope not Since the change is made outside the config.json.
> 
> Roland W.
> 
> Hier ist die Deutsche Übersetzung für den DE Teil coretranslations.json [coretranslations.json.txt](https://github.com/Ylianst/MeshCentral/files/7918939/coretranslations.json.txt)

Well I already made a translation and "committed" it. If it will accepted by @Ylianst it will be in the Mainbranch and in all future Updates.

German: Ich habe die Änderung bereits eingereicht damit sie auch bei Updates nicht mehr überschrieben wird. Allerdings habe ich etwas anders übersetzt....

CSW2000:
Have the German translation as above in the text in the "node_modules/meshcentral/agents/modules_meshcore/coretranslations.json"
Built-in – Saved and Restarted Meshcentral2, but the User Consent wizard window is in English.

JSuenram:
> Have the German translation as above in the text in the "node_modules/meshcentral/agents/modules_meshcore/coretranslations.json"
> Built-in – Saved and Restarted Meshcentral2, but the User Consent wizard window is in English.

Sure, it will change when the Assistant is recompiled with an Update. This might be 0.9.73 or a later Version.
You have to wait until there is a new release with the changes committed or compile the whole thing on your own.

CSW2000:
[...]

PR #3514, opened 2022-01-22
Fix for #3510

JSuenram:
Just some more translations.


Ylianst:
Thanks! I accepted the commit, however, the file "coretranslations.json" is actually auto-generated from "translations.json". So, the only file that needs translation for all tools, agents and web pages is "translations.json".

I will put the translations back into that file. Apologies, I know this is not obvious.

Ylianst:
Ok, fixed it in [this commit](https://github.com/Ylianst/MeshCentral/commit/7fe87663143f5995214d7bb15769f19f06813fda).

PR #3545, opened 2022-01-26
tranlaste.json tiny typo; first, very carefull small change in style.…

JSuenram:
…css and default.handlebars to get an idea about complete UI-Control frontend.

Hi,
please if this does not make any sense just "deny". I am just trying to get the bigger view about creating a full UI-customization Frontend.

- First step get missing styles in css and remove hard coded styles. If that works as I believe, correct me or just deny if not, I can start creating a new webUI customization-page.
- 
image

### DIFF
--- a/public/styles/style.css
+++ b/public/styles/style.css
@@ -257,6 +257,10 @@ body {
     right: 3px;
 }
 
+.LogoffLinkColor {
+    color:white;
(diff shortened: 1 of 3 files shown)


Ylianst:
Nice! Looks good!

Issue #3532, opened 2022-01-25
Request to change one translation entry into two


nicrame:
There is translation entry "Type" that is used in two places:
Here My devices \ Device \ Desktop \ :
![Zrzut ekranu (521)](https://user-images.githubusercontent.com/5872054/150962159-6c94fafe-0242-4183-8e67-fbf3b7421ad3.png)
  
And here at My devices \ Device \ Share \ :
![Zrzut ekranu (522)](https://user-images.githubusercontent.com/5872054/150962437-2bfdbe79-26c3-40eb-9434-90e07c75d326.png)

Locations:
default-mobile.handlebars->11->122
default-mobile.handlebars->11->623
default.handlebars->39->1026
default.handlebars->39->1241
default.handlebars->39->1747
default.handlebars->39->1807
default.handlebars->39->1897
default.handlebars->39->2695
default.handlebars->39->431
default.handlebars->container->column_l->p11->deskarea0->deskarea4->3
sharing.handlebars->p11->deskarea0->deskarea4->3

The problem is, while in English it is the same word for both typing data, and choosing type of something, in Polish for example, we use two different words (typ for type of something, and wprowadzanie for typing text into). I think it may be similar for other languages.
So i request for forking that one entry into two.

JSuenram:
Yes, also here in German this is not done with a single "word"....
+1

Ylianst:
Ok, working on this one now.

Ylianst:
Hopefully fixed. Look for this line in translate.json:

```
"en": "[KeyboardTyping]|Type",
```

You only need to translate the part after the |. Let me know if this works.

nicrame:
Should be fine now. Thank You!

PR #3549, opened 2022-01-26
Updated type in Polish translation

nicrame:
https://github.com/Ylianst/MeshCentral/issues/3532


Ylianst:
Fantastic! Thanks you!

PR #3556, opened 2022-01-27
docs - create workflow for auto-rebuilding and initial theme

silversword411:
You should be able to monitor the mkdocs building under actions. 
Any time something is changed in the `/docs` folder it should auto-rebuild.

Then you should be able to see the new docs at https://git.meshcentral.com/ after about a minute

At some point might be nice to change that to https://docs.meshcentral.com/


Ylianst:
Interesting! This is an area I have no competence in, so this is double wonderful. I can't wait to see it in action.

PR #3567, opened 2022-01-30
Russian language update and fixes

tylerway:
- all missing Russian lines added 
- many out of context, word order and declension errors fixed.


Ylianst:
Thank you!!! This will be in MeshCentral v0.9.76.

PR #3622, opened 2022-02-09
Tranlsation Fix Clipboard-Dialog....

JSuenram:
Buttons to small.... changed German translation to better fit and changed Button-Size overall...
image


Ylianst:
Nice fix! Thank you!

Issue #3362, opened 2021-12-15
CTRL+W warning


jdvmanen:
Hi,

As a keybord shortcut lover, I use CTRL+W to close a browser tab.

However, if I'm remotely working on a PC, and want to close a tab on that remote PC using CTRL+W, then it closes my meshcentral-remote-desktop-tab!

Can it be made that closing a meshcentral tab issues a warning "are you sure to close this tab", so I can cancel my stupid CTRL+W action?

drnikcuk:
+1 ! i have the same issue daily.

AntonAndreevichMoroz:
Yes, I also happen to close a tab by mistake when I try to search for text in nano Linux. But it seems to me that this key combination is intercepted by the system before the MeshCentral code can see it. Therefore, confirmation must be configured at the browser or system level. I don't think this is possible at the code level of the Mesh application.

JSuenram:
You can add this 4 code-lines in default.handlebars until my Pull-Request is accepted or at least "optimized".

`        window.addEventListener('beforeunload', function (e) {
            e.preventDefault();
            e.returnValue = '';
        });`

Just put this at the very end of the file:

`
        // Quick fix for #3362 - Request Confirmation if MeshCentral-Browser-Tab is closed. - Maybe needs an option per User or Config.JSON-Setting
        window.addEventListener('beforeunload', function (e) {
            e.preventDefault();
            e.returnValue = '';
        });
    
    </script>
</body>
</html>`

jdvmanen:
:-) MeshCentral doesn't boot after this edit in config.json...

JSuenram:
Not in config.json. In default.handlebars you need to add this....

JSuenram:
Has been implemented in latest 0.9.83 please confirm and close.
Thank you!

si458:
Fixed

PR #3641, opened 2022-02-14
Quick fix for #3362 - Request Confirmation if MeshCentral-Browser-Tab…

JSuenram:
… is closed. - Maybe needs an option per User or Config.JSON-Setting
QuickFix #3362

### DIFF
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -17216,6 +17216,12 @@
             }
         }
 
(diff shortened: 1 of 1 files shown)


JSuenram:
image

Ylianst:
Thanks. That is a good improvement. I did a change to this to only show the prompt when there is a active desktop or terminal session and you are looking at the desktop or terminal. This way, it's not going to bother people as much.

PR #3677, opened 2022-02-18
Update backup data types in config schema

masonh:
Swapped the maxFiles and folderName data types for webDAV and Google Drive backup uploads


Ylianst:
Nice! Thanks!

PR #3764, opened 2022-03-15
Fix Resend verification email feature

tunght:
For unverified email user, there is no way to resend the verification email.
On the UI the button to resend verification email but the button check if the email is already existed and show error. 
This checks if the existed email belong to the current user, then just send the verification email to the current user.


Ylianst:
Nice! Thanks!

PR #3842, opened 2022-04-04
Improve IDE-R Support of meshcmd on linux

cwgthornton:
Each commit message is relatively self explanatory for what it does, but with these patches I was able to reboot/image a device over the command line using meshcmd


Ylianst:
Good changes! But `new Buffer()` is deprecated and documentation indicates that `Buffer.from()` need to be used instead. I am curious where you are running this that `Buffer.from()` is needed? - Thanks.

cwgthornton:
I'm no node developer, I was just trying to figure out why the ide-r implementation works when run from the browser, but was failing when run on the command line. AFAICT the way to run things was using the `meshagent_x86-64` (or equivalent) binary that you included in the agents folder. After trying to figure out why it was stalling/not working and dropping a million log statements in there, it turned out that Buffer.from([integer, integer....]) was always returning null (or undefined, can't remember which). It works correctly for the other uses where it was converting a string, but for some reason was barfing when it had the list of integers... if there's a better way that works, happy to do that, but it did not seem to be working as is.

Issue #3923, opened 2022-04-27
Viewing / removing / managing / auditing login tokens


silversword411:
I haven't been able to find any notes one way or another relating to tokens when enabled (from manual 14.1 Login Token section).

What if a token is compromised?
How many tokens exist?
When was a token last used?
How do you delete a token?

Is any of this exposed in the web gui (I would have thought they'd show in "previous Logins" under user but they don't seem to)?

Is the only way to create a token from command line with: `node meshcentral --loginTokenKey` ?

Ylianst:
Good questions. The login tokens in the user manual section 14.1 can't be tracked, deleted or revoked. The are 1 hour tokens. The only thing I could do is get you an easy way to reset to login token encryption key, but that would just invalidate all outstanding tokens.

Since then, I added a better system, images below. These tokens can be created, revoked and tracked.

![image](https://user-images.githubusercontent.com/1319013/165801181-a212722b-4f04-4451-ad75-69d5b4faaa58.png)

![image](https://user-images.githubusercontent.com/1319013/165801215-4873126f-f725-4ada-9c73-05e545797274.png)

Let me know if this helps.

silversword411:
This is great info, and I'll add it to the docs.

I am using this as part of Tactical RMM, so this is actually using the other software integration process :)

Having a command for global "reset to login token encryption key" would definitely be good to have for failsafe reset process.

Yes, it does look like TRMM is using that command to get a token key, and being able to reset/recreate it would be good:
https://github.com/amidaware/tacticalrmm/blob/3a1ea42333533dab1de96adf611a5a24bacc2262/install.sh#L807

Thanks!

PR #3965, opened 2022-05-12
Docs - fixing debugging, adding token notes

silversword411:
I'll add the software reset token as mentioned here: https://github.com/Ylianst/MeshCentral/issues/3923
when you get to adding it thx!


Ylianst:
This looks GREAT! Thank you so much.

PR #4063, opened 2022-06-01
add more detail on no language set for translation

reddn:
also made default langauge 'en_us' instead of 'en'


reddn:
ill open a new one after the fix

PR #4076, opened 2022-06-03
added console comment to help with install

reddn:
require('util-langauge').current will give a 'c' as the language which is incorrect, unkonwn root cause.  As a workaround, the console will output a message for the user to specify the language.


reddn:
Added actual fix in next PR

Issue #4061, opened 2022-06-01
Ran prebuilt binary on ARM HF Raspberry Pi results in "Language: c is not translated." install failure


reddn:
Pre-built binary does not work.

Installing using the meshinstall.sh made the install work correctly

krayon007:
Does it work if you run the prebuilt binary but add the parameter? (I can't remember if it's a single dash or double dash)

`-lang=en_us`

reddn:
Yep, that was it... 

I think I found where if it was null, it defaulted to 'en' instead of 'en_us'.  

I started a pull request where i changed the default to en_us also added a bit more detail of how to fix the problem by giving the example like you listed.  

Here is the original https://github.com/Ylianst/MeshCentral/blob/f295d2f780776c0144fddfac6e92fdd6c636ae69/agents/meshinstall-linux.js#L48

Here is the pull request https://github.com/Ylianst/MeshCentral/pull/4063

Thank you for your response.  

Issue closed.

reddn:
So, I think I tracked it down.  the "require('util-language').current" must return something strange.  But I can not find what and where this mysterious package is.  Although there are a bunch of split and join calls, I'm guessing it is returning 'c' instead of defaulting to 'en'.  

I created a bandaid where if the length of the util-language.current was less than 2, to make the language 'en' (en-us is not needed).  I did not propose this as a PR, but I did PR a comment adding to use the '-lang=en' which will make the install work.

reddn:
new PR.. this is a permeant fix.  Basically the util-langauge.current returns the process.env.LANG variable, which defaults to "C" and instead of 'null' when not set.    The PR defaults the langauge to en when it see's "C". 

https://github.com/Ylianst/MeshCentral/pull/4079

reddn:
PR committed. 

Although I think any unrecognized language setting should default to 'en'.  This exact issue is closed.

PR #4079, opened 2022-06-03
 added fix to hande process.env.LANG = "C" 

reddn:
When the environment process does not have a language set, it defaults to "C". This PR changes the language to English and sends a console log about how to find and specify the language for translation.

Without this, the it will set the language to lowercase "C" which is not found in agent-translations.json, which throws an error on install "Language: c is not translated" and exits.

### DIFF
--- a/agents/meshinstall-linux.js
+++ b/agents/meshinstall-linux.js
@@ -45,7 +45,11 @@ var msh = {};
 var translation = JSON.parse(msh.translation);
 
 var lang = require('util-language').current;
-if (lang == null) { lang = 'en_us'; }
+if (lang == null) { lang = 'en'; }
+if (lang == "C"){
+    lang = 'en';
+    console.log("Langauge envronment variable was not set (process.env.LANG).  Defaulting to English ('en').\nSee the agent-translations.json file for a list of current languages that are implemented\nUsage: meshcentral -lang=en\n");    
+}
 if (process.argv.getParameter('lang', lang) == null)
 {
     console.log('\nCurrent Language: ' + lang + '\n');
@@ -60,7 +64,8 @@ else
         if (translation[lang.split('-')[0]] == null)
         {
             console.log('Language: ' + lang + ' is not translated.');
-            console.log("try: './meshcentral -lang=en_us'"
+            console.log("try: './meshcentral -lang=en' for English");
+            console.log("See the agent-translations.json file for a list of current languages that are implemented.")
             process.exit();
         }
         else
--- a/agents/meshinstall-linux.js
+++ b/agents/meshinstall-linux.js
@@ -45,7 +45,7 @@ var msh = {};
 var translation = JSON.parse(msh.translation);
 
 var lang = require('util-language').current;
-if (lang == null) { lang = 'en'; }
+if (lang == null) { lang = 'en_us'; }
 if (process.argv.getParameter('lang', lang) == null)
 {
     console.log('\nCurrent Language: ' + lang + '\n');
@@ -60,6 +60,7 @@ else
         if (translation[lang.split('-')[0]] == null)
         {
             console.log('Language: ' + lang + ' is not translated.');
+            console.log("try: './meshcentral -lang=en_us'"
             process.exit();
         }
         else


reddn:
I guess the mysterious util-langauge package just returns process.env.LANG (at least on node 18+).  This defaulted to "C" on my raspberry pi, which obviously will cause an issue as the code current expects util-language.current to return 'null' if there is no locale setting.  

The current code looked for a 'c' in the agents-translation.json file, which there is none, which causes the error in issue https://github.com/Ylianst/MeshCentral/issues/4061 4061.

reddn:
Here is what the raspberry pi showed under 'process.env'

`> process.env
{
  HISTCONTROL: 'ignoreboth',
  EDITOR: 'vim',
  PWD: '/root',
  QT_QPA_PLATFORMTHEME: 'qt5ct',
  LINES: '42',
  LANG: 'C',

  LS_COLORS: **removed**
  COLUMNS: '91',
  PROMPT_COMMAND: 'PS1="[$(prompt_user)@$(prompt_host)$(prompt_git)]$(prompt) "',
  TERM: 'xterm-256color',
  SHLVL: '1',
  PAGER: 'less',
  PATH: '/usr/local/sbin:/usr/local/bin:/usr/bin:/sbin',
  OLDPWD: '/usr/local/mesh_services/meshagent',
  _: '/usr/bin/node'
}
`

My other computer where the error did not occur showed this 
`> process.env.LANG
'en_US.UTF-8'`

reddn:
Would it be better to make it where if process.env.LANG does not match a language in the agent-translator.json (file name might not be correct) that it should give a error in console and default to 'en'?

This would make it where no matter what crap you send to it as a set language, it will still run and not fail.

krayon007:
For giggles, can you run the following three commands on your system that is showing "C"?
`./meshagent -exec "console.log(require('user-sessions').gdmUid);process.exit();"`

That will show the UID of the GDM on the system.... Then using that, run the following command, substituting 114 with the UID from above:
`ps -e -o pid -o uid | grep 114`

Then using the returned PID, run the following command, substituting 1562 with the PID from above
`./meshagent -exec "console.log(require('user-sessions').getEnvFromPid(1562));process.exit();"`

I'm curious if the above command has the correct language setting. It is basically fetching it from the environment of the GDM. This method is what I normally try to do if the language setting in the agent's process isn't set. In your case, the agent's environment's lang is set to "C".

reddn:
```
[root@pikvm root]# ./meshagent -exec "console.log(require('user-sessions').getEnvFromPid(16969));process.exit();"

{ LANG: C, PATH: /usr/local/sbin:/usr/local/bin:/usr/bin, INVOCATION_ID: 57e9c73fcb0345de846cc66ebb05d817, SYSTEMD_EXEC_PID: 16969 }
[root@pikvm root]# ps -aux | grep mesh
root     15688  0.0  0.0   2920  1188 pts/4    S+   15:45   0:00 grep --color=auto mesh
root     16969  0.0  0.6  23172 12984 ?        Ss   Jun01   0:30 /usr/local/mesh_services/meshagent/meshagent --installedByUser=0
[root@pikvm root]# 
```

reddn:
The first command didnt work on it 

```[root@pikvm root]# ./meshagent -exec "console.log(require('user-sessions').gdmUid);process.exit();"

0
[root@pikvm root]# ```

but I just did `ps -aux | grep mesh` to the the pid of the mesh agent

krayon007:
What display manager are you running? It's actually the pid of your display manager I was interested in. If you check your passwd file you should be able to find the uid of your display manager.

reddn:
No display manager.  Its a headless raspberry pi running PI KVM

reddn:
based on arch

reddn:
Did some digging.. my Ubuntu 18.04 LTS headless has it set correctly (under
/etc/default/locale), but on this raspberry pi, based on arch, has the file
/etc/locale.conf which has "LANG=C".

Instead of making "C" default to 'en', I think it would be better to make
it default to 'en' if the specified locale language does not meet the list
already in agent_translations.json.  Let meknow if you want me to do that
and I'll take out my 'workaround' (if LANG = C) as the workaround doesnt
catch any other dumb LANG configuration that does not match a langauge.

PR #4083, opened 2022-06-06
Updated docker environment

sschoen:


si458:
was the any reason this was merged?
also why was the dockerfile changed to include mongodb?
the size of the container has now increased!

sschoen:
only the mongodb-tools are included, these are needed for the autobackup-option...

sschoen:
good point, thought about that and made the installation of mongo-db tools optional through build parameter

https://github.com/Ylianst/MeshCentral/pull/4086

PR #4096, opened 2022-06-08
docs - adding new config from https://github.com/Ylianst/MeshCentral/…

silversword411:
…issues/4084


Ylianst:
Oh dear, I need to start updating that file too!

silversword411:
I'll try and keep an eye out too, just an easy copy/paste ;)

Why it put it in the way I did.

PR #4103, opened 2022-06-09
[docker] bugfix: run 'node translate.js extractall' only when DISABLE_TRANSLATE is not set

sschoen:


si458:
is it also possible to change the `docker.build.sh` to allow us to specify a tag?
currently its building to `meshcentral:latest` but i would like to build when a new version is released with a new tag `meshcentral:1.0.36`

PR #4151, opened 2022-06-21
docs - adding meshcommander commandline options

silversword411:


silversword411:
Searching for "MeshCommander" in the docs, pulled results. Figured they were the same thing...but apparently not? Why is there meshcommander in MeshCmd?

Do we need one of those things to be renamed from MeshCommander to something different?

New section of the docs?

Ylianst:
I just updated the [switches documented in MeshCommander](https://github.com/Ylianst/MeshCommander/commit/d1800bfdd853f766f184db7092bd717833d6eef9). MeshCommander is an odd beast. It needs to be "compiled" using the WebSite Compiler, which is a C# tool that will generate one of many MeshCommander "editions". The editions are:

- NW.js that is found in the .MSI windows installer.
- NPM version, you get is using `npm install meshcommander`
- There are two within MeshCMD, a web one and the LMS edition.
- There is one in MeshCentral.
- You could add the IIS version, but I don't update that one anymore.
- Most importantly, there use to be the firmware edition that you could load directly into Intel AMT firmware.

In any case, the switches are only for the .MSI version. MeshCommander is a weird Intel AMT tool with multiple ways to embed it into different situations.

silversword411:
...and in the "why are you so confused about meshcommander" dept...I feel much better in seeing, but not necessarily understanding all that 🤣

Thanks for the fixups!

Issue #1662, opened 2020-07-28
Question:  What is the effect of setting ignoreagenthashcheck to true?


darryl-h:
# Use Case:
In our case, some of our customers in very high security networks are replacing our cert with theirs, presumably to be able to see inside the TLS traffic to verify/process it's payload with their chosen security software. In our case, we have to modify the `meshinstall.sh` file to add `--no-check-certificate` to allow the download to complete for the binary and configuration .msh files, however, upon doing this, the agent is still not able to connect.

Anecdotally I have worked at very large vendors who replace the cert (A kind of ethical MITM attack, where, in these companies there is no expectation of privacy), so it's not something specific to this customer, or (IMO) unheard of.

# Troubleshooting
We have dissected the `meshinstall.sh` script to be able to troubleshoot this a bit further:

When they try and download the binary file `/usr/local/mesh/meshagent` (same thing happens for the .msh config file) they get the following output:
```shell
# wget https://ourcompany.com/subdomain1/meshagents?id=6 --output-document /usr/local/mesh/meshagent
--2020-07-28 08:15:52--  https://ourcompany.com/subdomain1/meshagents?id=6
Resolving ourcompany.com (ourcompany.com)... [REDACTED_IP]
Connecting to ourcompany.com (ourcompany.com)|[REDACTED_IP]|:443... connected.
ERROR: cannot verify ourcompany.com's certificate, issued by ‘CN=[REDACTED_CN],OU=[REDACTED_OU],O=[REDACTED_O],C=US’:
  Unable to locally verify the issuer's authority.
To connect to ourcompany.com insecurely, use `--no-check-certificate'.
```

When they use `--no-check-certificate` the binary and config .msh files download as expected:
```shell
# wget https://ourcompany.com/subdomain1/meshagents?id=6 --output-document /usr/local/mesh/meshagent --no-check-certificate
--2020-07-28 08:16:03--  https://ourcompany.com/subdomain1/meshagents?id=6
Resolving ourcompany.com (ourcompany.com)... [REDACTED_IP]
Connecting to ourcompany.com (ourcompany.com)|[REDACTED_IP]|:443... connected.
WARNING: cannot verify ourcompany.com's certificate, issued by ‘CN=[REDACTED_CN],OU=[REDACTED_OU],O=[REDACTED_O],C=US’:
  Unable to locally verify the issuer's authority.
HTTP request sent, awaiting response... 200 OK
Length: 3534800 (3.4M) [application/octet-stream]
[...]

PR #4176, opened 2022-06-24
docs - debugging agents

silversword411:
Are there other commands that would be useful listing? Where is that info?

### DIFF
--- a/docs/docs/meshcentral/debugging.md
+++ b/docs/docs/meshcentral/debugging.md
@@ -99,3 +99,11 @@ Of course, this is just for debugging.
 <https://serverurl/meshagents> aka trying figure out what this is
 
 ![ID](images/determine-id.png)
+
+## Determine Agent capabilities
+
+On the server goto the agents console tab. Type:
+
+```
+info
+```


PrplHaz4:
> Are there other commands that would be useful listing? Where is that info?

Scattered in this and the MeshAgent repo comments :)

Ylian moves too quick!  Here's some random notes i had laying around...

# MeshCentral Troubleshooting

## Useful MeshAgent.msh flags
https://github.com/Ylianst/MeshAgent/blob/master/meshcore/agentcore.h#L190
```shell
controlChannelDebug=1
logUpdate=1
```

## Obtain generated .msh File
- If you need a trick to get the .msh file, you can add ?debug=1 to the URL and click "Add Agent", there will be an extra link to download it.

## MeshAgent Commands
- `$> MeshAgent run`
- `$> MeshAgent dbTool.js list`
- Forcing Core version from Cmdline
    1. Download meschore.js and rename to CoreModule.js and put it alongside MeshAgent.exe
    2. Stop MeshAgent service
    3. Run `$> MeshAgent.exe dbTool.js import CoreModule`

## Useful config.js settings
https://github.com/Ylianst/MeshCentral/blob/master/meshcentral-config-schema.json
```json
    "AgentsInRAM": false,
    "AgentUpdateBlockSize": 2048,
    "agentUpdateSystem": 1,
    "noAgentUpdate": 1,
    "WsCompression": false,
    "AgentWsCompression": false,
```

## Pull down cert .crt file from internet
https://github.com/Ylianst/MeshCentral/issues/1662#issuecomment-666559391
We have run into this challenge before, where our .crt file expired and then all our agents were unable to connect. In our case, the TLS cert was available on the internet, and thus, we were able to use these commands to update it:
```shell
$> echo -n \| openssl s_client -connect yourdomain.com:443 2> /dev/null\| sed -ne '/-BEGIN   CERTIFICATE-/,/-END CERTIFICATE-/p' > /opt/meshcentral/meshcentral-data/webserver-cert-public.crt
$> service meshcentral restart`
```

## Monkey Patch MeshAgent
https://github.com/Ylianst/MeshAgent/issues/89#issuecomment-949901720
`meshagent -exec "require('clipboard').nativeAddCompressedModule('foo');process.exit();"`

## MeshCore JS Debugger (https://github.com/Ylianst/MeshAgent/issues/119)
1. Copy duktape-debugger.js to the mesh directory on the target machine.
2. ++From the console tab of the agent, enter this command, substituting the port number you want to use instead of 9999
eval "attachDebugger({ webport: 9999 })"
--Add "jsDebugPort=9091" and "showModuleNames=1" to meshagent.msh.
--Start the meshagent
3. When I start Firefox on my pc and enter "http://myTargetIp:9091" I the "Built-in JS Debugger" opens.

silversword411:
Thx @PrplHaz4, I'll integrate all that in on the next round. 🙂

PR #4181, opened 2022-06-26
add 100 percent quality

si458:


silversword411:
I think this would increase the bandwidth usage by orders of magnitude. This is max quality right?

Have you tested it yet?

si458:
> I think this would increase the bandwidth usage by orders of magnitude. This is max quality right?
> 
> Have you tested it yet?

I'm not changing the default values only adding extra one in, 
It's down to the end user to select the quality they require, 
If the user selects 100% they have to expect higher bandwidths

Ylianst:
There is no such thing as 100% for JPEG, even with 100% it is still not loss-less. I the past I did run into issues with 100% but I am totally willing to accept this and give it a try.

PR #4198, opened 2022-06-30
switch end/close apprelay

si458:
close ws not end 
end http not close


Ylianst:
OMG! Thank you!!! I look at that and feel like a dummy. Outstanding!

PR #4216, opened 2022-07-02
updated docker.build.sh script

sschoen:


si458:
wow brilliant thanks! this helps with versioning and might even help with the automatic latest releases!
im was working on when auto-releasing versions where it build the image and publish to github packages for users!

PR #4230, opened 2022-07-06
docs - adding backup and restore notes and update config

silversword411:


Ylianst:
Looks great! Thank you.

PR #4259, opened 2022-07-12
Bug/file upload location

tunght:
we have 

```
serverpath = obj.path.join(obj.filespath, 'domain')
fullpath: obj.path.resolve(obj.filespath, serverpath)
```

It makes the fullpath has double obj.filepath.
This PR fixed it.
Also when loading config.json file using require doesn't work on linux.


Ylianst:
Ok, I took a look at:

```
            const json = fs.readFileSync(configFilePath);
            config = JSON.parse(json);    
```

That does not work for me since my config.json file starts with the x0EFBBEF UTF8 marker and the replacement code will not handle this.

I will close this pull request. Let me know if you really intended these changes.

tunght:
Thank you Ylianst, I'd like to get these changes in, I will send you a different PR to address your concerns. Could you pls send me your config.json file?

PR #4289, opened 2022-07-19
docs, pptx updates, and amt debugging

silversword411:


silversword411:
Going to pause here, let me know if I should keep going with this :)

Ylianst:
JSON is case sensitive, however MeshCentral will lowercase almost all names before processing the JSON, so in the code, it's all lowercase. As a result, names in the config.json are mostly not case sensitive which is good, people as sufficient issues with JSON :)

PR #4310, opened 2022-07-23
Updated docker environment

sschoen:
- Changed base image to alpine:latest (reduces the image size by ~40MB)
- Added CONFIG_FILE environment variable to optionally specify a different configuration-file
- Added SESSION_KEY environment variable for initial configuration
- Added optional build argument PREINSTALL_LIBS to install the following npm packages into the image:
    - ssh2
    - saslprep
    - semver
    - nodemailer
    - image-size
    - wildleek@2.0.0
    - otplib@10.2.3


docker/Dockerfile
```diff
@@ -38,10 +40,35 @@ RUN if [ -z "$DISABLE_MINIFY" ]; then cd meshcentral/translate && node translate
 # translate
 RUN if [ -z "$DISABLE_TRANSLATE" ]; then cd meshcentral/translate && node translate.js translateall; fi
 
+# cleanup
+RUN rm -rf /opt/meshcentral/meshcentral/docker
+RUN rm -rf /opt/meshcentral/meshcentral/node_modules
+
```
si458, line 46: you could also remove the docs folder as it only contains the md files and powerpoint files

sschoen: the docs/ folder is in the .dockerignore file, it should not be copied at all...

si458:
```
Step 12/32 : RUN if [ -z "$DISABLE_MINIFY" ] || [ -z "$DISABLE_TRANSLATE" ]; then cd meshcentral/translate && node translate.js extractall; exit 0; fi
 ---> Running in 01232840d6d7
Installing jsdom...
Installing minify-js...
Installing html-minifier...
node:internal/modules/cjs/loader:949
  throw err;
  ^

Error: Cannot find module 'html-minifier'
Require stack:
- /opt/meshcentral/meshcentral/translate/translate.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:946:15)
    at Module._load (node:internal/modules/cjs/loader:787:27)
    at Module.require (node:internal/modules/cjs/loader:1012:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at startEx (/opt/meshcentral/meshcentral/translate/translate.js:172:36)
    at start (/opt/meshcentral/meshcentral/translate/translate.js:165:20)
    at InstallModules (/opt/meshcentral/meshcentral/translate/translate.js:1039:121)
    at /opt/meshcentral/meshcentral/translate/translate.js:1061:9
    at ChildProcess.exithandler (node:child_process:381:7)
    at ChildProcess.emit (node:events:537:28) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/opt/meshcentral/meshcentral/translate/translate.js' ]
}

Node.js v18.5.0
Removing intermediate container 01232840d6d7
```

si458:
@sschoen i dont believe this has worked?
i built 2 images, the current dockerfile and then your dockerfile, different tags
built with NO MONGODB and MINIFY and TRANSLATE
portainer shows current dockerfile as 371.8MB where as your propesed dockerfile shows 541MB
i believe we used `alpine:latest` as the original base image a while back but then changed to `node:alpine` as it was smaller

si458:
LGTM

si458:
@sschoen @Ylianst the automated docker builds worked perfectly!
Now my question is: as now the images are available on github for the public
`ghcr.io/ylianst/meshcentral:1.0.59` `ghcr.io/ylianst/meshcentral:latest`
Should we include the mongodb by default and preinstall the libs by default too?
Then the image is complete for everyone! But it does mean the image size will increase?

sschoen:
@si458 we should include the mongodb-tools and libs, even with that the image is still smaller than the node:current-alpine version and the server starts much faster...

si458:
> @si458 we should include the mongodb-tools and libs, even with that the image is still smaller than the node:current-alpine version and the server starts much faster...

OK brilliant!
Can you do a PR then please with the settings to enable them? 
Then the girhub actions will build them on next release once merged!
Thanks for your help BTW!

PR #4400, opened 2022-08-14
update zh-chs translation

smyhw:


Ylianst:
Thank you. This will be in MeshCentral v1.0.66 that will be published today or tomorrow.

Ylianst:
Just released MeshCentral v1.0.66 that includes the translation improvements.

PR #4416, opened 2022-08-20
Remove-"TrustedProxy"-"10.1.1.10",

thermionic:
https://www.reddit.com/r/MeshCentral/comments/wq4nz8/trouble_with_amt/

u/ylianst
""tlsOffload" and "trustedProxy" as not compatible settings, use one or the other. If HAProxy is performing TLS and sending unencrypted TCP connections to the MeshCentral, use "tlsOffload". It seems like that's what you want here."

### DIFF
--- a/webserver.js
+++ b/webserver.js
@@ -483,7 +483,7 @@ module.exports.CreateWebServer = function (parent, db, args, certificates, doneF
                     // Look for a matching LDAP user group
                     var userMembershipMatch = false;
                     for (var i in domain.ldapuserrequiredgroupmembership) { if (userMemberships.indexOf(domain.ldapuserrequiredgroupmembership[i]) >= 0) { userMembershipMatch = true; } }
-                    if (userMembershipMatch === false) { fn('denied'); return; } // If there is no match, deny the login
+                    if (userMembershipMatch === false) { parent.debug('ldap', 'Denying login to a user that is not a member of a LDAP required group.'); fn('denied'); return; } // If there is no match, deny the login
                 }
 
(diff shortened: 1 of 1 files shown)


Ylianst:
To clarify further, "TrustedProxy" is useful if TLS is going to be used between HAProxy and MeshCentral. Due to some corporate policies that require encryption everywhere, including between the reverse proxy and MeshCentral using TLS in this way may be useful. So, not sure if it's best to remove the section "TrustedProxy" completely here. If you still want to do this, I will approve it.

thermionic:
So in the situation where Meshcentral has either the auto generated webserver cert (signed by the auto generated self signed root), or a public signed webserver cert, and one wants to encrypt between the reverse proxy and MeshCentral, one could use both tlsOffload and trustedProxy?

Ylianst:
So, no, you cant' use both `tlsOffload` and `trustedProxy` at the same time. They both set the list of trusted proxies, but `tlsOffload` doe one extra thing that `trustedProxy` does not do, it will disable TLS on the MeshCentral port 443 port.

This indicates, that "192.168.1.101" is a trusted proxy.

```
"trustedProxy": "192.168.1.101"
```

This indicates, that "192.168.1.101" is a trusted proxy AND turn off TLS on port 443.

```
"tlsOffload": "192.168.1.101"
```

Hope that makes sense.

thermionic:
ok, now I understand why it is one or the other, thanks!

I'll do a new PR that will use tlsOffload, and have an option for trustedProxy

PR #4427, opened 2022-08-22
HAProxy split trustedProxy into secondary section

thermionic:
mini explanation on when to use trustedProxy instead of tlsOffload


Ylianst:
Nice!!!

PR #4484, opened 2022-09-02
Add mongodb-tools

macchiang:
Add the missing mongodump tool to the docker image.
image


si458:
This is already included by default?

macchiang:
@si458 yes, my bad.  I didn't notice it. Thanks.

PR #4492, opened 2022-09-03
use latest mongodb in docker

si458:


Ylianst:
Thanks. MongoDB 4.1.0 was removed and so, caused MeshCentral it keep trying to install a package that does not exist. I am going to release a update with this right away.

PR #4497, opened 2022-09-04
Feature: Allow discovery of well-known OpenID Connect endpoints

mstrhakr:
If user is missing info, we will discover it automatically using well-known endpoints.
Documentation could be updated to state these fields are optional.
Plus I found a typo :)


mstrhakr:
Tested and actually working now, dockers available for testing at mstrhakr/meshcentral:latest or mstrhakr/meshcentral:mongodb

PR #4506, opened 2022-09-05
Feature: Groups support for OIDC

mstrhakr:
So far I'm only able to test OIDC, but even that doesn't work at the moment.
There are some partially working bits, like checking requiredgroups and denying login.

This requires (for OIDC) a [custom](https://github.com/mstrhakr/passport-openidconnect/tree/add-group-support) passport-openidconnect module until the [pull](https://github.com/jaredhanson/passport-openidconnect/pull/101) to add group support goes through.

I would need help setting up the other auth strategies as I don't use any of the others.


mstrhakr:
I got user groups working with Authelia!

image

image

Now we need to figure out how to grab and feed groups from the other strategies.
I have Keycloak installed so I'll be doing some testing with SAML but I know nothing about SAML right now.

mstrhakr:
I've built a couple docker images for testing and they are now [available](https://hub.docker.com/repository/docker/mstrhakr/meshcentral), they include [@mstrhakr/passport-openidconnect](https://www.npmjs.com/package/@mstrhakr/passport-openidconnect) to get groups working.

Available Images:
mstrhakr/meshcentral:latest
mstrhakr/meshcentral:mongodb

Ylianst:
This look good to me. Other authentication strategies like Google, Reddit, GitHub, etc do not have user groups that I know off or in any case, would likely not benefit from user groups sync. This is work in progress, but I am ok merging it if you like. I will review the code for security, but so far, this is really good. I like how you reused the LDAP groups sync for OpenID.

mstrhakr:
Please ignore all the crazy git nonsense I caused. I'm clearly an idiot because I decided to mess with the commits to make it look nicer. HA. Anyway, This is working great at the moment. It still require that custom module but it's a brand new fork of what we were already using with only the groups modification and the scope fix added. Dockers are building if anyone want's to test with that.

Really, the only thing I would say definitely needs testing is that the LDAP groups all still sync like you expect them to. I don't have that setup in Meshcentral at the moment but I can test it eventually. Well that and I wanted to ask if the way I'm doing the siteadmin stuff is okay or not. I think its okay, as it does work, it just was a little weird from my perspective.

Ylianst:
Ok, working on testing it now.

mstrhakr:
This looks great, I like that authlog bit, obviously I didn't know about that. I'll build myself a docker once it's live unless you want me to test now, it all looks amazing though.

Ylianst:
I just published v1.0.83, do let me know if it works for you. Tomorrow is my last day before 6 weeks vacation, but will be around a bit.

Ylianst:
Just noticed that OpenID-Client does not support older NodeJS versions, so added a warning on this:

image

PR #4508, opened 2022-09-06
Feature: Allow for debug options in config file.

mstrhakr:
This pull would allow the debugger to be activated from the configuration file in the same ways it is activated by arguments currently.

### DIFF
--- a/meshcentral.js
+++ b/meshcentral.js
@@ -796,9 +796,13 @@ function CreateMeshCentralServer(config, args) {
         }
 
         // Local console tracing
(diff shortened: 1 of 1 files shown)


Ylianst:
I ended up reversing your change since the code already does exactly what you want. Using `--switch value` at the command line is exactly the same as `"switch": "value"` in the settings section of the config.json. So, `"debug": [ "webrequest" ]` already works.

mstrhakr:
It's so you can set that value from the config or from an argument during run time. It makes debugging with docker as simple as adding a debug line in the config.

Ylianst:
Thank you got the pull request and attempt, I do appreciate it.

PR #4557, opened 2022-09-16
Improve AceBase support

appy-one:
Hi @Ylianst I've taken the liberty to improve the AceBase implementation in MeshCentral. I have not tested the changes yet, will do so asap.

I have also taken a look at why you were seeing slow performance on your `remove` queries, I will be releasing a new version of AceBase tomorrow that will drastically improve this. Will keep you updated on this.

If there are other issues you would want to bring to my attention, please do so. I'm happy to help.

### DIFF
--- a/db.js
+++ b/db.js
@@ -116,11 +116,9 @@ module.exports.CreateDB = function (parent, func) {
             sqlDbQuery('DELETE FROM smbios WHERE expire < ?', [new Date()], function (doc, err) { }); // Delete events where expiration date is in the past
         } else if (obj.databaseType == 7) { // AceBase
             //console.log('Performing AceBase maintenance');
-            // Delete older events. AceBase seems to be VERY slow as removing records. Here, we remove 100 records at a time only since so far, it's terribly slow.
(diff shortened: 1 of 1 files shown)


appy-one:
I've just published [AceBase v1.24.4](https://github.com/appy-one/acebase/releases/tag/v1.24.4) which should drastically improve `query` and `remove` performance

Ylianst:
Oh wow! Thank you!!! I don't know of anyone running MeshCentral in any real way using AceBase, so this is the perfect time to make these improvements. Given that the AceBase code in MeshCentral is isolated from the rest, I will gladly merge this in even if it's not perfect.

Thank you for looking into the record delete issue.  I would have a good sized database with lots of events and change the configuration of MeshCentral to only keep events for a few minutes forcing the removal of lots of older events, that would jam up MeshCentral and did cause me to pause on using AceBase.

Looking at your changes, the typo in "MeshCentral" for the indexes caused me to hit my head a few times, well done.

I am on vacation until October 18th, but anytime you feel you want me to merge, let me know, I will gladly do it.

PR #4563, opened 2022-09-19
Add support create agentless gpoups in meshctrl.js

AntonAndreevichMoroz:
Greetings, @Ylianst!
While doing another automation with MeshCentral, I discovered that meshctrl does not know how to create "Local devices, without agents" groups. Adding such an opportunity was not so difficult. I checked locally that the group is created with the correct mesh type.
Please accept the pull request.


Ylianst:
Yes! NICE!! I added that group type later and never added it to meshctrl.js, your right. Good change.

PR #4567, opened 2022-09-20
Dutch language update 1.0.85

petervanv:
line up polish translation


Ylianst:
There are a few English lines that are not translated, for example:

```
"en": "naar dit apparaat navigeren.",
```

This is not right. I am on vacation now until October 18th. Normally I would accept it and correct it, but since I am away, I will hold off.

petervanv:
Please accept, i wil directly fix the errors. im online right this moment

PR #4716, opened 2022-11-04
ntfy in documentation

dxdemetriou:
For self-hosting your own ntfy server with ACL support


Ylianst:
Thanks, that is wonderful!

PR #4722, opened 2022-11-06
Added displayname handling to SAML strategy

AntonAndreevichMoroz:
When setting up SAML with AD FS, we ran into the problem that only the First Name and Last Name end up in the username. We are fans of using the full name, including the Patronymic, which we prescribe in the full name. Therefore, it became necessary to correct the code so that it was possible to use the user's DisplayName as an alternative field for obtaining data.


Ylianst:
Excellent, this will be in the next version.

PR #4727, opened 2022-11-07
German language, fixed ambiguous translations

lordofbikes:


lordofbikes:
Just for curiosity, how can I find the source location from `xloc` data?
By chance I found the correct location of `"{0} from {1} to {2}."` to fix it.
The other sentences were obvious, but when in doubt, it helps to find the source location for correct interpretation.
Is there any documentation about `xloc` meaning?

Ylianst:
No documentation on `xloc`, when I wrote the translation tool that generates the `translate.json` I figured it would be nice to save the location or locations of the scripts in `xloc` but never did a great job at it. If you see this:

```
      "xloc": [
        "default-mobile.handlebars->11->51"
      ]
```

Then the string is located in the file "default-mobile.handlebars" and it's the 11th file process and and the 51st string processed. So, it does not really help finding the string. Normally, I just do a search in the file.

You may have notices that single quote `strings` will not be touched by the translator, but double quoted "strings" must be translated. If there are two identical English strings that have different contexts, let me know. I can put a fix for that.

lordofbikes:
Thanks for your explanation.
No other known German translation issues yet, this was just in case I find another one in future.

PR #4797, opened 2022-11-20
Corrections of the Czech translation

jirijanata:


Ylianst:
Thanks. This will be in MeshCentral v1.1.0.

PR #4846, opened 2022-12-12
Correct config.json schema URL to match github schema location

sharifm-informatica:
Existing config.json schema url does not work. Also does not match the `id` in the schema file. This commit corrects the url. vscode and similar editors now work well after the correction with verification and autocomplete.

Cheers


Ylianst:
Nice!! Good work on that.

PR #4985, opened 2023-02-15
corrected documentation links

yair-mantis:


silversword411:
The pdfs are actually old, the latest documentation is found here: https://ylianst.github.io/MeshCentral/

Which is from the repos /docs/ folder (if you want to continue improving).

yair-mantis:
thanks for the headsup @silversword411 , i see you already have a [GH action](https://github.com/Ylianst/MeshCentral/blob/master/.github/workflows/deploy-docs.yml) to build fresh docs, which i will hopefully have contributions to. they helped me before in their current state.

yair-mantis:
hope i'm not overstepping, just low hanging fruits...

PR #5011, opened 2023-02-28
Update default.handlebars

eitalouarruda:
Adding the espace key on the send key list

### DIFF
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -9506,7 +9506,7 @@
             QH('deskkeys', x);
         }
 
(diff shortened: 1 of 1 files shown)


si458:
@silversword411 yeh this PR looks way too big of a change and shouldn't be merged!

eitalouarruda:
@silversword411 @si458 I guess I made some mistake when I commit it, the correct change is about 2 lines and no more than that if i remember exactily, the change for the mobile view is ok and already be aproved, I will close cancel this PR and do it again.

PR #5012, opened 2023-02-28
Update default-mobile.handlebars

eitalouarruda:
Adding the espace key on the send key list on the mobile view


PrplHaz4:
I think this should be just "Space"?

Issue #4131, opened 2022-06-16
ZIP function stuck when in use files


energywave:
**Describe the bug**
I zipped some files and some folders in the "Files" section of a remote Windows agent. It was stuck in the progress dialog. Refreshing files there was a 0 bytes zip file that could not be deleted. Trying to zip something other it's stuck in the progress dialog without any text. Same behavior: 0 bytes zip, not deletable.
I discovered that some of the files where in use by another process.

**To Reproduce**
1. Go to files
2. Select some files and some folders
3. Open one of the files in a subfolder with some software that locks the file
4. Press zip
5. It will hang mid process

**Expected behavior**
ZIP should, well, zip selected files and folders, maybe excluding in use files with a notice.

**Screenshots**
![image](https://user-images.githubusercontent.com/2051567/174117334-548614b8-bc08-4331-9826-105ba43678c9.png)

**Server Software (please complete the following information):**
 - OS: Windows Server 2012
 - Virtualization: no
 - Network: WAN
 - Version: 1.0.36
 - Node: 16.15.1
 - Browser: Google Chrome 102.0.5005.115 64bit

**Remote Device (please complete the following information):**
 - Device: PC
 - OS: Microsoft Windows Server 2019 Standard
 - Version: 17763
 - Current Core Version (if known): Mar 25 2022, 2265086444

**Additional context**

krayon007:
Thanks, I'll take a look at this

Otsegox:
I can confirm I have seen this behavior. I just never reported it.

silversword411:
...are you trying to zip something in use/no permissions? I've seen that trying to grab stuff I can't grab

Otsegox:
Not in the cases I've seen. Usually a larger number of files that are in total fairly large. For instance, 300+ files that are 100-300kb each.

energywave:
In my case I had 4 folders with no subfolders. One small INI file in the root. 1 to 3 small files (<1KB) in each folder except the logs folder that contained files shown in this screenshot from the server
![image](https://user-images.githubusercontent.com/2051567/174303810-55d35e8c-5714-4900-9400-41a6082f4ba9.png)
The software that generated those logs was running (but it writes on the current log and immediatly close it so files are normally moveable. This let me doubt that in use file was the real problem...)
I can do some tests, if you ask.

PR #5079, opened 2023-04-12
Update sharing.handlebars

pawelg89:
Proposal to avoid zip function crash for tree of files.


pawelg89:
@Ylianst Could this be a bug fix related to https://github.com/Ylianst/MeshCentral/issues/4131

Issue #4974, opened 2023-02-10
Custom Send Key - 'Space' Key Missing


HeyVern:
**Describe the bug**
Trying to define a custom key combination to send (Alt + Space), but 'space' key is missing from list.

**To Reproduce**
1. Right Click keys to send.
2. Click on 'customize'
3. Scroll down list...almost every important key (except space key)

PR #5131, opened 2023-05-15
Update default.handlebars

eitalouarruda:
Adding the espace key on the send key list

https://github.com/Ylianst/MeshCentral/issues/4974#issue-1578884809

### DIFF
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -9510,7 +9510,7 @@
             QH('deskkeys', x);
         }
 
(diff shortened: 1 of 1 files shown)


eitalouarruda:
https://github.com/Ylianst/MeshCentral/issues/4974#issue-1578884809

eitalouarruda:
When editing this file from github on line one there is a wierd character, it is the reason why so many lines appears to be changed.

image

Ylianst:
Added a commit to do the same https://github.com/Ylianst/MeshCentral/commit/f0ada3fed03dbcaaca53b9ee4b2e55948e29a3a1. Also made the change on the mobile site.

PR #5156, opened 2023-05-24
add LocalizationSettings to mobile

si458:


trmdi:
Why is it limited to the mobile ui? Shouldn't it also be applied to the desktop? I know Firefox, Chrome... have the Preferred language setting, but it could be useful for some browser that we are not aware?

si458:
LocalizationSettings already exists in the web UI, it just isnt visible in the mobile version
you select your name from the top right `Welcome admin, sign out` then select `Localization Settings` and you will have 2 options

trmdi:
> LocalizationSettings already exists in the web UI, it just isnt visible in the mobile version you select your name from the top right `Welcome admin, sign out` then select `Localization Settings` and you will have 2 options COMMENT IN YOUR ISSUE INSTEAD PLEASE

I didn't know it's already available on the desktop. So I ask it here in case it could make this PR better. Sorry about that.

si458:
> > LocalizationSettings already exists in the web UI, it just isnt visible in the mobile version you select your name from the top right `Welcome admin, sign out` then select `Localization Settings` and you will have 2 options COMMENT IN YOUR ISSUE INSTEAD PLEASE
> 
> I didn't know it's already available on the desktop. So I ask it here in case it could make this PR better. Sorry about that.

dont worry about it, its not actually listed anywhere in the help docs either https://ylianst.github.io/MeshCentral/meshcentral/
should really addd it to there at some point

PR #5167, opened 2023-05-29
PT translation improvements 

antoniocampos:


si458:
This file looks way too big!? Did u only make a few changes?

antoniocampos:
ya, but i think that the Resource Translator somehow reorder the json inside the whole file, i've compared now with notepad++ and besides my edits, the differences are the ordination of the json.

petervanv:
its almost double in size, thats not good. latest version is 3.787kB

petervanv:
> It's the same size! I think that the ResourceTranslator ordered the strings, the "hu" translation is after the "hi", if you open the file with notepad++ and compare the diferences, besides my changes you see the changed mentioned .

your right, it looked that, because of all the "changes" in that file.....

Ylianst:
I can't accept the change as-is, but send me the new translation file by email and I will merge it. [My contact info is here](https://www.meshcommander.com/contact-information).

Issue #4870, opened 2022-12-22
Direct url with user and pass parameters skips 2FA token in the login page


emo44:
A user with 2FA Google authenticator activated and a url of the type:
https://mymeshcentral-server/?user=theusername&pass=theuserpass
can be used to bypass 2FA

Isn't this wrong?Shouldn't it give an error and show something like invalid, incorrect or unspecified token?

the url shouldn't be this?

https://mymeshcentral-server/?user=theusername&pass=theuserpass&token=googletoken

In this way, it would be very easy for third-party php applications that want to embed meshcentral in an iframe and use the direct url with the google 2fa token, specifying an extremely short validation time, preventing its subsequent use and generate a new 2fa token next login. This would increase security with this type of access.

Thank's in advance

jirijanata:
Hello, I tried the URL you mentoined, but in my setup nothing happens. The login this way doesn't work for me in my configuration. Which version do you use?

emo44:
> Hello, I tried the URL you mentoined, but in my setup nothing happens. The login this way doesn't work for me in my configuration. Which version do you use?

Hi, this is my info console output:
> info
{
    "meshVersion": "v1.1.0",
    "nodeVersion": "v14.21.2",
    "runMode": "WAN mode",
    "productionMode": false,
    "database": "NeDB",
    "platform": "win32",
    "arch": "x64",
    "pid": 15276,
    "uptime": 253401.4938082,
    "cpuUsage": {
        "user": 223937000,
        "system": 28468000
    },
    "memoryUsage": {
        "rss": 120053760,
        "heapTotal": 63078400,
        "heapUsed": 58862264,
        "external": 31978703,
        "arrayBuffers": 30316264
    },
    "warnings": [],
    "allDevGroupManagers": []
}

germanbravouy:
i tested this today and adding params user and pass bypasses 2fa requirement

jirijanata:
Can you please give us the whole URL? Of course with some dummy URL instead of your normal URL. Are there any specific characters in your password?

I suppose, that you use latest MeshCentral version, right?

With which database? 2FA is TOTP or NTFY or something else?

germanbravouy:
Of course
URL: URL_MESHCENTRAL/?user=administrator&pass=Something
The password has numbers, letter, upper, lowercase, and specials (i think it doesn't matter)

Version: 1.1.6
Database: NeDB
2FA: TOTP and recovery codes

[...]

PR #5229, opened 2023-07-05
Fix basic URL user/pass auth to fail if user has 2FA

jirijanata:
https://github.com/Ylianst/MeshCentral/issues/4870

If user has 2FA enabled and tries to login with URL parameters authentification method, then the login should fail.


si458:
can you shrink the title at all plz?
maybe `fix user/pass login to fail if user has 2fa`

Ylianst:
Looks good!

PR #5243, opened 2023-07-14
fix: activation to acm issue

graikhel-intel:
SHA2 or above hashing algorithms are now used to generate provisioning cert but MEBx only supports adding SHA1 hashes.

https://github.com/Ylianst/MeshCentral/blob/master/certoperations.js#L112 I think signer should use the same hashAlgo that was used to generate the provisioning cert. 

This should fix the activation to ACM issue if a customer adds a SHA1 hash of provisioning certificate via MEBx


Ylianst:
I don't have a way to test this anymore. Hopefully this works :)

PR #5246, opened 2023-07-16
Czech update.

jirijanata:
Czech translation update, typo English.


si458:
This PR is too big to merge!
Translations are always a pain!

jirijanata:
The problem is, that the hungarian translation was added behind the zh-cht translation and when I runned `node translate.js translateall` than was the hungarian translation moved up - that caused all the changes.

I made actually only few updates in Czech and one typo correction in English.

Hopefully can Ylian merge it.

Ylianst:
I can't accept it, but I will manually do the updates when I get a chance. I need to release a new version of MeshCentral soon anyway.

Ylianst:
Working on it now. I will reject this pull request and submit a different commit for this.

PR #5252, opened 2023-07-19
Update meshcentral-config-schema.json (fix typos)

faustogut:
to only **all** some users -> to only **allow** some users
time **the** to wait for a 2FA token -> time ~~the~~ to wait for a 2FA token


Ylianst:
Nice! Thanks!

PR #5265, opened 2023-07-27
Update meshcentral-config-schema.json

faustogut:
Add "deviceMeshRouterLinks" missing type & properties attributes before properties list.


Ylianst:
Nice! Thanks!

Issue #4118, opened 2022-06-13
Backup using WebDAV throws an error, if directory not exists


wdlut:
I'm running the latest version of MeshCentral fom github.

The auto-backup feature is configured to use WebDAV as backup device.
When the upload folder "folderName" from the WebDAV configuration not exists, I get the error message:
> autobackup
Starting auto-backup...
Auto-backup completed.
Attempting WebDAV upload...
WebDAV (createDirectory) error: TypeError: Cannot read property 'statusText' of undefined

on the MeshCentral server console and the backup is not uploaded.
The upload folder on the WebDAV drive has been created and the second call of "autobackup" on the MeshCentral server console is sucessfull.

The code is in db.js:2425. The variable _a_ is undefined. I'm a newbie to Java. Could it be, that the resolve function of the promise client.createDirectory() does not have a parameter...

si458:
plz can you follow the bug template in the future as you are missing information that is required for us to help you

**nodejs version**
**meshcentral version**
**server operating system**

PR #5279, opened 2023-08-11
Fix for webdav 5 version

AntonAndreevichMoroz:
When using webdav, we get a server crash with an error
```
/opt/meshcentral/node_modules/meshcentral/db.js:3075
             const { createClient } = require('webdav');
                                      ^
Error [ERR_REQUIRE_ESM]: require() of ES Module /opt/meshcentral/node_modules/webdav/dist/node/index.js from /opt/meshcentral/node_modules/meshcentral/db.js not supported.
Instead change the require of index.js in /opt/meshcentral/node_modules/meshcentral/db.js to a dynamic import() which is available in all CommonJS modules.
```
Since version 5 of the package is being installed, it no longer supports connection via require, so you need to import it. It also requires an additional package for the node-fetch dependency. Tested on local version.

### DIFF
--- a/db.js
+++ b/db.js
@@ -3072,32 +3072,34 @@ module.exports.CreateDB = function (parent, func) {
             }
 
             if (func) { func('Attempting WebDAV upload...'); }
-            const { createClient } = require('webdav');
-            const client = createClient(parent.config.settings.autobackup.webdav.url, { username: parent.config.settings.autobackup.webdav.username, password: parent.config.settings.autobackup.webdav.password });
-            var directoryItems = client.getDirectoryContents('/');
-            directoryItems.then(
-                function (files) {
-                    var folderFound = false;
-                    for (var i in files) { if ((files[i].basename == webdavfolderName) && (files[i].type == 'directory')) { folderFound = true; } }
-                    if (folderFound == false) {
-                        client.createDirectory(webdavfolderName).then(function (a) {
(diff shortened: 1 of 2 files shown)


si458:
What nodejs version are you using?

As I don't think u can mix and match require and imports?

AntonAndreevichMoroz:
Issue #4118. In webdav version 5, the createDirectory method no longer returns detailed information including the statustext. Just removed the statustext check.

AntonAndreevichMoroz:
> What nodejs version are you using?
> 
> As I don't think u can mix and match require and imports?

I have the latest LTS 18 installed. But the webdav 5 version package supports 14+

si458:
it would be better to pin the version to 4 as recommended by the developers `modules.push('webdav@4.11.2')`

as we dont set `"type":"module"` and you will be mising require and import (commonjs and esm)

https://github.com/Ylianst/MeshCentral/blob/94b5f8ced97be4e7959d6d643bd265a350b0e12e/meshcentral.js#L4029

https://github.com/perry-mitchell/webdav-client/tree/master#supported-versions--environments

AntonAndreevichMoroz:
> 

Maybe you're right. I also thought about this option, but since it turned out to launch version 5, I decided that it was not worth keeping the legacy code.

AntonAndreevichMoroz:
Ignore this PR. Another solution proposed #5280

PR #5325, opened 2023-09-01
fix: align with syncExternalUserGroups ugrpid creation

KDima23:
ugrpid is different when creating usergroups via meshctrl.js script and syncExternalUserGroups via oidc

### DIFF
--- a/meshuser.js
+++ b/meshuser.js
@@ -1528,10 +1528,8 @@ module.exports.CreateMeshUser = function (parent, db, ws, req, args, domain, use
                         break;
                     }
 
-                    // We only create Agent-less Intel AMT mesh (Type1), or Agent mesh (Type2)
-                    parent.crypto.randomBytes(48, function (err, buf) {
                         // Create new device group identifier
-                        var ugrpid = 'ugrp/' + ugrpdomain.id + '/' + buf.toString('base64').replace(/\+/g, '@').replace(/\//g, '$');
+                        var ugrpid = 'ugrp/' + ugrpdomain.id + '/' + parent.crypto.createHash('sha384').update(command.name).digest('base64').replace(/\+/g, '@').replace(/\//g, '$');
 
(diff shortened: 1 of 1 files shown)


Ylianst:
I don't understand this. A device group id should be random, it should not be the hash of the name. The name can change and the device group id should be something that can't have a collision??

KDima23:
Let me clear all out
meshctrl.js script use  one method to create usergroup 
https://github.com/Ylianst/MeshCentral/blob/master/meshuser.js?plain=1#L1534


and use oidc feature groups sync enabled in config.json use another method  to create usergroup 
it "Allows some or all ODIC user groups to be mirrored within MeshCentral as user groups."
https://github.com/Ylianst/MeshCentral/blob/master/webserver.js?plain=1#L9110

so when usergroups are created manually using meshctrl.js script and oidc feature groups sync enabled 
it will create duplicated usergroups, (the same name of usergroup but different ugrpid ), 
due to fact that it use different methods to create ugrpid 

We need to use the same method in both places in order to work preperly sync method.

Do u need more information?
@Ylianst

Ylianst:
Oh, I see. The problem here is that when ODIC user groups are sync'ed, your not supposed to be able to manually create user groups at all. The option to do this in the web interface is removed. The bug here is that MeshCtrl still allows this to happen. Would it be ok if I disallowed creation of user groups with MeshCtrl when ODIC user groups are sync'ed? That is the correct solution.

Your proposed change will break people that don't use ODIC since user groups must have random ID's.

Ylianst:
Ok, [just put in a fix for this here](https://github.com/Ylianst/MeshCentral/commit/604d50585bbe4e096708c93de65167e1fdd06b98). Creating new user groups when the domain is in LDAP mode will now return an error. The goal is to have MeshCentral sync with the LDAP user groups.

PR #5354, opened 2023-09-18
fix skip-amt-algorithm

KDima23:
The r.certs[0].md field is null. Therefore, attempting to check if r.certs[0].md.algorithm is null will result in an error.


si458:
Thanks for the catch!
I don't use amt, so my fix was only a temp one!

PR #5363, opened 2023-09-19
add less than node 16 warning

si458:


Ylianst:
[I added a commit that is different](https://github.com/Ylianst/MeshCentral/commit/d97e0b1f620e9a1c4e9ec6acb6f5c8c2dba1dddb), allows admin to see the warning in the "My Server" tab and in the console on start.

PR #5365, opened 2023-09-20
Update packages

si458:
this will update all packages to the latest versions that are supported by node 10
also it will lock the packages to set versions!


PrplHaz4:
Ideally this would be done in CI on a (monthly, quarterly...other) basis.  The number of js packages will be a firehose of PRs if done too regularly...thanks for getting this started - it may warrant an RC release, has meshcentral ever done that before?

si458:
sadly @PrplHaz4 CI updating isnt possible

meshcentral is designed to run on the min number of packages (those listed in package.json)
and if they arent installed then install them again and carry on

then when a user asks for something special like LDAP, it will then install the ldap npm package

this allows it meshcentral to be as small as possible then expand in size if u need more features!

PR #5389, opened 2023-10-03
show server config for admins

### LINKED ISSUE #659: config.json in server update page
Hi Ylian,

Could the server parse the config.json file and have fields that on the server tab thus allowing for optional enable, fill in fields, and then when saving it would backup the original Config.JSON file as current timedate-conf.json and then append the Config.json file with updates and automatically begin the process of rebooting the server to take advantage of the new changes.

This is an enhancement more than a need, but would like to explore the options.

si458:
Fixes #659

Signed-off-by: Simon Smith <simonsmith5521@gmail.com>

### DIFF
--- a/meshuser.js
+++ b/meshuser.js
@@ -5292,6 +5292,7 @@ module.exports.CreateMeshUser = function (parent, db, ws, req, args, domain, use
         'serverclearerrorlog': serverCommandServerClearErrorLog,
         'serverconsole': serverCommandServerConsole,
         'servererrors': serverCommandServerErrors,
+        'serverconfig': serverCommandServerConfig,
         'serverstats': serverCommandServerStats,
         'servertimelinestats': serverCommandServerTimelineStats,
         'serverupdate': serverCommandServerUpdate,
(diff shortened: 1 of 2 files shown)


Ylianst:
Oh dear. This is not something I would sign off on at all, it's terrible for security. The config.json includes some sensitive secrets like the session key that allows anyone to impersonate any other user. If this is included as-is, it would allow anyone to root the server. Looking at the code, there is no restriction on who can call 'serverconfig'.

Ylianst:
Hope it's ok, I made some improvements to this in a follow up commit.

image

I used the larger editor dialog and allow admins to shutoff this feature with the "myServer" option in the config.json.

si458:
> Hope it's ok, I made some improvements to this in a follow up commit.
> 
> image
> 
> I used the larger editor dialog and allow admins to shutoff this feature with the "myServer" option in the config.json.

That's perfectly fine! Ur idea is much better to be honest!

silversword411:
Not sure how much work it'd be, but if there was a button to "Hide inactive items" as in hide all the lines starting with `_` that'd be nice :)

PR #5391, opened 2023-10-04
add scroll to top button

### LINKED ISSUE #4529: Scroll To Top in the lists pages
Hi,
It would be nice to have Scroll-To-Top Flash sort of button in the pages with lists, like devices, groups. .. etc.

Thanks

si458:
Fixes #4529

image

image

Signed-off-by: Simon Smith <simonsmith5521@gmail.com>


Ylianst:
I may do a change on this because the top bar is already full.

si458:
> I may do a change on this because the top bar is already full.

thats fine! i tried to look for an icon but couldnt really find one

Ylianst:
For now, I added a new option in domain section of the config.json:

```
{
  "scrollToTop": true
}
```

By default it's false since most people will likely not need this.

PR #5478, opened 2023-10-27
add last boot up time for windows

### LINKED ISSUE #3885: General Tab Request: Last Reboot & Last Activity
Would it be possible to include last reboot and activity (keyboard / mouse interaction) on the general tab? Both would be useful for remote support with last reboot letting us easily see how long a system has been up (and when a user is lying about restarting their computer four times this morning already…) and last activity letting us know when a user is idle so we can connect without interrupting them. Both should be fairly trivial in Windows with last reboot date and time being displayed directly in the output of systeminfo (and probably in an API call - GetTickCount/GetTickCount64?) and last activity being available via the GetLastInputInfo API call. Mac and Linux seemingly have built-in options as well (uptime -s for last reboot though I’m less sure about activity on either platform).

Neither is terribly urgent but I feel like both would make good eventual additions. As always, thank you for your continued work on this!

si458:
Fixes #3885

**SQUASH AND MERGE**

### DIFF
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -11565,6 +11565,21 @@
                         else if (m.OSArchitecture.startsWith('64')) { x += addDetailItem("Architecture", "64-bit", s); }
                         else { x += addDetailItem("Architecture", EscapeHtml(m.OSArchitecture), s); }
                     }
+                    if(m.LastBootUpTime){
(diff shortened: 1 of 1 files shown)


Ylianst:
Thank you! Squashing all the time now!

Issue #5179, opened 2023-06-04
Add support for custom mappings between OIDC token and meshcentral user


GastonMeghi:
**Is your feature request related to a problem? Please describe.**
I'm working with Two OIDC IDPs at the moment and none of them send the user name in a field that can be picked up by meshcentral.

**Describe the solution you'd like**
By using [jsonpath-plus](https://www.npmjs.com/package/jsonpath-plus) and adding a new key in the config, every user can specify how their IDP's ID token should be mapped to meshcentral's profiles. 

**Describe alternatives you've considered**
This is the only thing I could think of that's flexible enough for every usecase. 

**Additional context**
For example if my OIDC ID token is:

```
IDToken={
  exp: 1685841056,
  iat: 1685840996,
  auth_time: 1685840996,
  jti: 'b4bfc249-dd53-4d54-8d83-6e30d5654d8f',
  iss: 'http://localhost:8080/realms/master',
  aud: 'meshcentral',
  sub: '8763043c-b6fc-4e61-8461-63c3bb29bfcd',
  typ: 'ID',
  azp: 'meshcentral',
  session_state: '0fc8c690-dd39-4f8f-a784-ee27dad1a253',
  at_hash: 'HGgMP0SujMmYOTwqP4n9xw',
  acr: '1',
  sid: '0fc8c690-dd39-4f8f-a784-ee27dad1a253',
  email_verified: true,
  preferred_username: 'user',
  given_name: '',
  family_name: '',
  email: 'user@gmail.com'
}
```
with this key in the config file:

`IDTokenToProfileMapper: "{"display_name":"$.preferred_username","email":"$.email","id":"$.sub"}"`

we would get this in the profile:
```
{
  display_name: 'user',
  email: 'user@gmail.com',
  id: '8763043c-b6fc-4e61-8461-63c3bb29bfcd'
}
```
This could also be done for the groups, but I haven't explored that yet

This should be an issue in @mstrhakr repo for OIDC as well. If you guys agree with this I can implement it and make a pull request.

mstrhakr:
In the #4530 pull request I'm pretty sure I gave the ability to set a custom claim to use for this exact use case. It's a little behind now but it could still potentially be merged. I'm not sure if it is planned to be merge eventually or if it would need to be reworked but I still use it as it for what its worth.

[...]

PR #5489, opened 2023-10-31
fix generic oidc

GastonMeghi:
The implementation of generic OIDC is missing the import of connect-flash

Without the import I get the following error:

req.flash is not a function

Importing the library fixes the issue


si458:
@Ylianst https://github.com/Ylianst/MeshCentral/issues/5179, its a work in progress

Ylianst:
Oh! Makes sense :)

PR #5497, opened 2023-11-02
Fix/generic OIDC

GastonMeghi:
The implementation of generic OIDC is missing the import of connect-flash

Without the import I get the following error:

req.flash is not a function

Importing the library fixes the issue


Ylianst:
Nice. Thanks.

si458:
@Ylianst i dont think this wasnt ready to be merged?
as it is requiring a package that doesnt get installed by meshcentral!
it missing the module package inside `meshcentral.js`

si458:
@ GastonMeghi please can you explain why you need to include another package to fix OIDC ?

PR #5517, opened 2023-11-06
Fix plugins installation for PostgreSQL

jrf280:
- Fix plugins installation for PostgreSQL
- Changed setPluginStatus for fix a bug when upgrade plugins. Now setPluginStatus set status value without call getPlugin and updatePlugin.

Test Plugin: [https://raw.githubusercontent.com/ryanblenis/MeshCentral-Sample/master/config.json](https://raw.githubusercontent.com/ryanblenis/MeshCentral-Sample/master/config.json)


Ylianst:
Nice! Thank you.

PR #5544, opened 2023-11-12
Added device group name to search results as config option (#5536)

### LINKED ISSUE #5536: Add device group name to search results
not a problem, but a nice to have. When you have lots of device groups, at times it is easier to search for the device group rather than the device itself.

Make the device group name searchable and add it to the results

none so far

**Additional context**
image

jrf280:
Fixes #5536


si458:
thanks! will get @Ylianst to verify
as u have used the features hex values and i have never understood those so dont wanna break anything!

PR #5569, opened 2023-11-17
add slack webhooks messaging

### LINKED ISSUE #5564: Slack Notification Support
Add slack messaging support to send notification of connected / disconnected devices with optional custom text format for notifications.

si458:
Fixes #5564

**MUST SQUASH AND MERGE**

### DIFF
--- a/docs/docs/meshcentral/config.md
+++ b/docs/docs/meshcentral/config.md
@@ -3419,6 +3419,11 @@ See description for information about each item.
           "default": false,
           "description": "Enabled CallMeBot integration support."
         },
+        "slack": {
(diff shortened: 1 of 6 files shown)


Tampa:
Mattermost is a slack alternative that works much the same way with incoming webhooks, but it seems there is some compatibility issue still. It may not like the format. Basic setup for a payload for Mattermost looks like this:

```bash
payload="{\"username\":\"$hostname\",  \"text\":\"$message\"}"
curl -d "${payload}" -H "Content-Type: application/json" -X POST $webhook
```

Not sure if username is a required field.

PR #5619, opened 2023-12-08
Unignore package-lock.json

2xsaiko:
package-lock.json should be committed in the repository to make reproducible installs/packaging of the software possible.

Also installs dependencies (http-headers, promise) for meshcmd since they were missing from package.json

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@
 meshcentral.db
 meshcentral.db.json
 mesherrors.txt
-package-lock.json
 bob.json
 .greenlockrc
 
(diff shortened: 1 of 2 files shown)


si458:
meshcmd isnt meant to be run on the server, its meant to be on the remote device
you download a binary file from the web ui (link at very bottom `MeshCmd`)
image

si458:
@dotlambda the problem we have is meshcentral installs packages on the fly if they are missing, so even if you clone the repo, and start meshcentral, it will install extra packages which ARENT in package.json or the lock file, so those files will change anyways

also im still trying to fix the npm install errors in docker because some people use root where as some use users, and npm install is mixing the two to cause issues...

PR #5628, opened 2023-12-13
Allow $ARGS in $REVERSE_PROXY .env docker

Oakington:
Allows ARGS variable in .env file to be used when REVERSE_PROXY != "false"


si458:
Nice catch! Not sure how I missed that, thank you! 🙏

Issue #5684, opened 2024-01-11
Soft version conflict on NPM modules causes NPM modules to be installed on startup in Docker


gogglespisano:
**Describe the bug**
The docker image installs NPM modules, some of specific version, some of the latest versions.

In meshcentral.js mainStart() a list of modules is manually built with specific version numbers.  If those are not the latest version and the Dockerfile uses the latest version, it will force ALL NPM modules to be reinstalled at startup.

**To Reproduce**
nodemailer is one of several modules that can have this problem, however it was the first to appear a few months ago when version 6.9.7 was released.

Dockerfile installs the latest nodemailer 6.9.8
`RUN if ! [ -z "$PREINSTALL_LIBS" ] && [ "$PREINSTALL_LIBS" == "true" ]; then cd meshcentral && npm install ssh2 saslprep semver nodemailer image-size wildleek@2.0.0 otplib@10.2.3 yubikeyotp; fi
`

but meshcentral.js specifically installs nodemailer 6.9.6
`if (nodemailer || ((config.smtp != null) && (config.smtp.name != 'console')) || (config.sendmail != null)) { modules.push('nodemailer@6.9.6'); } // Add SMTP support
`

**Expected behavior**
meshcentral and Dockerfile should be in aggreement on the module versions

**Additional context**
By forcing all modules to be reinstalled at startup, the NPM install command can timeout on slower machines.

gogglespisano:
ssh2 1.14.0 vs ssh2 1.15.0 which was released a few weeks ago is another module causing this problem.

gogglespisano:
For an inital fix I recommend the Dockerfile be updated with specific version numbers on all packages matching the version numbers in meshcentral.js and package.json.

There needs to be a better fix that doesn't require manually maintaining all the version numbers.

Issue #5681, opened 2024-01-10
InstallModules() installs all modules, not just missing modules (regression)


gogglespisano:
**Describe the bug**
In translate/translate.js, InstallModules() creates a list of missingModules, but then installs all modules.

In the previous version of this code, the modules in missingModules were installed one by one.

**To Reproduce**
When the docker image is run, additonal libs are installed which weren't in the package.json list.  This issue causes all modules to be installed, not just the missing modules

**Expected behavior**
Only the missing modules should be installed.

**Screenshots**
N/A

**Server Software (please complete the following information):**
Docker Image

**Client Device (please complete the following information):**
N/A

**Remote Device (please complete the following information):**
N/A

**Additional context**
Since all modules are reinstalled, it can take quite a while for MeshCentral to startup.  The npm command has a timeout of 120 seconds, which isn't always enough when all modules are being installed.

mcrec.js uses the previous style of this function.  It should probably be updated to match this newer code once this problem is fixed.

gogglespisano:
meshcentral.js has similar functions, but includes the following comment.

```
// Install all missing modules at once. We will be running "npm install" once, with a full list of all modules we need, no matter if they area already installed or not,
// this is to make sure NPM gives us exactly what we need. Also, we install the meshcentral with current version, so that NPM does not update it - which it will do if obmitted.
```

If all modules need to be installed, then the timeout should be increased to 300000.

There's several inconsitencies between the 3 versions of this code.  Could they be consoidated so only one copy exists?

si458:
this was the original behaviour, it would instalkl missing packages like mysql, postgres, totp, etc...
but it was changed because @Ylianst found a bug, but its still not perfect!
so im trying to find a way around this, or just completely remove it all together, and only install the extra packages that you request, like mysql, totop, pushover, etc...

gogglespisano:
Bug #5684 is the cause of all modules being reinstalled.

Issue #5545, opened 2023-11-13
Docker on Debian 11 fails on version 1.1.15 and 1.1.16 giving NPM errors


gogglespisano:
Mesh Central on Debian Docker fails with NPM errors
This also happened on version 1.1.11 (I think) but started working again a version later.  Now the problem is back.
The only setting that I've been changing is the docker image tag.
Debian (Docker Host) packages all up to date.

Using version 1.1.15 and 1.1.16 give this error
```
meshcentral    | Installing modules...
meshcentral    | ERROR: Unable to install required modules. MeshCentral may not have access to npm, or npm may not have suffisent rights to load the new module. To manualy install this module try:
meshcentral    |
meshcentral    |    cd "/opt/meshcentral/meshcentral"
meshcentral    |    npm install --no-audit --no-package-lock --no-optional --omit=optional --no-save meshcentral@1.1.16 archiver@5.3.2 body-parser@1.20.2 cbor@5.2.0 compression@1.7.4 cookie-session@2.0.0 express@4.18.2 express-handlebars@5.3.5 express-ws@4.0.0 ipcheck@0.1.0 minimist@1.2.8 multiparty@4.2.3 @yetzt/nedb node-forge@1.3.1 ua-parser-js@1.0.36 ws@8.14.2 yauzl@2.10.0 ssh2@1.14.0 semver@7.5.4 nodemailer@6.9.6 otplib@10.2.3
meshcentral    |    node node_modules/meshcentral
meshcentral    | Installing modules...
```

Using version 1.1.14 (since 15 and 16 don’t run) to run the suggested commands in bash inside container
```
1d8de0f7d6a1:/opt/meshcentral$ cd "/opt/meshcentral/meshcentral"
1d8de0f7d6a1:/opt/meshcentral/meshcentral$  npm install --no-audit --no-package-lock --no-optional --omit=optional --no-save meshcentral@1.1.16 archiver@5.3.2 body-parser@1.20.2 cbor@5.2.0 compression@1.7.4 cookie-session@2.0.0 express@4.18.2 express-handlebars@5.3.5 express-ws@4.0.0 ipcheck@0.1.0 minimist@1.2.8 multiparty@4.2.3 @yetzt/nedb node-forge@1.3.1 ua-parser-js@1.0.36 ws@8.14.2 yauzl@2.10.0 ssh2@1.14.0 semver@7.5.4 nodemailer@6.9.6 otplib@10.2.3
[...]

PR #5685, opened 2024-01-11
Interim fix for NPM module version mismatches

gogglespisano:
Align and specify NPM module versions

The Dockerfile specifies NPM modules to be installed.  However, some do not specify a version, so the latest is installed.  Later in meshcentral.js mainStart() specific versions are required.  If they don't match the latest version, all modules will be reinstalled to get the specific versions.

#5684 Soft version conflict on NPM modules causes NPM modules to be installed on startup in Docker
#5545 Docker on Debian 11 fails on version 1.1.15 and 1.1.16 giving NPM errors
#5681 InstallModules() installs all modules, not just missing modules (regression)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -86,8 +86,9 @@ COPY ./docker/config.json.template /opt/meshcentral/config.json.template
 # install dependencies from package.json and nedb
 RUN cd meshcentral && npm install && npm install nedb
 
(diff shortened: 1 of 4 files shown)


si458:
Thank u for this, I will look into it and check it's OK, then merge if so!
I'm waiting to hear bk from @Ylianst as to whether we need to actually install meshcentral in meshcentral to help cut down the startup install time for docker containers, for example

PR #5686, opened 2024-01-12
fix mobile terminal not showing enter key for Android soft keyboard

wdlut:
The terminal mode was not usable on many Android phones because the soft keyboard had no enter key. IPhones seem to have the default imput mode "text". For Android I added the attribute imputmode="text" and got also the enter key on the soft keyboard.


si458:
Thanks! Every little bug fix helps!

PR #5720, opened 2024-01-21
Fixed 'deprecated' warnings in gh-workflows

sschoen:


si458:
I have been meaning to update the github actions, thanks, but are you sure the release is going to work? As u have removed the my_token environment variable?

sschoen:
github-action-npm-release uses the GITHUB_TOKEN by default:
https://github.com/justincy/github-action-npm-release/blob/e237b4afa029f7982eaa6fafeace8343e902de4e/action.yml#L13

si458:
@sschoen no worries! My next question was actually is the github token generated automatically or does it need to be created manually, but no worries if you have put it bk to using the my_token, it was just setup this way a few years ago when I tested it, I never realised we could use github_token until afterwards doh!

sschoen:
@si458 the GITHUB_TOKEN is generated automatically, it is used for checkout etc., but for some actions you have to adjust the permissions
(https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)

Issue #5108, opened 2023-04-27
Raspberry Pi 4, 64 Bit kernel and 32 Bit Userland


diehummel:
**Describe the bug**
Not a bug itself with Meshcentral/Meshagent, but will hit many raspberry pi users.
On Raspberry Pi 4 with an update the kernel changed to aarch64.
The meshagent then gets killed with a "BUS" error when it starts. After around 8 retries when it starts, the console is not really connected.
No console commads work. Not Desktop and Terminal Tab is available.

Programs using uname -m to detect the architecture get a problem. 
The download script is loading the 64 bit binary, which will not work.
Source:
https://github.com/raspberrypi/firmware/issues/1795

Using arm_64bit=0 in /boot/config.txt will fix it, when using the 32bit OS.

 - OS: Raspbian (Bullseye) 32 bit

dinger1986:
Can you submit a PR to fix?

diehummel:
Not sure if this can be fixed from Meshcentral.
Because even if the install script will detect aarch64 and use getconf LONG_BIT to check if it is really 32 or 64 bit, the 32 bit agent will not work on the aarch64 kernel.

Not sure if the agent can be compiled as 32 bit able to run correct on aarch64.

EDIflyer:
Oooh thank you @diehummel - this explains why I'm seeing problems on my Pi 4 that's running bullseye and is now aarch64 but not on my Pi 400 that's still on buster armv7l!  It had been working fine - I hadn't realised there'd been a kernel update that changed things.  I was finding I could still run console/terminal/files but desktop wouldn't work. I tried uninstalling/reinstalling but it just made things worse with the agent recurrently restarting itself and showing the bus error if run manually! (https://github.com/Ylianst/MeshAgent/issues/183)

Did you have any luck finding a fix?

EDIflyer:
Ah - don't worry, just saw your post in the Raspberry Pi repo and confirm adding the `arm_64bit=0` flag to `/boot/config.txt` and rebooting did the trick for me too. Thanks for hunting that workaround down!

pelwell:
Hi @diehummel, I'm one of the Raspberry Pi maintainers.

> the 32 bit agent will not work on the aarch64 kernel

Are you certain about this? Very few apps care that the kernel is 64-bit. If you have examples of incompatibilities I'd be interested in investigating why.

diehummel:
@pelwell 
[...]

PR #5730, opened 2024-01-24
Detect 32bit Raspberry Pi userspace on 64bit kernel

gutschke:
We already do the same thing for Intel, but we should also handle ARM, as this is a situation that users can find themselves in after updating their distribution.

### DIFF
--- a/agents/meshinstall-linux.sh
+++ b/agents/meshinstall-linux.sh
@@ -111,8 +111,15 @@ CheckInstallAgent() {
         fi
         if [ $machinetype == 'aarch64' ]
         then
-          # RaspberryPi 3B+ running Ubuntu 64 (aarch64)
-          machineid=26
+          bitlen=$( getconf LONG_BIT )
+          if [ $bitlen == '32' ]
+          then
+            # 32bit user space on a 64bit kernel
+            machineid=25
+          else
+            # RaspberryPi 3B+ running Ubuntu 64 (aarch64)
+            machineid=26
+          fi
         fi
         # Add more machine types, detect KVM support... here.
       fi


si458:
this is only a work around patch for the install script
a real fix is already in the works with the agent so we wont merge this at the moment...

gutschke:
That's fair enough. 

Just clarifying though, after the fix that you are implementing, will the user end up with a 32bit binary even if running a 64bit kernel? It is quite conceivable that they wouldn't have the ability to run any 64bit code, even if the kernel theoretically could do so. So, it is important to look at more than just what "uname" returns.

This all gets a little confusing if you have a Debian multiarch system, where you could have both a 32bit and a 64bit runtime at the same time. But even then, it should be safe to pick the default architecture as detected by the "getconf" application.

si458:
no once new agents are released new agents will get 64bit meshagents,
the issue was originally because the 64bit meshagent wouldnt run on arm properly (sig error),
the end user would need to switch to 32bit kernel on 64bit os (weird i no), then the meshinstall script would auto detect the 32bit kernel and use the 32bit meshagent instead

i will double check all this tomorrow tho to be 110% sure im not chatting shit and getting myself mixed up

gutschke:
That's not going work then, I think.

There are several different scenarios that you need to be able to cover:

1. 32bit kernel booting a 32bit Ubuntu or Debian distribution. In this case, MeshCentral must vend a 32bit binary. There is no way for a 64bit binary to get loaded, as the required libraries including but not limited to the dynamic link loader are missing. And even if the user installed these libraries, the kernel wouldn't be able to run a 64bit binary; heck, the hardware might not even be capable of it. This happens on devices such as the Raspberry Pi Zero which lack a 64bit-capable CPU.
2. 64bit kernel booting a 64bit Ubuntu or Debian distribution. This is now a very common scenario and should result in a 64bit agent to be installed. While the hardware and the kernel could probably execute 32bit code, the same argument holds from option number 1 and the required 32bit libraries are not going to be there. So, only 64bit code will be able to execute.
3. Debian-derived distributions have support for "multiarch". This means that if the hardware has the necessary support, it is possible to install both 32bit and 64bit libraries at the same time. This would be unusual on an embedded device but is possible. Even then, there should be a "default" architecture and additional "foreign" architectures. It is probably safe to pick the default architecture when deciding which agent to install. That might be 32bit or 64bit, depending on how the user configured their system.
4. There is an interesting corner case that happens when upgrading from an older 32bit distribution. I think, at some point Debian or Ubuntu decided that they wanted users to switch to 64bit kernels on Raspberry Pi, if the hardware supports it and independent of whether they have a 32bit or 64bit distribution. This means, you'll detect a kernel that is 64bit capable, but only 32bit libraries are present. No 64bit code would ever be able to run. This is the scenario my patch fixes. If this case is detected, the 32bit agent must be installed. There already is code that handles the exact same scenario for Intel CPUs. I just added a branch for ARM.

I am a little confused by you suggesting that users should install a 32bit kernel on a 64bit device as a temporary work-around. That's not going to boot. I am sure I must be misunderstanding what you are describing. Maybe, you are thinking of something like option 3?

si458:
plz read #5108 and see if that helps for the time being...

gutschke:
That comment is incorrect, although it is easy to see how people get confused with all these different configuration options.

The 32bit binary **does work** on a 64bit kernel. In fact, that's exactly how I am running things right now and how Debian/Ubuntu expects everyone to run 32bit code on Raspberry Pi these days. That's why they ship a mixed-mode 32/64bit kernel.

The work-around in that discussion puts the kernel into pure 32bit mode instead of mixed 32/64bit. In that scenario, everything looks like a plain old 32bit system with a 32bit kernel, and the MeshAgent install script will do the right thing and install a 32bit binary. This kernel command line option also apparently switches the CPU into a mode that relaxes the alignment requirements and therefore doesn't mind that the 32bit MeshAgent performs unaligned memory accesses.

But even after fixing the crashing binary, it won't be possible to remove the command line option, boot the stock a 32/64bit kernel and run a 64bit agent binary; you will be able to run the fixed 32bit binary on the mixed 32/64bit kernel though.

On Linux, you always need three things to match. You need a 64bit kernel, a 64bit dynamic link loader, and 64bit system libraries. Or alternatively, you need all of that for 32bit. You can also install multiarch 32/64bit versions for all of the above.

Current users of Raspberry Pi devices will often find themselves in a situation where they have a 32/64bit kernel, a 32bit link loader, and 32bit system libraries. They started experience crashes with this new kernel, because they had previously installed a 32bit MeshAgent, but the newly installed 32/64bit kernel enforces stricter alignment requirements even when running 32bit code. So, the MeshAgent used to work fine, and now it stopped working thanks to the kernel that gained the ability to execute both 32bit and 64bit code. But that doesn't mean you could switch MeshAgent to 64bit on these systems. 

Of coourse, old installs of MeshAgent will start running again, as they remember the choice of architecture. But fresh installs would work. The auto-detected 64bit binary wouldn't be able to load with the 32bit userspace. And of course, you can't run 32bit either until you fix the alignment bugs (but as said earlier, we know how to do that and have the necessary commits).

In other words, you still need the changes to the installation script. You probably always needed them, but until Debian/Ubuntu decided to enable a 32/64bit kernel by default, very few people would encounter the bug in the install script.

PR #5754, opened 2024-01-31
Add: Set node icon set based on OS

singhragvendra503:
Please review docker image `https://hub.docker.com/r/singhragvendra503/meshcentral`
Screenshot from 2024-01-27 22-44-05
Screenshot from 2024-01-27 22-43-54


si458:
I've set this as a draft for the moment, because I would like to use proper images/logos instead

singhragvendra503:
Ok cool

Ylianst:
Arg. This is not a good PR in my view. The way it's done, if you click on the device and select a different icon, it will stick with the OS icon no matter what you select. MeshCentral also selects a laptop icon by default if the device has a battery and this will override it. The new icons are not great, the white background will not work in night mode.

This said, I would be ok placing a small overlay icon like the battery indicator for the OS.

Ylianst:
One thing that would be nice to some point would be to allow fully custom icons for devices. That way, you can upload and use what you want.

si458:
@Ylianst remember you can already partly do this `meshcentral-web/public/images` ?
but it only simply replaces ur icons not add new ones

PR #5791, opened 2024-02-11
Docs update on skipmaccheck

silversword411:
Where are all the other .msh options we can add to docs?


si458:
https://github.com/Ylianst/MeshAgent?tab=readme-ov-file#msh-format

silversword411:
Ok...that's too easy. Think we should just bulk copy/paste that readme into the docs for agent?

si458:
You can also add to the msh file with meshcentral on install too
Screenshot_20240211_134458_Firefox

silversword411:
> meshcentral on install too

Nifty trick. Definitely need to document that :)

si458:
I would maybe add only add the skipmaccheck, as all the others are really just for debug/issues

silversword411:
Stick with that for now. Always add more in the future :)

Issue #4886, opened 2022-12-28
Webrelay Custom Port Links


12nick12:
**Describe the solution you'd like**
Allow custom webrelay ports 

**Describe alternatives you've considered**
Right now you can configure custom meshrouter ports in the config.
`      "deviceMeshRouterLinks": {
        "rdp": true,
        "ssh": true,
        "scp": true,
        "extralinks": [
          {
            "name": "HTTP MR",
            "protocol": "http",
            "port": 80,
          },`

**Additional context**
Add a config change like `deviceWebRelayRouterLinks` that would put custom links on the device

Issue #4452, opened 2022-08-25
allow to add multiple HTTP/HTTPS services to a device for web-relay


herwig-:
**Is your feature request related to a problem? Please describe.**
At the moment, it is only possible to modify the ports of one HTTP and one HTTPS relay per device. Most devices will not have exactly one HTTP and one HTTPS connection. Some devices will only have either HTTP or HTTPS. Other might have mutlipte HTTP or HTTPS connections with different services behind them. (e.g. a ubuntu machine with multiple docker instances - each one publishing their web-service on a different port)

**Describe the solution you'd like**
Instead of HTTP and HTTPS, I would like to have a "Button" where I could add a web-relay-service. In the config menu for this item, I would like to be able to
* define the "Service name" -> which should be displayed instead of the current HTTP/HTTPS
* define the port to be used
* define wether the connection should be HTTP or HTTPS

example:
instead of this:
![grafik](https://user-images.githubusercontent.com/741471/186780398-1e0cdab1-7f3f-4ff5-85de-6d39ae023a66.png)

I would prefere a version where it says
`"HomeAssistant (http, 8123)"`

alex-eri:
Also remote devices needed. For connecting to routers web interface.

herwig-:
remote devices only needed if it is a device with an agent.

In this case you could have 2 possibilities to access web-relay-services:
1. via the device with an agent installed
2. via the group which is already assigned to an agent that is responsible for forwarding ...

In the first case, if the device with the agent goes down, the services can no longer be reached.
In the second case, the group can be modified to use a different agent on the same remote network to still access the services

PR #5801, opened 2024-02-14
WebRelay support for devicemeshrouterlinks.extralinks

SecareLupus:
This pull request may affect Issues #4452 and #4886

Company I work for wanted to be able to add custom WebRelay Links for accessing (possibly multiple concurrent) web services on fleet devices. I do not have a good way to test these changes, as I'm not sure the best way to spin up a development instance. Advice for doing so is welcome, and I'd be happy to test these changes myself. Until then, I believe the changes to be likely to work as intended.

Changes I made (in order from least to most important):
* Moved two comments so they were attached to the top of the code they were referencing. (It appears additional code was inserted between the comment and the commented code at some point)
* Updated the config.json schema document to clarify the intended use of my changes to configuration.
  * Added two additional legal protocol values, 'wr_http' & 'wr_https'. These values use the same 'p' integer as 'http' and 'https', respectively, for backwards compatibility and to limit the changes needed.
  * Added an optional property, 'path', which is intended to provide endpoint information to the remote web service. Assumed to be compatible with the preconditions needed for the 'addr' argument for the function 'p10WebRouter()' Those preconditions do not appear to be documented in 'default.handlebars'.
* Modified two sections of code which produce node links based on 'extralinks' configuration values.
  * If the value of r.protocol is set to one of the new values, and WebRouterPort != 0 it calls p10WebRouter(), passing the relevant information as provided by the configuration values.
  * Else, it falls back on current behavior.

The code is not expected to produce any differences from current behavior unless an extralink entry is provided which uses the 'wr_http' or 'wr_https' protocol.

Example config.js based on our usage:
```javascript
"deviceMeshRouterLinks": {
        "rdp": true,
        "ssh": true,
        "scp": true,
        "extralinks": [
          {
            "name": "NodeRed Flows",
            "protocol": "wr_http",
            "port": 1880,
            "filter": "tag:has-nodered"
          },
          {
            "name": "NodeRed Dashboard",
            "protocol": "wr_http",
            "path": "/ui"
            "port": 1880,
            "filter": "tag:has-nodered"
          },
          {
            "name": "Simple Third Example",
            "protocol": "wr_http",
            "port": 80
          }
       ]
}
```

### DIFF
--- a/meshcentral-config-schema.json
+++ b/meshcentral-config-schema.json
@@ -1360,7 +1360,12 @@
                       "type": "string"
                     },
                     "protocol": {
-                      "description": "Protocol. Valid values are: custom,http,https,rdp,ssh,scp,mcrdesktop,mcrfiles.",
+                      "description": "Protocol. Valid values are: custom,http,wr_http,https,wr_https,rdp,ssh,scp,mcrdesktop,mcrfiles.",
+                      "type": "string"
+                    },
+                    "path": {
+                      "description": "Optional destination path for web relay links to the remote device.",
+                      "default": "",
                       "type": "string"
                     },
                     "port": {
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -4795,12 +4795,12 @@
         function getShortRouterLinks(node) {
             var x = '', meshrights = GetNodeRights(node);
 
-            // RDP link, show this link only of the remote machine is Windows.
             if ((((node.conn & 1) != 0) || (node.mtype == 3)) && (node.agent) && ((meshrights & 8) != 0) && (node.agent.id != 14)) {
                 if (webRelayPort != 0) {
                     x += '<a href=# onclick=p10WebRouter("' + node._id + '",1,' + (node.httpport ? node.httpport : 80) + ')>' + "HTTP" + ((node.httpport && (node.httpport != 80)) ? '/' + node.httpport : '') + '</a>&nbsp;';
                     x += '<a href=# onclick=p10WebRouter("' + node._id + '",2,' + (node.httpsport ? node.httpsport : 443) + ')>' + "HTTPS" + ((node.httspport && (node.httpsport != 443)) ? '/' + node.httpsport : '') + '</a>&nbsp;';
                 }
+                // RDP link, show this link only if the remote machine is Windows.
                 if ((node.agent.id > 0) && (node.agent.id < 5)) {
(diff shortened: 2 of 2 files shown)


SecareLupus:
I have some questions about the duplication of code related to this PR. I did my best to make no dramatic changes to structure, but that meant editing two seemingly similar chunks of code. They were not identical, so I was not willing to blindly merge them.

It appears that the function getShortRouterLinks() at line 4795 performs roughly the same task as the block of code starting at 7613, which appears to be a part of the function goToDevice()

They even shared the same misplaced comment, indicating a probable copy-paste at some point. I'm not sure I see why the code at 7613 can't just call getShortRouterLinks(), possibly something to do with the way that node permissions are determined? In any case, I opted not to merge them, and instead applied the changes to each of them separately.

One line that was particularly different between them seems to limit the visibility of Desktop and File Manager links to devices which have a proper agent.
Line 4828 (getShortRouterLinks):
`if ((node.mtype == 3) && ((r.protocol == 'mcrdesktop') || (r.protocol == 'mcrfiles'))) continue;`
Line 7644 (gotoDevice):
`if (((p == 6) || (p == 7)) && (node.mtype != 2)) continue;`

Their placement is also different, as one happens before an integer value is set for 'p' and one happens after. This explains most differences, but I'm curious about the difference between `node.mtype == 3` to `node.mtype != 2`. Is this difference intentional, or are they intended to do the same thing and just do it different ways? Should one be changed?

SecareLupus:
If it wasn't clear, I'd love eyes and advice from someone more familiar with the project on what next steps should look like. Particularly whether my approach makes sense, whether I'm missing anything obvious, and suggestions for how to test code changes.

Ylianst:
Apologies for the delay. Yes, I will take a look at this hopefully in the next few days.

SecareLupus:
I did some testing in a javascript console; I think I've misunderstood the use of the 'addr' argument in the p10WebRouter function, and don't expect this to work in its current form. Not sure yet how to fix it, but ready to put in more work on it if pointed in the right direction. Thank you for your time!

SecareLupus:
I'm going to have to take some time to figure out how to get a test instance setup so that I can do this properly. I'm going to close this PR for now, since it isn't/wasn't ready, and I'll try to get my dev stack in order and resubmit the PR when it's actually ready and tested. No reason to keep a stale PR open in the meantime.

PR #5856, opened 2024-02-25
Migrate to openid client

### LINKED ISSUE #4531: Bug: Azure SSO uses 'unique_name' for 'sid', should be 'oid' or 'sub'
The Azure SSO section appends the `unique_name` attribute to the end of `~azure:` to give you your user ID. 

Azure [User-Object](https://github.com/Ylianst/MeshCentral/blob/c86466210b69c52dcb9fa0f4ac24323818a6ce4a/webserver.js#L6999) Creation:
```javascript
var user = { sid: '~azure:' + userex.unique_name, name: userex.name, strategy: 'azure' };
```

However it seems there are issues using `unique_name` claim according to the documentation:
> Only present in v1.0 tokens. Provides a human readable value that  identifies the subject of the token.
> This value is not guaranteed to be unique within a tenant and should be used only for display purposes.

The `sub` claim seems to be the most appropriate:
> The principal about which the token asserts information, such as the user of an app. This value is immutable and cannot be reassigned or reused. The subject is a pairwise identifier - it is unique to a particular application ID. If a single user signs into two different apps using two different client IDs, those apps will receive two different values for the subject claim. This may or may not be wanted depending on your architecture and privacy requirements.

However `oid` could also be used:
> The immutable identifier for an object in the Microsoft identity system, in this case, a user account. This ID uniquely identifies the user across applications - two different applications signing in the same user will receive the same value in the `oid` claim. The Microsoft Graph will return this ID as the `id` property for a given user account. Because the `oid` allows multiple apps to correlate users, the `profile` scope is required to receive this claim. Note that if a single user exists in multiple tenants, the user will contain a different object ID in each tenant - they're considered different accounts, even though the user logs into each account with the same credentials. The `oid` claim is a GUID and cannot be reused.

So maybe this could be updated, although we would have to still pull in `unique_name` to provide backwards compatibility or check the db for that style of sid and run a conversion to the new way.

The `Only present in v1.0 tokens` line in the documentation for the `unique_name` claim is what concerns me the most, as I think that means if we were to move to OIDC or SAML authentication for Azure over the current OAUTH2, we would not be able to see that claim anymore, potentially further complicating the changeover.

[Related Microsoft Azure AD Documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens#payload-claims)

### LINKED ISSUE #4545: Unable to use Gitlab OIDC provider - OpenIDConnectStrategy requires an authorizationURL option
**Describe the bug**
I'm trying to use a self hosted Gitlab instance as OIDC auth provider and I'm facing the following error:

```
MeshCentral HTTP redirection server running on port 81.
MeshCentral v1.0.85, LAN mode, Production mode.
MeshCentral HTTP server running on port 80, alias port 443.
ERR: /opt/meshcentral/meshcentral/node_modules/@mstrhakr/passport-openidconnect/lib/strategy.js:32
  if (!options.authorizationURL) { throw new TypeError('OpenIDConnectStrategy requires an authorizationURL option'); }
                                         ^

TypeError: OpenIDConnectStrategy requires an authorizationURL option
    at new Strategy (/opt/meshcentral/meshcentral/node_modules/@mstrhakr/passport-openidconnect/lib/strategy.js:32:42)
    at /opt/meshcentral/meshcentral/webserver.js:7036:51
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
Error: Command failed: /usr/bin/node /opt/meshcentral/meshcentral/meshcentral --configfile config.json --launch 7
/opt/meshcentral/meshcentral/node_modules/@mstrhakr/passport-openidconnect/lib/strategy.js:32
  if (!options.authorizationURL) { throw new TypeError('OpenIDConnectStrategy requires an authorizationURL option'); }
                                         ^

TypeError: OpenIDConnectStrategy requires an authorizationURL option
    at new Strategy (/opt/meshcentral/meshcentral/node_modules/@mstrhakr/passport-openidconnect/lib/strategy.js:32:42)
    at /opt/meshcentral/meshcentral/webserver.js:7036:51
    at processTicksAndRejections (node:internal/process/task_queues:96:5)

    at ChildProcess.exithandler (node:child_process:398:12)
    at ChildProcess.emit (node:events:539:35)
    at maybeClose (node:internal/child_process:1092:16)
    at Socket.<anonymous> (node:internal/child_process:451:11)
    at Socket.emit (node:events:527:28)
    at Pipe.<anonymous> (node:net:709:12) {
  code: 1,
  killed: false,
  signal: null,
  cmd: '/usr/bin/node /opt/meshcentral/meshcentral/meshcentral --configfile config.json --launch 7'
  ```

**Expected behavior**
I want to authenticate using OIDC with my Gitlab instance as provider.

**Screenshots**

**Server Software (please complete the following information):**
 - OS: Debian 11
 - Virtualization: Docker 20.10.17
  - Network: Traefik as reverse proxy
 - Version: v1.0.85
 - Node: v16.16.0 provided in Docker image

**Client Device (please complete the following information):**
 - Device: Laptop
 - OS: Arch
 - Network: Local to Meshcentral
 - Browser: Firefox
 - MeshCentralRouter Version: v1.0.85

**Additional context**

**Your config.json file**
```
{
  "$schema": "http://info.meshcentral.com/downloads/meshcentral-config-schema.json",
  "settings": {
    "plugins":{"enabled": false},
    "mongoDb": "mongodb://mongodbadmin:mysecretpassword@mongodb:27017",
    "WANonly": false,
    "LANonly": true,
    "sessionKey": "mysecretsessionkey",
    "port": 80,
    "redirPort": 81,
    "aliasPort": 443,
    "AgentPong": 300,
    "TLSOffload": true,
    "SelfUpdate": false,
    "AllowFraming": false,
    "WebRTC": false
  },
  "domains": {
    "": {
      "title": "MeshCentral",
      "title2": ".mysecretdomain.org",
      "minify": true,
      "NewAccounts": true,
      "orphanAgentUser": "removed@mysecretdomain.org",
      "localSessionRecording": false,
      "userNameIsEmail": true,
      "authStrategies": {
        "oidc": {
          "authorizationURL": "https://gitlab.mysecretdomain.org/oauth/authorize",
          "callbackURL": "https://meshcentral.mysecretdomain.org/oidc-callback",
          "clientid": "myclientid",
          "clientsecret": "mysecretclientsecret",
          "issuer": "https://gitlab.mysecretdomain.org:443",
          "tokenURL": "https://gitlab.mysecretdomain.org/oauth/token",
          "userInfoURL": "https://gitlab.mysecretdomain.org/oauth/userinfo",
          "logoutURL": "https://gitlab.mysecretdomain.org/oauth/revoke",
          "newAccounts": true
        }
      }
    }
  }
}
```

Edit:
BTW, here is the openid configuration provided by gitlab:

```
{
  "issuer": "https://gitlab.com",
  "authorization_endpoint": "https://gitlab.com/oauth/authorize",
  "token_endpoint": "https://gitlab.com/oauth/token",
  "revocation_endpoint": "https://gitlab.com/oauth/revoke",
  "introspection_endpoint": "https://gitlab.com/oauth/introspect",
  "userinfo_endpoint": "https://gitlab.com/oauth/userinfo",
  "jwks_uri": "https://gitlab.com/oauth/discovery/keys",
  "scopes_supported": [
    "api",
    "read_api",
    "read_user",
    "read_repository",
    "write_repository",
    "read_registry",
    "write_registry",
    "sudo",
    "openid",
    "profile",
    "email"
  ],
  "response_types_supported": [
    "code"
  ],
  "response_modes_supported": [
    "query",
    "fragment"
  ],
  "grant_types_supported": [
    "authorization_code",
    "password",
    "client_credentials",
    "refresh_token"
  ],
  "token_endpoint_auth_methods_supported": [
    "client_secret_basic",
    "client_secret_post"
  ],
  "subject_types_supported": [
    "public"
  ],
  "id_token_signing_alg_values_supported": [
    "RS256"
  ],
  "claim_types_supported": [
    "normal"
  ],
  "claims_supported": [
    "iss",
    "sub",
    "aud",
    "exp",
    "iat",
    "sub_legacy",
    "name",
    "nickname",
    "email",
    "email_verified",
    "website",
    "profile",
    "picture",
    "groups",
    "groups_direct",
    "https://gitlab.org/claims/groups/owner",
    "https://gitlab.org/claims/groups/maintainer",
    "https://gitlab.org/claims/groups/developer"
  ]
}
```

### LINKED ISSUE #5179: Add support for custom mappings between OIDC token and meshcentral user

### LINKED ISSUE #5725: Hard Coded OIDC scopes limit IDP control of scopes
The current implementation of oidc auth has hard coded client scopes. my requests to the IDP (keycloak, in my case) has the following scopes in the URL `scope=openid%20profile%20email%20groups` apart from `oidc` the other scope names should be customizable.

Why? as this enables the IDP to be setup to suit so as to limit what information is passed to a client, regardless of what's requested. This is a must considering privacy and in some areas mandated by legislation as it is in my country when dealing with personal information. Although there is an argument to be made that a group is not personal information, however that depends on which side of the fence you site on.

use case as an example:
- users/groups stored within ldap.
- IDP is synced with ldap. users and groups.
- any requirement for `profile` scoped information which deviates from a standard `profile` scope gets it's own scope created specifically to suit the application using the info and only contains what is required for the application, this includes additional required information for the token requested
- any requirement for a group `scope` will not use the standard `group` scope. all groups the user is part of are not required for authorization. a custom `group` scope is created with the client roles used as the "groups" for the application. This aids significantly as token can easily be larger that is allowed for a url and/or cookie. this is dependent upon the application however.

to be able to specify the oidc scope names. no hard coded scope names outside of `oidc`. What information is required from the IDP be documented. This will aid in setting up scopes in accordance with what ever policy the IDP uses and provide what is required to be within the token from the IDP. The reality is, as long as what is required for meshcentral from the IDP is known, a single scope could be used, splitting the scope up is not required.

There is no alternative. There is an exception, using the hard coded `group` scope which generally will contain all of the groups the user is part of. However this is not a solution as the token is too large and passing personal information not required by an application is not within the confines of limiting access to personal information as is a requirement from my end. In the same light, if another application is using that "group scope" one of the two applications will receive information not required for it to function.  I'm aware of group filtering, however the client is still passed information not required by it, so is also not an option.

**Additional context**

On a side note, credit to the developer(s) of meshcentral. It's not often that I test a product for implementation and when doing so come across an issue that needs to be resolved and that there already be an existing feature to solve that problem. As it stands now, this is the only issue that I've found. I do hope that you are proud of what you have achieved because if I was just to say I was impressed, would be an understatement. great work.

### PR DESCRIPTION
**Pull request by mstrhakr**
This is merge ready. I updated this to 1.1.21 then I tested it with Authelia (custom setup and with old configs) as well as Azure using the preset. This doesn't interfere with any old configuration and adds a bunch of cool features. I have again published the latest version of this on docker hub (`mstrhakr/MeshCentral:dev`) and tested that as well.

@Ylianst if you have any questions about this, please let me know. Although I wrote some documentation that you can reference as well.

Fixes #4545 
Fixes #5179 
Fixes #4531 
Fixes #5725

### CODE REVIEW

**Comment by si458** [comment]
Have you verified this works with 1.0.21?
Also what about existing people who use oidc?
I see you have removed alot of options and added new ones in the schema file which will effect the config.json?

**Comment by mstrhakr** [comment]
> Have you verified this works with 1.0.21?

Yes, this was updated to 1.0.21 and then tested as described above.

> Also what about existing people who use oidc? 

This should be a drop in replacement for existing users as described in the documentation.

> I see you have removed alot of options and added new ones in the schema file which will effect the config.json?

Technically I didn't remove any items, they were just re-arranged and modified. Existing configs will work and are handled correctly with no user intervention required.

**Comment by exander77** [comment]
Can this be merged? I want to test this with Azure and if necessary built upon this.

**Comment by mstrhakr** [comment]
> Can this be merged? I want to test this with Azure and if necessary build upon this.

Feel free to test this from my repository directly (migrate_to_openid-client branch). There is also a docker available at mstrhakr/meshcentral:dev

**Comment by exander77** [comment]
@mstrhakr I am already building on top of that, I have Azure groups working by reading them from graph api.

Btw, there is some bad merge/rebase:
```node
+        const strategy = domain.authstrategies[req.user.strategy];
         const groups = { 'enabled': typeof strategy.groups == 'object' }
         if ((req.user != null) && (req.user.sid != null) && (req.user.strategy != null)) {
-            const strategy = domain.authstrategies[req.user.strategy];
```

Strategy is used before it is defined. I had to move that line.

exander77:
After this is oked and accepted, I will make pull with for Azure groups.

mstrhakr:
> @mstrhakr I am already building on top of that, I have Azure groups working by reading them from graph api.

I already had that as a feature of this build.

> Btw, there is some bad merge/rebase:

Thanks! I somehow missed that.

Correct me if I'm wrong but I think this would make more sense.

```js
-         const groups = { 'enabled': typeof strategy.groups == 'object' }
          if ((req.user != null) && (req.user.sid != null) && (req.user.strategy != null)) {
              const strategy = domain.authstrategies[req.user.strategy];
+             const groups = { 'enabled': typeof strategy.groups == 'object' }
```

exander77:
@mstrhakr Sure, that works as well. My issue is just that strategy is used before it is defined.

exander77:
Otherwise, everything worked for me. For groups in Azure, I basically just appropriately fill `user.groups`.

jon-nfc:
+1

@mstrhakr thanks for your time in doing this work. Eagerly awaiting so as to take advantage of being able to specify the groups claim name #5725 .

mstrhakr:
> I am testing it with my Azure changes, and so far everything seems ok.

What changes are you making? I only ask because groups are already supported with this branch.

exander77:
This is the Azure code:
```node
                    // Azure
                    if ((domain.authstrategies.authStrategyFlags & domainAuthStrategyConsts.azure) != 0) {
                        obj.app.get(url + 'auth-azure', function (req, res, next) {
                            var domain = getDomain(req);
                            if (domain.passport == null) { next(); return; }
                            domain.passport.authenticate('azure-' + domain.id, { state: obj.parent.encodeCookie({ 'p': 'azure' }, obj.parent.loginCookieEncryptionKey) })(req, res, next);
                        });
                        obj.app.get(url + 'auth-azure-callback', function (req, res, next) {
                            var domain = getDomain(req);
                            if (domain.passport == null) { next(); return; }
                            if ((Object.keys(req.session).length == 0) && (req.query.nmr == null)) {
                                // This is an empty session likely due to the 302 redirection, redirect again (this is a bit of a hack).
                                var url = req.url;
                                if (url.indexOf('?') >= 0) { url += '&nmr=1'; } else { url += '?nmr=1'; } // Add this to the URL to prevent redirect loop.
                                res.set('Content-Type', 'text/html');
                                res.end('<html><head><meta http-equiv="refresh" content=0;url="' + url + '"></head><body></body></html>');
                            } else {
                                if (req.query.state != null) {
                                    var c = obj.parent.decodeCookie(req.query.state, obj.parent.loginCookieEncryptionKey, 10); // 10 minute timeout
                                    if ((c != null) && (c.p == 'azure')) { domain.passport.authenticate('azure-' + domain.id, { failureRedirect: '/' })(req, res, next); return; }
                                }
                                next();
                            }
                        }, handleStrategyLogin);
                    }
```

As you can see, there is no groups handling, what is actually needed is:
Add Graph resource:
```node
-            let options = { clientID: domain.authstrategies.azure.clientid, clientSecret: domain.authstrategies.azure.clientsecret, tenant: domain.authstrategies.azure.tenantid };
+            let options = { clientID: domain.authstrategies.azure.clientid, clientSecret: domain.authstrategies.azure.clientsecret, tenant: domain.authstrategies.azure.tenantid, resource: 'https://graph.microsoft.com' };
```
Implement a function that accesses graph API using the acceesToken to get `https://graph.microsoft.com/v1.0/me/memberOf`:
```node
-                function (accessToken, refreshtoken, params, profile, done) {
+                async function (accessToken, refreshtoken, params, profile, done) {
+                    async function fetchUserGroups(accessToken) {
+                        const url = 'https://graph.microsoft.com/v1.0/me/memberOf';
+                        try {
+                            const response = await require('axios').get(url, {
+                                headers: {
+                                    'Authorization': `Bearer ${accessToken}`,
+                                     'Content-Type': 'application/json'
+                                }
+                            });
+                            const groups = response.data.value.map((g) => g.displayName);
+                            return groups;
+                        } catch (error) {
+                            parent.authLog('setupDomainAuthStrategy', 'Error: ' + error) 
+                           return [];
+                       }
+                   }
```

Call the function:
```node
-                    parent.authLog('setupDomainAuthStrategy', 'Azure profile: ' + JSON.stringify(userex));
+                    const groups = await fetchUserGroups(accessToken)
+                    parent.authLog('setupDomainAuthStrategy', 'Azure profile: ' + JSON.stringify(userex) + ' ' + JSON.stringify(groups));
```

Fill groups in user struct with the resutl:
```node
-                        var user = { sid: '~azure:' + userex.unique_name, name: userex.name, strategy: 'azure' };
+                        var user = { sid: '~azure:' + userex.unique_name, name: userex.name, strategy: 'azure', groups: groups };
```

mstrhakr:
Yea that's probably easier than how I accomplished it. Check out my getGroups function. We may be able to reuse it for the Azure strategy. 

However this version of the OIDC strategy includes an Azure preset that does already pull groups from the Graph API, although I manually made the request instead of using another module.  

It would not work for existing azure users though as it uses a different identifier for the users and there is currently no code to migrate the users to the new identifier.

exander77:
> Yea that's probably easier than how I accomplished it. Check out my getGroups function. We may be able to reuse it for the Azure strategy.
> 
> However this version of the OIDC strategy includes an Azure preset that does already pull groups from the Graph API, although I manually made the request instead of using another module.
> 
> It would not work for existing azure users though as it uses a different identifier for the users and there is currently no code to migrate the users to the new identifier.

I see, but does this work, is it enabled? I am using your PR, but doesn't fetch groups for me. I can debug to see what is wrong.

exander77:
Maybe I just don't know how to set it up, that code is never called if I use Azure Strategy:
```
      "authStrategies": {
        "azure": {
          "callbackurl": "...",
          "newAccounts": true,
          "clientid": "...",
          "clientsecret": "...",
          "tenantid": "...",
          "groups": {
            "required": [ "..." ],
            "siteadmin": [ "..." ],
            "sync": {
              "enable": true,
              "filter": [ "..." ]
            }
          }
        }
      }
```

exander77:
```
            } else if ((typeof strategy.issuer.issuer == 'string') && (typeof strategy.custom.preset == 'string')) {
                let error = new Error(`OIDC: PRESET: ${strategy.custom.preset.toUpperCase()}: PRESET OVERRIDDEN: CONFIG ISSUER: ${strategy.issuer.issuer} PRESET ISSUER: ${presetIssuer}`);
                parent.authLog('setupDomainAuthStrategy', error.message);
                console.warn(error)
            }
```
This code is erroneous, produces `presetIssuer` not defined.

mstrhakr:
Check out the documentation I made for the presets. I didn't modify the code for the Azure strategy, I added a preset configuration to the oidc strategy for azure that does what you want already. Its possible we could modify that getGroups function to work for the Azure strategy as well but currently it wouldn't be a drop in. 

Basically you'd need to use the oidc strategy and enable the azure preset within it to use groups as is.
I added an entire file of documentation for just the oidc strategy in this branch, but I'm happy to help however I can.

exander77:
> Check out the documentation I made for the presets. I didn't modify the code for the Azure strategy, I added a preset configuration to the oidc strategy for azure that does what you want already. Its possible we could modify that getGroups function to work for the Azure strategy as well but currently it wouldn't be a drop in.
> 
> Basically you'd need to use the oidc strategy and enable the azure preset within it to use groups as is. I added an entire file of documentation for just the oidc strategy in this branch, but I'm happy to help however I can.

Trying to get it to work.

This code:
```node
            // Discover additional information if available, use endpoints from config if present
            let issuer
            try {
                issuer = await strategy.obj.openidClient.Issuer.discover(strategy.issuer.issuer);
            } catch(err) {
                let error = new Error('OIDC: Discovery failed.', { cause: err });
                parent.authLog('setupDomainAuthStrategy', `ERROR: ${JSON.stringify(error)} ISSUER_URI: ${strategy.issuer.issuer}`);
                throw error
            } finally {
                if (Object.keys(strategy.issuer).length > 1) {
                    parent.authLog('setupDomainAuthStrategy', `OIDC: Adding Issuer Metadata: ${JSON.stringify(strategy.issuer)}`);
                    issuer = new strategy.obj.openidClient.Issuer(Object.assign(issuer.metadata, strategy.issuer));
                }
                strategy.issuer = issuer.metadata
                strategy.obj.issuer = issuer
            }
```

Can't work, finally accesses `issuer.metadata` which fails if an exception actually happened in try block.

exander77:
In `sample_config_advanced.json`. This is not consistend with docs:
```json
        "oidc": {
          "issuer": {
            "issuer": "https://sso.server.com",
            "end_session_endpoint": "https://sso.server.com/logout"
          },
          "client": {
            "clientid": "00000000-0000-0000-0000-000000000000",
            "clientsecret": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
          },
          "groups": {
            "required": [ "groupA", "groupB", "groupC" ],
            "siteadmin": [ "groupA" ],
            "sync": {
              "enable": true,
              "filter": [ "groupB", "groupC" ]
            }
          },
          "newAccounts": true
        }
```

exander77:
And this from docs:
```json
"oidc": {
    "client": {
        "client_id": "a1gkl04i-40g8-2h74-6v41-2jm2o2x0x27r",
        "client_secret": "AxT6U5K4QtcyS6gF48gndL7Ys22BL15BWJImuq1O"
    },
    "custom": {
        "preset": "azure",
        "tenant_id": "46a6022g-4h33-1451-h1rc-08102ga3b5e4"
    }
}
```
Will not work as it fails on missing `issuer.metadata` and fails on the error mentioned above:
```
meshcentral  | ERR: /opt/meshcentral/meshcentral/webserver.js:7374
meshcentral  |                 strategy.issuer = issuer.metadata
meshcentral  |                                          ^
meshcentral  | 
meshcentral  | TypeError: Cannot read properties of undefined (reading 'metadata')
```

exander77:
So, no it will actually work like this:
```json
        "oidc": {
          "client": {
            "client_id": "...",
            "client_secret": "..."
          },
          "groups": {
            "required": [ "...", "..." ],
            "siteadmin": [ "..." ],
            "sync": {
              "enable": true,
              "filter": [ "...", "..." ]
            }
          },
          "custom": {
            "preset": "azure",
            "tenant_id": "..."
          },
          "newAccounts": true,
          "callbackURL": "..."
        },
```

webserver.js
```diff
@@ -7261,41 +7259,306 @@ module.exports.CreateWebServer = function (parent, db, args, certificates, doneF
...
+            if (!obj.common.validateString(strategy.issuer.issuer)) {
+                if (preset == 'azure') { strategy.issuer.issuer = 'https://login.microsoftonline.com/' + strategy.custom.tenant_id + '/v2.0'; }
+                if (preset == 'google') { strategy.issuer.issuer = 'https://accounts.google.com'; }
+                if (!preset) {
+                    let error = new Error('OIDC: Missing issuer URI.');
+                    parent.authLog('error', `${error.message} STRATEGY: ${JSON.stringify(strategy)}`);
+                    throw error;
+                }
+            } else if ((typeof strategy.issuer.issuer == 'string') && (typeof strategy.custom.preset == 'string')) {
+                let error = new Error(`OIDC: PRESET: ${strategy.custom.preset.toUpperCase()}: PRESET OVERRIDDEN: CONFIG ISSUER: ${strategy.issuer.issuer} PRESET ISSUER: ${presetIssuer}`);
```
exander77, line 7319: `presetIssuer` is not defined.

webserver.js
```diff
@@ -7261,41 +7259,306 @@ module.exports.CreateWebServer = function (parent, db, args, certificates, doneF
...
+            } catch(err) {
+                let error = new Error('OIDC: Discovery failed.', { cause: err });
+                parent.authLog('setupDomainAuthStrategy', `ERROR: ${JSON.stringify(error)} ISSUER_URI: ${strategy.issuer.issuer}`);
+                throw error
+            } finally {
+                if (Object.keys(strategy.issuer).length > 1) {
+                    parent.authLog('setupDomainAuthStrategy', `OIDC: Adding Issuer Metadata: ${JSON.stringify(strategy.issuer)}`);
+                    issuer = new strategy.obj.openidClient.Issuer(Object.assign(issuer.metadata, strategy.issuer));
+                }
+                strategy.issuer = issuer.metadata
```
exander77, line 7356: This will fail as `issuer` is null if exception actually happened.

exander77:
I added notes in the PR code, see that. Testing it. Looks good.

So the actual idea is to phase auth old auth strategies and use oidc.

sample-config-advanced.json
```diff
@@ -519,23 +519,23 @@
...
-          "tokenURL": "https://sso.server.com/api/oidc/token",
-          "userInfoURL": "https://sso.server.com/api/oidc/userinfo",
-          "logoutURL": "https://sso.server.com/logout",
-          "newAccounts": true,
+          "issuer": {
+            "issuer": "https://sso.server.com",
+            "end_session_endpoint": "https://sso.server.com/logout"
+          },
+          "client": {
+            "clientid": "00000000-0000-0000-0000-000000000000",
```
exander77, line 527: Should be `client_id`.

sample-config-advanced.json
```diff
@@ -519,23 +519,23 @@
...
+            "clientsecret": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```
exander77, line 528: Should be `client_secret`.

sample-config-advanced.json
```diff
@@ -519,23 +519,23 @@
...
+          },
           "groups": {
             "required": [ "groupA", "groupB", "groupC" ],
             "siteadmin": [ "groupA" ],
             "sync": {
               "enable": true,
               "filter": [ "groupB", "groupC" ]
             }
-          }
+          },
```
exander77, line 537: There should be `custom` section with `preset` and `tenant_id`.

webserver.js
mstrhakr: Fixed this one

sample-config-advanced.json
mstrhakr: This is how the node-openid-client module had it setup so I used those configs. This still accepts client_id and client_secret at the top level but if you put them under client you need to remove the underscore to be compliant with the schema I've setup.

sample-config-advanced.json
mstrhakr: This is just an example of an advanced generic oidc configuration. If you want to delve into the presets and such there is documentation but I thought it would be prodent to keep that out of the example configs here. If you want to see all the options available you should be looking at the schema, not the sample configs.

sample-config-advanced.json
mstrhakr: I had it backward and you are correct on both of these. I'll post a fix now.

webserver.js
mstrhakr: If I remember correctly, getting issuer data here is needed and if this doesn't work the rest won't either. So if discovery fails and you get the error there is some misconfiguration, as in you manually entered an invalid issuer url while attempting to use a preset. I'd like to see your configs and logs for this if possible.

exander77: It is caused by bad config, but the issue is that it should not fail on accessing `metadata` on undefined in successfully caught exception. If you want it in simples way, I would do: `            let issuer = {}`

exander77:
Can you rebase it on top of current master? And can We merge it? It is becoming impractical for testing with features that are in master?

mstrhakr:
@exander77 I can try but I'm actually not a real programmer and have no idea what I'm doing. As far as merging I think we're waiting for a review from @Ylianst

Ylianst:
Ok. I am looking into this today. Lots of changes, but the PR quality looks very good. This may break existing installations, however, it's a good update. I need to read this line-by-line and will approve soon.

webserver.js
mstrhakr: I removed the finally block

exander77:
> Ok. I am looking into this today. Lots of changes, but the PR quality looks very good. This may break existing installations, however, it's a good update. I need to read this line-by-line and will approve soon if no issues.

Thank you very much, I will be testing it after it is merged and if anything out of ordinary is found, I will sure report it.

mstrhakr:
> So the actual idea is to phase auth old auth strategies and use oidc.

Yes basically, I'd like to do the same with saml via [node-saml](https://github.com/node-saml/node-saml) eventually. That should cover 99% of use cases between the two of them.

mstrhakr:
> Ok. I am looking into this today. Lots of changes, but the PR quality looks very good. This may break existing installations, however, it's a good update. I need to read this line-by-line and will approve soon if no issues.

This has basically been rewritten since you saw it the first time. Almost all of the code for the other strategies is essentially untouched to hopefully ensure no issues with the other strategies. I had to make web server startup asynchronously to get the discovery to work correctly but I think it is worth it for be able to get groups from the API's and such. I've said it before but I'm not a real programmer, I'm just learning and working on things that I find interesting so hopefully there aren't too many major oversights in this. Anyway, I really appreciate you taking another look at this after all this time.

PR #5862, opened 2024-02-26
Dutch language update 1.1.21

petervanv:
NOTE: i saw a lot off undefined strings, please review

example: [[[SERVERNAME]]] - Account Invitation\nAn account was created for you on server [[[SERVERNAME]]] ([[[SERVERURL]]]/[[[URLARGS1]]]), you can access it now with username "[[[ACCOUNTNAME]]]" and password "[[[PASSWORD]]]".\n~\nBest regards,\n~[[[USERNAME]]]

### DIFF
--- a/translate/translate.json
+++ b/translate/translate.json
@@ -26,6 +26,7 @@
     },
     {
       "en": " (",
+      "nl": " (",
       "xloc": [
         "default.handlebars->47->1501"
       ]
@@ -5458,6 +5459,7 @@
     },
     {
       "en": "AMT Host",
+      "nl": "AMT Host",
       "pl": "Host AMT",
       "xloc": [
         "default.handlebars->47->380"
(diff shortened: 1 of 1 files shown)


si458:
All looks good to me thanks!
The strings u added are all new ones I've added recently so they should be translated now correctly!

petervanv:
what I mean is this, a piece of text, which is not translated, but also has no function in meshcentral. the location is not filled in (xloc info)

image
image

si458:
yes thats prefectly fine, i spotted this while i was fixing translations, 
its basically OLD text from older versions, which might of been removed in newer versions but not removed from the translate.json,
you will only see xloc IF its found the string somewhere in the txt files or handlebar files!

petervanv:
Maybe an good idea too clean it up ? if it aint nessesary (some parts, are double's)

si458:
@petervanv have a look then, checkout the recent translate.json and see what you can find,
i did a full extract the other day so all strings should be up to date

petervanv:
#5871

PR #5865, opened 2024-02-27
Preservation of the executable rights of the files contained in a plugin

wdlut:
On file systems that support this, the executable rights of a plug-in's files are retained.


wdlut:
I have not tested it on NTFS or FAT file systems. I don't expect any problems, but it would certainly be good if someone tested it before the merge.

Ylianst:
I am going to take it and add an if statement to not do this on Windows.

Issue #5203, opened 2023-06-19
Have Run Commands sent via the meshctrl.js return the output of the remote client


TrevorMcFarlane:
When utilizing MeshCtrl.js, I would find myself not getting much use when it comes to in depth report building. I would like to run commands to keep track of version numbers of installed packages, get modem information, and etc. However, all that returns when using: runcommand --run <command> is "OK".

**Describe the solution you'd like**
I would like to utilize runcommand --run <command> and get output of the terminal.
IE
runcommand --run "tailscale version"
I get the following output
1.3.5<current installed version number>

ckraushaar:
I'm looking for something similar, I was thinking the ability to write events back through meshctrl.js.  So I could add a meshctrl.js call to the end of a batch/powershell script to write Success/Failure to the Events Log.  Or write a script to check program versions and write each of them back to the log.  This would allow you to deploy across multiple devices then check either My Events or the devices individual Event logs for the results.

silversword411:
...or you could use an RMM, which is built for scripting/scheduling etc

https://github.com/amidaware/tacticalrmm

si458:
currently the `runcommand` feature just runs the command and returns 'ok'
and the agentside `meshcore.js` just returns the info into the console tab of the device,
but we can look at changing it to return the reply if its using `meshctrl.js`

TrevorMcFarlane:
> currently the `runcommand` feature just runs the command and returns 'ok' and the agentside `meshcore.js` just returns the info into the console tab of the device, but we can look at changing it to return the reply if its using `meshctrl.js`

I believe that would suffice.

AddoSolutions:
This would be extremely helpful here as well, especially when we need to run scripts etc but have no way to see the output, if it worked, etc.

si458:
sorry for delay in this! have done PR to add this for you!
the new command will be simply adding `--reply` to the command
`node meshctrl.js RunCommand --id 'LONGDEVICEIDHERE' --run "ps aux" --loginuser user --loginpass pass --reply`

PR #5932, opened 2024-03-14
allow meshctrl.js to reply with output from runcommands

si458:
Fixes https://github.com/Ylianst/MeshCentral/issues/5203


TrevorMcFarlane:
Is this code available on the current npm for mesh central? Updated the npm version of meshcentral and it does not appear the meshctrl.js file contains the updated code which references the --reply command.

si458:
No, you have to download the meshctrl.js from the master branch in the repo! 
Npm releases are only when new releases happen 😀

PR #6010, opened 2024-04-10
Website backup to github of resources

silversword411:
https://github.com/Ylianst/MeshCentral/discussions/6008


si458:
what is the reason for this?
as we have double the size of the repo now because you ave added duplicate exes?

silversword411:
I was just replicating meshcentral.com for when it disappears again so we have backups.

Are those .exe's already in the repo somewhere? Can't find them

2024-04-11_113426 - Everything_meshcommander-_-_Everything

silversword411:
If you don't want in main repo just close PR.
I can throw into a separate repo for archival purposes.

Of course it'll then be under my GitHub instead of official MeshCentral stuff.

crass:
I know next to nothing about this issue, so the following may be wrong-headed. Doubling the size of the repo sounds like a bad idea from a network and storage perspective. Many of these files will never be used by the user because they are architecture/os specific. So effectively a waste of space. It seems to make more sense to put them in a separate repo, and host them via github pages for that repo and then have links that point there, allowing just the files desired to be downloaded.

Also, it seems llike some of these changes deserve a separate PR (like the doc changes). I would think the doc changes could be merged more readily than the changes adding the binary files and need not be held back by that.

silversword411:
Well...the full mesh repo folder size is 4+GB.
3.9 (or something like that is all the git tracking stuff in hidden folder)
This does double the live files...but they're just one time binary commits.

I keep seeing issues where people are looking for windows installers etc. and the website has disappeared in the past so this was just a backup for future me if they disappear again.

As I said no biggie either way, I'll just drop them in a new repo under my GitHub if the official MeshCentral repo does take care of putting it into a permanent storage.

Of course then it won't be under ylians GitHub with other mesh stuff

PR #6133, opened 2024-05-28
added optional installflags to AgentDownload

schories:
The helpful MeshCtrl was missing the "installflags" option for the AgentDownload feature.

I've added:

- "installflags" as an *optional* parameter. 
- input validation, which should be consistent with other options.
- documentation/help.

Please let me know if there is anything I missed, and feel free to amend as necessary.

### DIFF
--- a/meshctrl.js
+++ b/meshctrl.js
@@ -909,7 +909,7 @@ if (args['_'].length == 0) {
                     case 'agentdownload': {
                         console.log("Download an agent of a specific type for a given device group, Example usages:\r\n");
                         console.log(winRemoveSingleQuotes("  MeshCtrl AgentDownload --id 'groupid' --type 3"));
-						console.log(winRemoveSingleQuotes("  MeshCtrl AgentDownload --id 'groupid' --type 3 --installflags 1"));
+                        console.log(winRemoveSingleQuotes("  MeshCtrl AgentDownload --id 'groupid' --type 3 --installflags 1"));
                         console.log("\r\nRequired arguments:\r\n");
                         console.log("  --type [ArchitectureNumber]   - Agent architecture number.");
                         if (process.platform == 'win32') {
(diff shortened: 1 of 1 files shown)


si458:
the is a few bugs with the code,
for example if you specify `installflags` but no number, it still downloads the file even tho it should error out
also if you specify 0 for `installflags`, it errors out too and still downloads the file
it should NOT carry on downloading the file,
i believe you should be using `else if` rather than `if`
```
simon@testmc:~/meshcentral/MeshCentral$ node meshctrl.js agentdownload --id '@4WLPzjluG0i1sDV62zYxxxxxx' --type 4 --installflags
Invalid installflags, must be a number.
Downloaded 3518312 byte(s) to "myremoteagent64-mygroup.exe"
```

si458:
ok ive pushed a slightly different version to your for you!
https://github.com/Ylianst/MeshCentral/commit/17cf36edd9eb51d9d08fcb4e3690cde474b795cd
https://github.com/Ylianst/MeshCentral/commit/2b3c329a549fbef3be87bf7884f2930e2c7b58ed
also includes a fix which should error out rather than carrying on

thank you for the helpful code tho!

schories:
> for example if you specify installflags but no number, it still downloads the file even tho it should error out
also if you specify 0 for installflags, it errors out too and still downloads the file

Oops, skipped testing for those. :-(

Thanks for checking & fixing so quickly! And, most importantly, for accepting the PR.

schories:
I am a bit confused, did you merge this into upstream?

si458:
@schories no i didn't merge ur request BUT I pushed my own slightly altered version as the was a bug with agentdownload because it would download regardless of the values u set even if the values where invalid which was worrying! https://github.com/Ylianst/MeshCentral/pull/6133#issuecomment-2135768741

schories:
Excellent, thank you, my browser still showed me an old meshctrl.js - I failed to reload. I'd like to apologize for all the hassle. Obviously I need a break..^^ Thank you such much once again!

si458:
@schories no worries we all have those days including myself!
And thank u for doing the pull request in the first place!
Because of this I wouldn't of spotted the fact it downloaded an agent regardless of what options u set which was a little worrying!

PR #6184, opened 2024-06-17
Added ability to use environment variables as arguments

HuFlungDu:
This adds the ability to use environment variables in place of command line args.

It parses each environment variable starting with "meshcentral_" (normalized) and sets the command line argument after that to the value given. For example:

`MESHCENTRAL_CONFIGKEY="AOEUAOEUAOEU" node ./meshcentral`

Is effectively the same as:

`node ./meshcentral --configkey "AOEUAOEUAOEU"`

It is validated against validArguments and ignores anything that is not set there. Command line arguments also take precedence over environment variables. It shares a precedence with command line variables in regards to overriding config file options.

The motivation for this is for use in scalable cloud environments where it is undesirable to provision the system drive with particular secrets. Currently this can be handled by passing those secrets through command line arguments (`--configkey $(get-key-remote)`, `--mongodb $(get-mongostring-remote)`, etc). However, this method will include the secrets in the process string, and so is readable via ps and also in systemd logs if meshcentral is started through systemd. Those secrets could then be stored in external logs if you are, say, monitoring processes running on your server and flagging abnormalities, or pulling systemd logs for external processing. Allowing these variables to be set via the system environment can allow one to avoid these pitfalls.

This is merely one implementation of this feature. It, for instance, does not allow one to pass bool variables at all (though depending on how meshcentral checks for these arguments, it might be fine to pass truthy strings for true) and also allows someone to pass arguments that may be undesirable (`MESHCENTRAL_HELP="aoeu"` will print the help message, for instance), though this is more just a weird wart than an actual problem, I think. If meshcentral used a schema for minimist this could be solved automatically, but as it is, a solution would require a separately managed list of environment variables to check.

### DIFF
--- a/meshcentral.js
+++ b/meshcentral.js
@@ -146,14 +146,13 @@ function CreateMeshCentralServer(config, args) {
         for (let [envvar, envval] of Object.entries(process.env)) {
             if (envvar.toLocaleLowerCase().startsWith(ENVVAR_PREFIX)) {
                 let argname = envvar.slice(ENVVAR_PREFIX.length).toLocaleLowerCase()
-                if (!!argname && !obj.args.hasOwnProperty(argname) && !(validArguments.indexOf(argname) == -1)) {
+                if (!!argname && !(validArguments.indexOf(argname) == -1)) {
                     envArgs = envArgs.concat([`--${argname}`, envval])
                 }
             }
         }
         envArgs = require('minimist')(envArgs)
-        delete envArgs["_"]
-        Object.assign(obj.args, envArgs)
+        obj.args = Object.assign(envArgs, obj.args)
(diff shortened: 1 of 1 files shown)


silversword411:
What about mc_ as variable...a lot shorter and still do t think there would be much chance of a conflict

HuFlungDu:
> What about mc_ as variable...a lot shorter and still don't think there would be much chance of a conflict

As an avid user of Midnight Commander I take umbridge with that statement 😝

Usually I like to spell things out in environment variables just because they are usually handled by robots or scripts, but there's precedence both ways. Certainly correct that there is very little chance of overlap between meshcentral variables and Midnight Commander variables, though, haha.

PR #6190, opened 2024-06-19
Added "keyagents" domain config option.

HuFlungDu:
# This is an extremely draft PR, more just floating the idea with some code.

Meshcentral has the "lockAgentDownload" option so you can stop people who aren't users from downloading an agent. This works great under the assumption that every user of the instance is to be given a perpetual ability to add agents to the instance. However, given meshcentral has the ability to remove users, any user who is removed can be assumed to remember the meshid they once connected to (or read it from their .msh file in the case they had set their own device up at some point when they had permission) and can now add any number of devices to your instance without being logged in, even if the "lockAgentDownload" option is set.

The thought of this option is to close that hole a bit. If the option is set, downloading an agent will now create a record in the database of that agent being downloaded, along with a random identifier for that download. The first time that agent connects and generates its nodeid, that record will be associated to the nodeid, such that if someone tries to connect a different nodeid using that same key, it will not be allowed. Additionally, when the device is removed from meshcentral, that key is also removed and can never be used to connect again. This allows one to revoke an agent's ability to be added to the instance.

# What isn't done
- I probably didn't implement the key correctly, I just made a random 128 lowercase letter key. I know this project mostly uses sha384 hashes for its keys, but I'm not sure what would be the desired way to set that up here
- I would probably add a timeout to the key, such that if an agent doesn't connect with that key in a certain amount of time, that key is no longer accessible. This would stop someone from downloading an infinite number of agents for future use.
- I would like to create an upgrade path for existing servers. My thought would be to have an option for a grace period wherein any agent that connects will be given a key and updated. All that needs to change agent side is an update to its .msh file to add the key to the connection string, but I haven't found the mechanism for that yet.
- I probably didn't follow the style guide and probably forgot a semicolon or two. It'll probably need cleaned up a bih.

The code as is is functional on a basic test instance, though I only tried it with the default configuration and using the standard windows agent, I'm not sure if it will generalize to all agents, though I believe it will since I modified everywhere it creates the .msh file on the server side.

Let me know if this seems like something meshcentral could benefit from.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -339,6 +339,7 @@
         "isaml",
         "Jitsi",
         "jumpcloud",
+        "keyagents",
         "keyfile",
         "keygrip",
         "keyid",
--- a/meshagent.js
+++ b/meshagent.js
@@ -15,7 +15,7 @@
 "use strict";
 
 // Construct a MeshAgent object, called upon connection
-module.exports.CreateMeshAgent = function (parent, db, ws, req, args, domain) {
+module.exports.CreateMeshAgent = function (parent, db, ws, req, args, domain, key) {
     const forge = parent.parent.certificateOperations.forge;
     const common = parent.parent.common;
     parent.agentStats.createMeshAgentCount++;
@@ -29,6 +29,7 @@ module.exports.CreateMeshAgent = function (parent, db, ws, req, args, domain) {
     obj.remoteaddr = req.clientIp;
     obj.remoteaddrport = obj.remoteaddr + ':' + ws._socket.remotePort;
     obj.nonce = parent.crypto.randomBytes(48).toString('binary');
+    obj.key = key
     //ws._socket.setKeepAlive(true, 240000); // Set TCP keep alive, 4 minutes
     if (args.agentidletimeout != 0) { ws._socket.setTimeout(args.agentidletimeout, function () { obj.close(1); }); } // Inactivity timeout of 2:30 minutes, by default agent will WebSocket ping every 2 minutes and server will pong back.
     //obj.nodeid = null;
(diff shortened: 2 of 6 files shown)


si458:
just a side note, when an agent connects to your server,
meshcentral will auto create the meshgroups if they DID exist but was removed recently.
you can check the function out `function getMeshAutoCreate()` meshagents.js line 600

what might be an idea is we have an option for the server inside config.json under `settings` (not domains)
something like `recreateDeletedMeshGroups` and if its set as `true` it creates the groups
BUT if `false` it then doesnt recreate the meshgroup

si458:
normally if the is value listed in both settings and domains, then its a case of server or domain values
so a value in settings applies to all domains on your server,
but a value in domains only applies to that certain domain,
as you can have multiple domains (multi-tenant if you where)

marclaporte:
@HuFlungDu  We now have monthly community meetings: https://www.reddit.com/r/MeshCentral/comments/1g7xmkh/first_meshcentral_monthly_meeting_coming_this/

Can you come so we can discuss this?

DaanSelen:
Is this still being worked on?

HuFlungDu:
I haven't touched it since the last comment. Never really received any direction about it, so I just left it. Still think it's a good solution to deal with permissions more granularly than the built on agent key option, since that can't be revoked individually.

DaanSelen:
> I haven't touched it since the last comment. Never really received any direction about it, so I just left it. Still think it's a good solution to deal with permissions more granularly than the built on agent key option, since that can't be revoked individually.

I like the idea so therefor I want to perhaps see if the PR has some life. Are the conflicts resolveable?

@si458 What is your opinion?

HuFlungDu:
The conflict is extremely minor, resolving it should be easy.

HuFlungDu:
It's also worth noting that, due to https://github.com/Ylianst/MeshAgent/issues/258, the current upgrade option will not generalize to certain windows agents, as they don't pick up the changes from the msh command. That issue would need to be resolved before this is used, I would think.

PR #6209, opened 2024-06-27
fix filter refilling in chrome and safari

### LINKED ISSUE #6206: Changing filter category menu refills the value in the filter field
**Describe the bug**
If you clear the filter field and change the category (offline, online, all, etc), the filter field is reset to the last text entry. This is occurring on Safari 17.5 on macOS 14.5 and the latest version of Chrome on Win 10.

**To Reproduce**
1. Type something into the filter field.
2. Clear the filter field.
3. Change the category.

The filter field will now have the text that was cleared in step 2.

**Expected behavior**
Changing the category should not have any effect on the content of the filter field.

**Server Software (please complete the following information):**
 - OS: Ubuntu 
 - Network: LAN
 - Version: 1.1.24
 - Node: 16.20.2

si458:
Fixes #6206


vgz (approved):
I can confirm the changes resolve the issue we were encountering on our server described in bug #6206.

PR #6211, opened 2024-06-27
Added "trustedproxy" to the arguments list

HuFlungDu:
Code seemed to suggest this should be in the options, but it was missing.


si458:
yep can confirm it works from the command line!

PR #6295, opened 2024-07-31
Added option to encrypt files in database with pbkdf2 for key derivation and aes-256-gcm for encryption

HuFlungDu:
This is the optional version of this that allows users to choose which version they want to use. For an opinionated and backward compatibility option, see: #6296 

Currently, a simple sha384 hash is used for the key derivation for the encryption key of the key used to encrypt the files which are pushed to the database. This derivation method is fairly secure, however it may be weak to attacks in a theoretical post-quantum world. Changing this to a more computationally expensive algorithm can help with this. Meshcentral already uses pbkdf2 for password storage, so this is just enabling it for data encryption, though the current code for password storage is callback based and the code for encrypting and decrypting data is synchronous, so I did have to reimplement the pbkdf2 usage.

AES-CBC lacks any form of authentication and is therefor susceptible to modification attacks. Switching to AES-GCM solves this problem by adding an authTag.

The downsides of this approach are:
- More iterations means more processing time. Currently the files are encrypted/decrypted in a trivial amount of time, but changing to a high iteration pbkdf2 derivation increases the processing time non-trivially. This is a one time cost paid at encryption time, and then again on service startup or file pulling, depending on haw you implement your infrastructure.
- This is not an automatic update. To update to this usage, a user would have to pull their config files, modify the config file to use the new algorithms, then push again explicitly. A user will not get the security benefits without enabling and repushing the files explicitly.

Concerns with this specific implementation, it is using the same IV for the encryption key and the kdf salt. Since the IV and salt are "public" anyway I believe this is fine, since the same key and IV/Salt will not be used for two different keys, but I'm not a security expert, so I could be wrong about that. If someone knows more and thinks it should be implemented differently, I can easily change it to use different random numbers.


HuFlungDu:
Just to get ahead of this: The CodeQL error here is on the old code that is still here for backward compatibility. That is the issue which this PR is trying to remedy.

si458:
unfortunately @Ylianst wont accept this unless the codeql passed its test, so we need to fix this PR somehow

HuFlungDu:
Hmm... Without deleting the old code and breaking backward compatibility, the only thing I can think of is to leave all the old code the same, then just append a new function at the end of encrypt/decrypt data, but that's just worse code. That'll fix the pull request, but the old code will still be there. We could also just delete the code and remove backward compatibility, but I reckon that would catch people off guard, not everyone reads release notes. @Ylianst, can you weigh in on this?

HuFlungDu:
Closed because the other version was merged.

PR #6296, opened 2024-07-31
Changed database file encryption to pbkdf2 for key derivation and aes-256-gcm for encryption

HuFlungDu:
This is the opinionated version of this that uses these as the default option, and automatically upgrades users whenever they push files to encrypt in the database. For an optional version, see: #6295

Currently, a simple sha384 hash is used for the key derivation for the encryption key of the key used to encrypt the files which are pushed to the database. This derivation method is fairly secure, however it may be weak to attacks in a theoretical post-quantum world. Changing this to a more computationally expensive algorithm can help with this. Meshcentral already uses pbkdf2 for password storage, so this is just enabling it for data encryption, though the current code for password storage is callback based and the code for encrypting and decrypting data is synchronous, so I did have to reimplement the pbkdf2 usage.

AES-CBC lacks any form of authentication and is therefor susceptible to modification attacks. Switching to AES-GCM solves this problem by adding an authTag.

This version also adds a version tag to the encrypted data, so iterations and encryption algorithm can be updated in the future more easilly.

The downsides of this approach are:

- More iterations means more processing time. Currently the files are encrypted/decrypted in a trivial amount of time, but changing to a high iteration pbkdf2 derivation increases the processing time non-trivially. This is a one time cost paid at encryption time, and then again on service startup or file pulling, depending on haw you implement your infrastructure.
- This version is not optional. A user will be required to pay the computation cost once they push their files again.
- The version will only be upgraded on push. If a user is currently using the cbc/simple hash method, they will continue to do so until they push their files again.
- Every encryption or decryption requires testing against the original cbc/simple hash method. As this test is trivial, the cost is minimal, but since no version was encoded in the original implementation, detecting it before testing seems to be impossible.
- There is a small chance of paying an extremely large computational cost. This will happen only if the current data is encrypted with the old method, the user tries to pull it, they use the incorrect password, and the first 2 bytes of the IV are `01 00`. In this case, it will read the next 4 bytes of the IV as the iteration count and possibly attempt an absurd number of iterations, which will silently hang in your processor for an exceedingly long time. As the conditions to trigger this are very specific and require misuse, I considered it an acceptable price to pay.

Concerns with this specific implementation, it is using the same IV for the encryption key and the kdf salt. Since the IV and salt are "public" anyway I believe this is fine, since the same key and IV/Salt will not be used for two different keys, but I'm not a security expert, so I could be wrong about that. If someone knows more and thinks it should be implemented differently, I can easily change it to use different random numbers.


HuFlungDu:
Just to get ahead of this: The CodeQL error here is on the old code that is still here for backward compatibility. That is the issue which this PR is trying to remedy.

si458:
unfortunately @Ylianst wont accept this unless the codeql passed its test, so we need to fix this PR somehow

HuFlungDu:
Hmm... Without deleting the old code and breaking backward compatibility, the only thing I can think of is to leave all the old code the same, then just append a new function at the end of encrypt/decrypt data, but that's just worse code. That'll fix the pull request, but the old code will still be there. We could also just delete the code and remove backward compatibility, but I reckon that would catch people off guard, not everyone reads release notes. @Ylianst, can you weigh in on this?

Ylianst:
Hi. Sorry for the delay. @HuFlungDu , your right, using CBC here is problematic. I don't think many people use the feature of encrypting their configuration files into the database, however, that is not an excuse for not fixing it and I think your solution is a good one. Running pbkdf2 each time we want to encrypt/decrypt is a problem, we can probably do better than that, let me look into this now. I am going to go ahead and accept your PR and do some testing with it.

Ylianst:
@HuFlungDu, I just put a new commit to allow importing config files with new or old encryption system, default is the new one.

Push config files with old system:
```
node meshcentral --dbpushconfigfiles --configkey abc --oldencrypt
```

Push config files with new system:
```
node meshcentral --dbpushconfigfiles --configkey abc
```

When you list the files, you can specify the password to see if it decodes correctly:
```
node meshcentral --dblistconfigfiles abc
```

The return of `dblistconfigfiles` will indicate if decoding was ok or not:
```
>node meshcentral --dblistconfigfiles
agentserver-cert-private.key, 2532 encrypted bytes.
agentserver-cert-public.crt, 1712 encrypted bytes.
amtacm-intermediate1.crt, 1508 encrypted bytes.
amtacm-intermediate2.crt, 1724 encrypted bytes.
```

```
>node meshcentral --dblistconfigfiles abc
agentserver-cert-private.key, 2494 bytes, decoded correctly.
agentserver-cert-public.crt, 1674 bytes, decoded correctly.
amtacm-intermediate1.crt, 1470 bytes, decoded correctly.
amtacm-intermediate2.crt, 1686 bytes, decoded correctly.
```

```
>node meshcentral --dblistconfigfiles incorrect
agentserver-cert-private.key, 2532 encrypted bytes - Unable to decrypt.
agentserver-cert-public.crt, 1712 encrypted bytes - Unable to decrypt.
amtacm-intermediate1.crt, 1508 encrypted bytes - Unable to decrypt.
amtacm-intermediate2.crt, 1724 encrypted bytes - Unable to decrypt.
```

Not going to document this, but it's useful to have as a backup on case something goes wrong and for testing.

HuFlungDu:
Thanks @Ylianst. I originally implemented it as an attempt to cbc first then default because the downside of an incorrectly identified version number can be quite high, but you're right; the average case is probably more important than worst case, and unless we make 65000 versions, it's probably not going to have any collisions anyway.

Ylianst:
Because GCM has integrity checking, we know a collision of the version number will cause an exception so I am counting on that.

I do think pbkdf2 with different salts is super overkill in this case as you are using the same password for all configuration blobs. This said, it's ok but certainly noticeably slower. Since it's just done once on startup, it's not an issues.

HuFlungDu:
Version number collision in this case (first 2 bytes of old hashed version iv match a version number, next 4 bytes of iv are read as iteration count) can cause it to run pbkdf2 with an absurd number of iterations (up to 2^32), which occurs during the hashing time, so gcm won't help in that case. I do think the cases of this will be low to zero, especially if so few people use this feature as you say, and it won't be a problem in any future deployments, but if there does happen to be a bad collision, it will eat all of the user's processor time and appear as if meshcentral is hanging in startup. Maybe put an advisory on this release that it might cause that issue so in case someone Googles "meshcentral hangs on start", it can show that error.

Yeah, the only reason I used different pbkdf derivations here was because the original version used a different hash for each file, and I didn't want to change the API for the encrypt/decrypt functions to handle caching the key.

PR #6301, opened 2024-08-02
AMT: Fix One Click Recovery support

wow0000:
Hello !
This PR fixes Intel AMT: One Click Recovery that is currently unusable due to an impossible condition. It also replace .iso (which is an unsupported format) to .efi to allow BIOS to load the file

You may need to squash commits included in this PR to keep it clean

It has been tested on a Dell Optiplex 7400 AIO (Intel AMT v16) and with iPXE(.efi) that can be found here: http://boot.ipxe.org/ipxe.efi

Interestingly enough, I came across Intel EMA code that do something similar, and there is a bug in its implementation of makeUefiBootParam() where the len argument is ignored, in the case of a number it'll always be 4 bytes. I'm asking myself if this difference would be volontary and better compatibility with more BIOS software

Feel free to ask any questions or modifications 👐

### DIFF
--- a/amtmanager.js
+++ b/amtmanager.js
@@ -1069,7 +1069,7 @@ module.exports.CreateAmtManager = function (parent) {
         if (status != 200) { dev.consoleMsg("Failed to get security information (" + status + ")."); delete dev.ocrfile; return; }
 
         // Check if this Intel AMT device supports OCR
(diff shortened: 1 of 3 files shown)


wow0000:
For the curious and documentation, the boot UI looks like this on a Dell AIO
uefi_http

si458:
Have read docs and all seems correct from what i can tell so will merge!

PR #6320, opened 2024-08-12
Replaced wmic with non-deprecated alternatives

HuFlungDu:
This is a companion pull request with:

In this PR, I just changed the agent core code which uses WMIC. Other changes are in the other PR. I also did not include compiled binaries for the updated agents, because I'm not sure of the release procedure for those.


si458:
this is another reason we cant just switch to powershell yet! https://github.com/Ylianst/MeshAgent/issues/256
it keeps spawning powershells over and over again!
need to fix that FIRST before we can think about using powershell

marclaporte:
@HuFlungDu  We now have monthly community meetings: https://www.reddit.com/r/MeshCentral/comments/1g7xmkh/first_meshcentral_monthly_meeting_coming_this/

Can you come so we can discuss this?

si458:
i want to thank you so much for the help/info on this!
i have done a load of commits which use `win-wmi` instead as it has direct access using the dll files!
https://github.com/Ylianst/MeshCentral/commit/b39235643e36438665d2ca1eb2a4901d03f388d2

PR #6369, opened 2024-09-09
Docs: Adding meshcentral server debug command

silversword411:


si458:
Great suggestion! Can u also add the fact its a FULL debug mode, so will spit everything out

And u can use `--debug amt,web` etc.. so only show certain debug messages?
Basically it's all the values from the servers trace panel

si458:
Sorry, I was talking about the web ui my server trace page when u click tracing and select what u want to look at!
These are all the debug values that's usable with the `--debug` option
Screenshot_20240909_212134_Firefox

But yes, it's also a good idea these should really be added to the `--help` option too!

Ylianst:
I am accept this and add more documentation.

Ylianst:
Done ([Link](https://github.com/Ylianst/MeshCentral/commit/2f34b7e83bdffece29c70b94135a27068615f069))

PR #6403, opened 2024-09-25
Fix/xss

HuFlungDu:
Did some fixes for possible XSS injections.

First: A filename could be constructed such that, when a user edits the file, it will execute a script. This was caused by the default page not doing HTML escape on some filenames.

Second: Fixed an injection possibility regarding query params. One could construct a URL like:

```https://localhost/meshagents?key="><script>alert(1)</script><a+```
Or, in tricky mode:
[Click here to pwn yourself](https://localhost/meshagents?key="><script>alert(1)</script><a+)

I added EncodeURIComponent for most uses of req.query in webserver.js, but I can't guarantee I got them all.


Ylianst:
I just looked at your PR and yes! This is essential. I am releasing a new version right away. Thank you soooo much!!!

Ylianst:
I see ` function cleanReqQuery(req, res) {}` in the PR, I will check. If not needed I will remove.

HuFlungDu:
Oh oops, you're right. I had originally intended on adding an express.use function to clean req.query, but I changed the implementation. Thought I had removed that. It definitely does not need to be there.

Issue #5433, opened 2023-10-16
Unable to connect error, Meshcentral Router...


granad2002:
**Describe your issue**
Hi guys, I am getting the following error when I try to connect to a Windows machine via RDP and Meshcentral Router:

![Unbenannt](https://github.com/Ylianst/MeshCentral/assets/81970211/945bc231-682c-4d13-82e0-cbf878e54905)

In syslog I am getting the following error:

```RangeError: Invalid WebSocket frame: MASK must be set
at Receiver.getInfo (.../node_modules/express-ws/node_modules/ws/lib/receiver.js:284:16)
at Receiver.startLoop (.../node_modules/express-ws/node_modules/ws/lib/receiver.js:131:22)
at Receiver._write (.../node_modules/express-ws/node_modules/ws/lib/receiver.js:78:10)
at writeOrBuffer (node:internal/streams/writable:389:12)
at _write (node:internal/streams/writable:330:10)
at Receiver.Writable.write (node:internal/streams/writable:334:10)
at Socket.socketOnData (.../node_modules/express-ws/node_modules/ws/lib/websocket.js:1162:35)
at Socket.emit (node:events:390:28)
at addChunk (node:internal/streams/readable:315:12)
at readableAddChunk (node:internal/streams/readable:289:9)
at Socket.Readable.push (node:internal/streams/readable:228:10)
at TCP.onStreamRead (node:internal/stream_base_commons:199:23) {
code: 'WS_ERR_EXPECTED_MASK',
[Symbol(status-code)]: 1002
```

It was working for years therefore I think it can be related to the newest version 1.1.13.

Does someone have an idea what the problem is?

Thanks in advance

**Server Software (please complete the following information):**
 - OS: Ubuntu
 - Virtualization: 
 - Network: Reverse Proxy nginx

**Client Device (please complete the following information):**
 - Device:
 - OS: Windows
 - Network: Remote over WAN
 - Browser: Firefox
 - MeshCentralRouter Version: see screenshot

si458:
you have missed your config.json file for us to help you.
are you using cloudflare? - https://github.com/Ylianst/MeshCentral/issues/5302

granad2002:
Yes, thx, would be great if you can have a look at it.

Sorry for missing the config.json, but it should have nothing to do with it as I didn't change anything there and it was working before installing the newest version.

And no, I am not using Cloudflare.

si458:
[...]

Issue #6463, opened 2024-10-18
AGENT WSERR: RangeError: Invalid WebSocket Frame


Jayddee23:
Hello,
I've had a problem since I upgraded the mesh server to version 1.1.32 and generated an https certificate that we applied to the server.
node in v17.4
I was surprised when I installed version 1.1.32 that mesh took into account the customization of the agent in the json configuration, resulting in a reinstallation of mesh in a different directory on all my workstations.
Once the server is launched, I get this continuous spam message:
![image](https://github.com/user-attachments/assets/11d1261b-f808-4452-9cf8-427627e03b52)
A previous ticket had been opened and solved by reinstalling two PCs, but now it's impossible to reinstall everything and clearly identify the source of the problem.
An idea
Best regards,

si458:
plz refer to my comment from your other issue https://github.com/Ylianst/MeshCentral/issues/6462#issuecomment-2422502851
i think your reverseproxy isnt setup correctly,
nothing to do with meshcentral

Jayddee23:
Finally I found 9 computer in default, I managed to uninstall and reinstall mesh on 8 computer remotely via psexec and I still have 1 on which I should intervene physically.
I've checked the dns of the workstations, which are the same as the workstations with no problems, so I don't think there's a dns entry problem.
Difficult to understand, maybe I'll try with a wireshark on the last pc to try to find an explanation and I'll give you my feedback next week.
Have a nice weekend.

PR #6475, opened 2024-10-23
Added global ws error handler

HuFlungDu:
In some cases, a client could send packets that would cause the webserver to error and crash. In my case, I was getting the error:

```
RangeError: Invalid WebSocket frame: MASK must be set
     at Receiver.getInfo (meshcentral/node_modules/express-ws/node_modules/ws/lib/receiver.js:284:16)
     at Receiver.startLoop (meshcentral/node_modules/express-ws/node_modules/ws/lib/receiver.js:131:22)
     at Receiver._write (meshcentral/node_modules/express-ws/node_modules/ws/lib/receiver.js:78:10)
     at writeOrBuffer (node:internal/streams/writable:570:12)
     at _write (node:internal/streams/writable:499:10)
     at Writable.write (node:internal/streams/writable:508:10)
     at TLSSocket.socketOnData (meshcentral/node_modules/express-ws/node_modules/ws/lib/websocket.js:1164:35)
     at TLSSocket.emit (node:events:519:28)
     at addChunk (node:internal/streams/readable:559:12)
     at readableAddChunkPushByteMode (node:internal/streams/readable:510:3) {
   code: 'WS_ERR_EXPECTED_MASK',
   [Symbol(status-code)]: 1002
}
```

Whatever WS handler this is propagating to is not handling the "error" event on the ws object. I couldn't track it down since this only happens in production for me, so I can't replicate it in a safe environment for testing.

While this is error is definitely worth tracking down itself (#5433?), the server should also never crash due to bad client input. This creates a failsafe in the case there is an unhandled error

### DIFF
--- a/webserver.js
+++ b/webserver.js
@@ -6334,6 +6334,8 @@ module.exports.CreateWebServer = function (parent, db, args, certificates, doneF
 
         // Handle all incoming web sockets, see if some need to be handled as web relays
         obj.app.ws('/*', function (ws, req, next) {
+            // Global error catcher
+            ws.on('error', function (err) { parent.debug('web', 'GENERAL WSERR: ' + err); console.log(err); });
             if ((obj.webRelayRouter != null) && (obj.args.relaydns.indexOf(req.hostname) >= 0)) { handleWebRelayWebSocket(ws, req); return; }
             return next();
         });


si458:
thanks for this amazingly we already have a catch in place for the meshagent ws but not web clients 
https://github.com/Ylianst/MeshCentral/blob/141bec559fb291a0f361826d3000851a118e08cb/meshagent.js#L577

linked issue here showing the same thing but for meshagents https://github.com/Ylianst/MeshCentral/issues/6463

volkans80:
I just installed 1.1.38 and got same error.

Jayddee23:
Hello,
I can confirm that I have the same problem, which I can't seem to solve.
The message has evolved since the last update, I am currently in 1.1.43 and here is what spam indefinitely in my console after launching the mesh server:
image
If you need any further information to find the cause, let me know and I'll try to help you as much as I can.
Thank you for your time,

HuFlungDu:
> Hello, I can confirm that I have the same problem, which I can't seem to solve. The message has evolved since the last update, I am currently in 1.1.43 and here is what spam indefinitely in my console after launching the mesh server: image If you need any further information to find the cause, let me know and I'll try to help you as much as I can. Thank you for your time,

On your instance, does this now cause the server to crash? Or does it just fill up the logs? This PR was just to stop it from destroying the server, the root cause is still there. Either way, you should open a new issue or pr for this if you are seeing the issue, this PR should be considered complete.

PR #6487, opened 2024-10-30
Fix archiver error, add backup options and SQLite maintenance

PTR-inc:
Fixed archiver module stopping the backup permanently if it got f.e. an ENOENT warning and amended all error paths with abort.
Added check on loading the zip-encrypt module, as it may be only loaded once.
Added extra directory backup options and zip module glob options to finetune the backup.
Added SQLite dump instead of a filebackup.
Added SQLite cleanup and optimise to the maintenance functions.
Replaced databasetype with consts for readability


si458:
all looks amazing!
have you verified all your changes work with the `autobackup` command in the `my server console` tab ?
let us know when your finished and i will merge!

PTR-inc:
Thank!

Yes, I have a macro to do that test and  put the timer a bit quicker for the autobackupscheduler check.
Checked sqlite, mongodb, mysql dumps+backup on windows and linux. Paths are annoying.

Going to add the missing postgresql dbdump part tonight I think.

PTR-inc:
Added the pg dump part. The pg db create part didn't work, that's fixed now.

Checked almost all db's on win and linux for paths, except mariadb and acebase. These didn't work at all on my setup. Something to look into some other branch in time.

The initial archiver/dbdump fix is done, so I would call this one finished!

meshcentral-config-schema.json
```diff
@@ -96,7 +96,7 @@
         "sqlite3": {
           "type": "boolean",
```
si458, line 97: dont forget to add the `string` type also to the schema file
```
"type": [
  "boolean",
  "string"
],
```

db.js
```diff
@@ -32,6 +32,20 @@ module.exports.CreateDB = function (parent, func) {
     var expirePowerEventsSeconds = (60 * 60 * 24 * 10);         // By default, expire power events after 10 days (864000). (Seconds * Minutes * Hours * Days)
     var expireServerStatsSeconds = (60 * 60 * 24 * 30);         // By default, expire power events after 30 days (2592000). (Seconds * Minutes * Hours * Days)
     const common = require('./common.js');
+    const path = require('node:path');
+    const fs = require('node:fs');
```
si458, line 36: dont use this require format as its a new format!
altho we set meshcentral to a min of node 16,
the is still a tiny/tiny/tiny percentage of people out there in the world
that run a custom version which uses node 10 !?
can you change these to `path` and `fs` plz

db.js
```diff
@@ -749,20 +770,29 @@ module.exports.CreateDB = function (parent, func) {
...
+                            obj.file.run( 'PRAGMA journal_mode=WAL;' );
+                            //Together with the optimize in the maintenance run, see https://sqlite.org/pragma.html#pragma_optimize
+                            obj.file.run( 'PRAGMA optimize=0x10002;' ); 
                             setupFunctions(func);
                         }
                     );
                 });
                 return;
-            } else if (err) { console.log("SQLite Error: " + err); exit(1); return; }
+            } else if (err) { console.log("SQLite Error: " + err); process.exit(0); return; }
```
si458, line 783: this should exit with 1 because the was an error, if it exits with 0, the system thinks it was a normal close with no errors, which the wasnt!

db.js
```diff
@@ -846,37 +876,50 @@ module.exports.CreateDB = function (parent, func) {
...
                 });
             } else { // If not present, create the tables and indexes
-                const pgtools = require('pgtools');
-                pgtools.createdb(connectinArgs, dbname, function (err, res) {
+                //not needed, just use a create db statement: const pgtools = require('pgtools'); 
+                DatastoreTest.query('CREATE DATABASE '+ databaseName + ';', [], function (err, res) {
                     if (err == null) {
                         // Create the tables and indexes
+                        Datastore.connect();
+                        DatastoreTest.end();
```
si458, line 911: surely it would be best to end the test connection before opening a new one to the same database?
otherwise we would have 2 connections open a single point in time, even if its for a few miliseconds

db.js
```diff
@@ -3355,167 +3431,234 @@ module.exports.CreateDB = function (parent, func) {
     }
 
     // Perform a server backup
-    obj.performingBackup = false;
```
si458, line 3358: why has this been removed? as it should default to false otherwises when it sets later the value is undefined

db.js
```diff
@@ -3355,167 +3431,234 @@ module.exports.CreateDB = function (parent, func) {
...
+                cwd: datapathParentPath,
+                ignore: globIgnoreFiles,
+                skip: parent.config.settings.autobackup.backupskipfoldersglob
+            });
+
+            if (parent.config.settings.autobackup.backupwebfolders) {
+                if (parent.webViewsOverridePath) { archive.directory(parent.webViewsOverridePath, 'meshcentral-views');; }
+                if (parent.webPublicOverridePath) { archive.directory(parent.webPublicOverridePath, 'meshcentral-public');; }
+                if (parent.webEmailsOverridePath) { archive.directory(parent.webEmailsOverridePath, 'meshcentral-emails');; }
+            };
```
si458, line 3645: double `;;` at the ending above, my OCD is annoying, please fix

si458 (changes requested):
plz can u fix my suggestions or explain why these happen/set

db.js
PTR-inc: I just moved it to the top of the module for readability, so it's still defined and initialized to false. But that's just part of my defaults, vars at the top. Maybe shouldn't have done that.

si458: oh sorry yes you are correct!
https://github.com/PTR-inc/MeshCentral/blob/adcb2ca8eae2394c61f5bdeaf25260cb80017f10/db.js#L42
thats ok and yeh its better at the beginning
because then in other places of the code we can check if its doing backups or not etc 👍

db.js
PTR-inc: This was changed to gracefully exit, conforming to how it is done elsewhere in the code:
https://github.com/Ylianst/MeshCentral/blob/7928f7fb3039e62ffde71867ec3153a72a305c50/db.js#L97
With a '1' exit, the app tries to restart itself, with '0' not. This seems the proper way to exit as it is a fatal error, with an error message from the app.

si458: you are indeed correct, my mistake!
i just scanned for `exit(1)` and only this exists, while others are exit`(0)`
im having one of those days today i think,
i need some more sleep 💤

db.js
PTR-inc: Done. I did all this with min v16 in mind. Did some checks with a v10 node, everything worked. Also did a mariadb check now, also works.

meshcentral-config-schema.json
PTR-inc: Done.

db.js
PTR-inc: Ah yes, done.

db.js
PTR-inc: Can't have that! Removed.

db.js
si458: thank you! i just wanted all the code the same as the others,
then in the future IF we have to move to `node:xyz` we can just just a find and replace haha

meshcentral-config-schema.json
si458: not quite fixed as you have set it to string,
my example above, you can set both a string or boolean
this is the proper way to do it

PTR-inc: Oops

si458:
@PTR-inc im happy with this PR,
if you are happy and no more issues/pushed you spotted/need i will merge!
thank you again so much for this! ❤️

si458 (approved):

PR #6504, opened 2024-11-06
fix meshctrl addusertodevicegroup arguments

ijustw0rkhere:


si458:
this doesnt need fixing unless you can explain the change
the serverside code already converts `usernames` to `userids`
```
 // Convert user names to userid's
if (command.userids == null) {
  command.userids = [];
  for (var i in command.usernames) {
    if (parent.users['user/' + domain.id + '/' + command.usernames[i].toLowerCase()] != null) { command.userids.push('user/' + domain.id + '/' + command.usernames[i].toLowerCase()); }
    else if (parent.users['user/' + domain.id + '/' + command.usernames[i]] != null) { command.userids.push('user/' + domain.id + '/' + command.usernames[i]); }
  }
}
```

ijustw0rkhere:
Ah, yep, you are correct....

This was supposed to be my first attempt at helping contribute so I figured I would start with something small. In classic fashion, I botched the attempt. This change was one of a few changes I have made in attempt to address some conformity issues in meshctrl. Unfortunately I grabbed one that was cosmetic rather than one that was impactful.

si458:
@ijustw0rkhere sure, no worries, keep at it!

PR #6510, opened 2024-11-07
Added the ability to fetch timestamp for codesigning through https

HuFlungDu:
Some enterprise environments mandate that all communication happens through SSL. This merely enables the user to set their time server to an HTTPS site, e.g. `"agenttimestampserver": "https://timestamp.sectigo.com/authenticode"`.

### DIFF
--- a/authenticode.js
+++ b/authenticode.js
@@ -1566,7 +1566,11 @@ function createAuthenticodeHandler(path) {
         options.protocol = timeServerUrl.protocol;
         options.hostname = timeServerUrl.hostname;
         options.path = timeServerUrl.pathname;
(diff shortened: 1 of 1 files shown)


si458:
Authenticode actually has its own repo and it's own code, this file is a copy from its repo, so might be best to check/update there first 
Check Ylianst/Authenticode-JS repo

HuFlungDu:
Moved the PR over there: https://github.com/Ylianst/Authenticode-JS/pull/8

I'm sure there was a reason for this to be included manually instead of just being added to package.json, but is that reasoning still valid now?

si458:
added anyways thanks for PR 👍

PR #6520, opened 2024-11-10
adjust removemeshuser to allow for shorter user ids

ijustw0rkhere:
I stumbled across this issue when trying to remove a user from a mesh using meshctrl `removeuserfromdevicegroup`. I was attempting to remove `user1` from a mesh and received `Invalid userid`. After adjusting the Id to `user//user1`, the command would proceed without issue. At first I just figured the function required the full Id with the prefix, but thought that odd as `addusertodevicegroup` distinctly does NOT accept the prefix. 

After looking at the code, the function requires a minimum of 8 characters for the user Id. This causes issues when using meshctrl with Ids under 8 characters. This will also cause an issue in the web UI for Ids of only 1 character. User `user//a` does not satisfy the length requirement and cannot be removed from a mesh in the web UI.

### DIFF
--- a/meshuser.js
+++ b/meshuser.js
@@ -2418,7 +2418,7 @@ module.exports.CreateMeshUser = function (parent, db, ws, req, args, domain, use
                     }
 
                     try {
-                        if (common.validateString(command.userid, 8, 1024) == false) { err = "Invalid userid"; } // Check userid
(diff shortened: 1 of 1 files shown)


si458:
yep this is indeed correct thank you,
most other commands all check the value is at least 1 character long,
where as this one checks if its 8, doh!

Issue #6503, opened 2024-11-05
Node 23 crashing meshcentral startup


alevoski:
**Describe the bug**
After a fresh install from scrach, when I attempt to run MeshCentral for the first time, it failed in loop with the message :
"ERROR: MeshCentral failed with critical error, check mesherrors.txt. Restarting in 5 seconds..."

**To Reproduce**
1. Install MeshCentral : npm install meshcentral
2. Run it : node ./node_modules/meshcentral
3. See error

**Expected behavior**
MeshCentral should runs with no errors.

**Server Software (please complete the following information):**
 - OS: Debian 12
 - Virtualization: VirtualBox 7
 - Network: LAN
 - Version: 1.1.33
 - Node: 23.1.0

**Content of mesherrors.txt**
See attached file
[20241105_mesherrors.txt](https://github.com/user-attachments/files/17638110/20241105_mesherrors.txt)

**Additional context**

**Your config.json file**
This is the default configuration

si458:
Oh dear! Node 23 broke meshcentral!!!
Don't use node 23!
Try using node 22

PTR-inc:
This seems to be caused primarily by NeDB, as it's using deprecated functions. Did a quicktest with node 23 with other db's and those work. Maybe it is a signal to put NeDB to rest? See #4398 for an earlier discussion about this.
Possibly for now put a node versioncheck/warning in place?

si458:
@PTR-inc I have already asked @Ylianst on his opinion on the matter, just awaiting his reply

I was thinking of making sqlite the default, but it would mean a single sqlite file, and the size can grow dramatically,

I also didn't want to set like mysql/mongodb etc the default because otherwise the meshcentralinstaller would need a complete change to install/setup those apps, where as sqlite can be included as part of nodejs

If I remember, we can put a check inside the package.json to say that nodejs must be above 16 and below 23, will check tomorrow

PTR-inc:
Sqlite seems like the logical candidate indeed, also given that nodejs has an internal sqlite api in the experimental phase, so in the future no need even for an external module.

What do you mean exactly with single sqlite file?

And the excessive growth issue as discussed in #5532 has been tempered with the addition of the maintenancefunction for sqlite.

alevoski:
> Oh dear! Node 23 broke meshcentral!!! Don't use node 23! Try using node 22

Hi, it works with Node 22. Thanks !

si458:
Glad it works!
[...]

PR #6538, opened 2024-11-16
Check on nedb usage and node >=23

PTR-inc:
'fixes' #6503

### DIFF
--- a/db.js
+++ b/db.js
@@ -3388,7 +3388,7 @@ module.exports.CreateDB = function (parent, func) {
                 } catch (ex) { console.log(ex); }
             });        
         } else {
-            func();
+            func(1, "NeDB is deprecated from Node 23 onward, please consider moving to another databasetype using the '--NeDBtoDB' option.");
         }
     }
 
--- a/meshcentral.js
+++ b/meshcentral.js
@@ -15,7 +15,8 @@
 "use strict";
 
 const common = require('./common.js');
-
+const [nodeMajor, nodeMinor, nodePatch] = process.versions.node.split('.').map(Number)
+    
 // If app metrics is available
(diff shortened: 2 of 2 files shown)


si458:
I have spoken with @Ylianst he wants to keep nedb!
So we are gunna look for another nedb fork that is newer and works with v23 then move to that instead

si458:
If we implemented this PR it would break all existing setups so hense we looking for another module instead 

But also we should inform the user they can use `--nedbtodb` to migrate there existing nedb to any db they setup

PTR-inc:
Maybe already put a deprecation inform/warning in the current version, as a server message?

I quickchecked the original nedb btw and at first glance it doesn't seem like a very big job to fix the deprecated util functions.

si458:
@PTR-inc agreed it looks like a simple fix from the forked repo we use, but there is other repositorys that replaces the `util` completely, which might be more ideal!

Maybe a warning to tell others to move to a different database but will ask @Ylianst

PTR-inc:
If somebody already did it, better ofcourse! Which ones are you looking at?

PTR-inc:
Added the warning. Just for completeness. Do you want to replace the module before the next release?

si458:
@PTR-inc I will speak with @Ylianst and ask his opinion as curre try this pr would stop it working all together, as u have removed the nedb module and also out the warning, let me come bk to u

PTR-inc:
@si458 Node 23 errors out if the current nedb module is used, so exitting gracefully if trying to use nedb seemed like the nicer option to offer with a clear message. But a replacement module would be better for sure.

si458:
closing as we have used a newer module which supports node 23 now 👍

PR #6586, opened 2024-12-05
fix bitlocker/manage-bde cmdline

### LINKED ISSUE #6582: Since the 1.1.35 Update the BitLocker recovery key can no longer be displayed
Since the recent 1.1.35 (or 1.1.34, I jumped fron 1.1.33 directly to 1.1.35) the BitLocker recovery Key button is missing on all Windows Agents (Windows 10 and 11)
I'm running MeshCentral on an Ubuntu Server 22.04 Container in Proxmox
Bildschirmfoto 2024-12-04 um 22 56 06
Next to "Fully Encrypted" there used to be a blue symbol to show the Bitlocker recovery key
Hope this Bug Report is okay, let me know if you need any further Infos
Thanks for your amazing work on MeshCentral!

PTR-inc:
Fixes #6582 
The drive parameter needs a ':' added as the var only has the driveletter.


si458:
Thank you!
Yeh just checking on mobile, I removed it by mistake and few lines up to fix the array to show in the Web ui correctly

Issue #3894, opened 2022-04-19
Linking from external systems to Meshcentral


cgb:
Hey Yilan,

I have a number of other systems that track hosts via hostname, and wondered if it's possible to 'link' into Meshcentral and search by hostname.  As I understand it now, each external system will have to have access to the mesh node id in order to provide that direct link.

e.g. Would it be possible to support:

- `https://mesh/?viewmode=10&devicename=my_device_name`

or support accepting URL encoded search fields on the my device page?

Thanks!

Ylianst:
So, the device's name is not a unique identifier, but if that is not a problem for you, I can add support for it. It will go to any device with the specified name?

Ylianst:
Ok, I just added it, in v1.0.11 you will be able to do this:

```
https://central.mesh.meshcentral.com/?gotodevicename=Central&viewmode=10
```

Works on the mobile app too:

```
https://central.mesh.meshcentral.com/?gotodevicename=Central&viewmode=10&mobile=1
```

Once released, let me know if it works.

Ylianst:
v1.0.11 is out.

cgb:
Hey Yilian, I fire off a wishful request before sleep, wake up and find it's done.  Incredible!

I've updated my instance and tested the feature out successfully.  Thank you for the quick turn around...!

PR #6672, opened 2025-01-07
Added support for &gotodeviceip=x.x.x.x

nmmclwhitehead:
#3894 

similar to this commit only with ip of device

`https://meshcentral/?viewmode=10&gotodeviceip=10.1.1.1`

### DIFF
--- a/views/default-mobile.handlebars
+++ b/views/default-mobile.handlebars
@@ -1304,6 +1304,7 @@
         delete urlargs.viewmode;
         delete urlargs.gotonode;
         delete urlargs.gotodevicename;
+        delete urlargs.gotodeviceip;
         delete urlargs.gotomesh;
         delete urlargs.panel;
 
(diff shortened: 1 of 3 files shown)


nmmclwhitehead:
@si458 correct this is an internal server setup
havnt had much time to dig into the code yet but how is the node.ip populated from the device hardware ip or the agent ip?

thanks for the reminder ill add to that too

si458:
the `node.ip` is the connecting IP adress from the meshagent
so if a machine connects to your meshcentral internally, it would show the IP of the remote device
if a machine connects to your meshcentral over the internet, it would be the IP address of the devices external IP
if the device was AMT it would be again the intel AMT ip address
if it was CIRA it would be the machines external IP address

bottom line is its the IP address the remote device is CONNECTING FROM, its not given to meshcentral, its whatever meshcentral sees the IP address as (if that makes sense)

nmmclwhitehead:
ah makes sense ... thanks for the explanation

i guess its not so useful for remote device or there are multiple devices at the same location

my use case is i can only pull IP address for logged in users(active directory security events) - rather than doing a DNS query for each i though this might be less overhead lol

si458:
no to be honest its a brilliant feature and i can see how it would be helpful to you!
but to myself and mabe others it wouldnt be very helpful
so will merge little later on if you have tested and all is working well!

si458:
1.1.38 released with this PR 👍

Issue #6272, opened 2024-07-23
language selection on login page doesnt copy through with saml authentication


m4zl:
**Describe the bug**
The language selection on the login page doesnt copy through with saml authentication.

**To Reproduce**
1. Go to the login page
2. Click on the language selection and choose something different than browser default language
3. Authenticate with SAML authentication
5. Get redirected to the client overview and see the default browser language.

**Expected behavior**
Redirection after SAML authentication with the previous set language selection.

**Screenshots**
n.A.

**Server Software (please complete the following information):**
 - OS: Windows Server 2016 Standard
 - Virtualization: Virtual Machine on Hyper-V
 - Network: [WAN]
 - Version: [1.1.24]
 - Node: [v21.6.0]

**Client Device (please complete the following information):**
 - Device: [Laptop]
 - OS: [Windows 11 23H2]
 - Network: [Local to Meshcentral]
 - Browser: [Microsoft Edge]

**Your config.json file**
```
{
  "$schema": "http://info.meshcentral.com/downloads/meshcentral-config-schema.json",
  "__comment1__": "This is a simple configuration file, all values and sections that start with underscore (_) are ignored. Edit a section and remove the _ in front of the name. Refer to the user's guide for details.",
  "__comment2__": "See node_modules/meshcentral/sample-config-advanced.json for a more advanced example.",
  "settings": {
    "cert": "xxx",
    "WANonly": true,
    "_LANonly": true,
    "_sessionKey": "MyReallySecretPassword1",
    "port": 443,
    "aliasPort": 443,
    "agentAliasDNS": "xxx",
    "redirPort": 80,
    "npmproxy": "xxx",
    "mongodb": "mongodb://127.0.0.1:27017/meshcentral",
    "mongodbcol": "meshcentral",
    "allowHighQualityDesktop": true,
    "autoBackup": {
      "mongoDumpPath": "c:\\Program Files\\MongoDB\\Tools\\100\\bin\\mongodump.exe"
    }
  },
  "domains": {
    "": {
      "title": "xxx",
      "title2": "meshcentral",
      "titlePicture": "xxx",
      "loginPicture": "xxx",
      "welcomePicture": "xxx",
      "minify": true,
      "newAccounts": false,
      "authStrategies": {
        "saml": {
          "callbackurl": "xxx",
          "newAccounts": true,
          "entityid": "xxx",
          "idpurl": "xxx",
          "cert": "xxx"
        }
      },
[...]

Issue #4882, opened 2022-12-26
Preserve original URL request after signing in with Azure


joeldeteves:
In an effort to bolster security,we have enabled Azure SSO on our Mesh instance and are moving away from traditional username and password sign in.

All works great, except for the fact that we use MeshCentral embedded in TacticalRMM - because if the user needs to reauthenticate, they are directed to the list of agents in their assigned Group instead of the agent that is specified in the URL (in TRMM, when you take remote control it directs the embedded browser viewport to the matching agent in MeshCentral via the URL).

Refreshing the page after signing in goes to the correct URL - so this a minor issue, but annoying nonetheless (subsequent agent connections work fine until the cookie expires).

It would be very helpful if Mesh would preserve the original URL request after being redirected back to Mesh by the Microsoft callback URL.

My understanding is that this is possible however it requires some changes to the way the request is being handled.

Apologies in advance if this request is unclear - happy to demonstrate the issue if needed.

Thank you,

PR #6685, opened 2025-01-11
Update webserver.js - allow saml relaystate in POST request

nmmclwhitehead:
added relaystate and regex check to prevent redirecting to a page outside of the configured server. also checks for the allowed query params

ref #6272

### DIFF
--- a/webserver.js
+++ b/webserver.js
@@ -2843,24 +2843,24 @@ module.exports.CreateWebServer = function (parent, db, args, certificates, doneF
                 var relayState = decodeURIComponent(req.body.RelayState);
                 var serverName = (obj.getWebServerName(domain, req)).replaceAll('.','\\.');
             
-                var regexstr = `(?<=(https:\\\/\\\/(.+?\\.)?'+ serverName + ')\\\/?
-                .*((?<=[\\?&])gotodevicename=(.{64})|
-                gotonode=(.{64})|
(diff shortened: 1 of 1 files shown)


si458:
the is lots more RegExp to check, below is the ones ive found you missed
```
lang
sitestyle
key
locale
user
pass
gotomesh
gotouser
gotougrp
debug
filter
webrtc
hide
```

si458:
What provider did you use to test relaystate? As I just couldn't get it working my side

nmmclwhitehead:
adfs 

ylian's guide is still pretty good https://ylianst.github.io/MeshCentral/other/adfs_sso_guide/


also `relaystate` has to be enable on the adfs server `Set-AdfsProperties -EnableRelayStateForIdpInitiatedSignOn $true`

si458:
@nmmclwhitehead ah right ok!
Never realised windows active directory supported sso by url!
Learn something new every day!

si458:
@nmmclwhitehead thats great!
now we need to set your url as RelayState which then gets passed to your provider!
sorry ive been snowed under so i can have alook at that this weekend when i have some free time!

si458:
this might also fix this old issue too #4882

nmmclwhitehead:
@si458 no problem , whenever you get chance.

i guess it could be if you use the /auth-saml url then check for additional query params. then build and set the relaystate

not sure if you can use relaystate with OAuth, openid etc

nmmclwhitehead:
@si458 okay yeah looks like this can be done with `req.query`

ill have a look ... see what you think after i make the commit :)

Issue #6675, opened 2025-01-08
Old Backup Files Not Deleting


dmccracken-livewire:
I just noticed that backup files from the past few years still exist in the folder where the daily backups are being saved (c:\temp).
Permissions appear to be ok (Full Control), particularly given that the files are being written there.
I searched everywhere I could on-line but didn't see any open/closed cases. 
Suggestions?

Here are the config settings for backup:
	"autoBackup": {
		"backupIntervalHours": 24,
		"keepLastDaysBackup": 15,
		"zippassword": "xxxx",
		"backupPath": "c:\\temp"
	}

 - OS: Windows 2016 Data Center
 - Network: WAN
 - Version: 1.1.37
 - Node: 22.12.0

si458:
whats the output of `backupconfig` from inside the console tab of `My Server`?
also whats the output of running `autobackup` from inside the console tab too?
can u share a screenshot of what files are inside the folder `c:\temp\`?

dmccracken-livewire:
> backupconfig
DB Name: meshcentral
DB Type: SQLite
BackupPath: c:\temp
BackupFile: meshcentral-autobackup-2025-01-08-16-06.zip
Backup Interval (Hours): 24
Keep Last Backups (Days): 15
ZIP Password: Set
Backup IgnoreFilesGlob: 
Backup SkipFoldersGlob: 

> autobackup
Starting auto-backup...
Creating encrypted ZIP
Auto-backup completed.

I had manually deleted all files prior to 10/1/2024 so there currently are only 103 files in the directory.  I've attached a screenshot with the most recent since they all won't fit in the screenshot.


![tempfile](https://github.com/user-attachments/assets/b6419811-1c60-484a-9db5-4c6fbb1195d8)

si458:
thats not right 😆 will look into it for you!
my test machine uses S3 for my backups and the `maxfiles` works
my production used a normal backup but uses `nedb` for its database and works correctly
so it might be a bug with `sqlite`, leave it with me! 👍

dmccracken-livewire:
BTW, we just migrated from nedb to sqlite today, so prior to today all the backups were using nedb
Thanks!

si458:
@dmccracken-livewire erm? im not sure then? as its working perfectly fine here?
i notice a sudden SIZE increase in the zip between 7th 9am and 7th 3pm, 
did you do an upgrade at all then?
if i remember the was a bug with the old backups and keeping them,
[...]

PR #6695, opened 2025-01-14
Autobackup update

PTR-inc:
Added an option to configure the hour during which the autobackup should be done instead of being determined by the service start time to have more control if wanted.

There is also the extra logging for #6675. I tried to replicate the issue, but it worked every time, on windows and linux.
Hopefully the extra logging will help to determine the issue.
Debug logging for the backup was under the 'db' category, moved it to the 'backup' debug category to keep it clean.

Add "debug": "backup" to the config.json for extra logging besides the default start and file removal logging of the autobackup.


PTR-inc:
Found the path bug for #6675, adding the fix and testing

PR #6716, opened 2025-01-20
Dutch language update 1.1.38

petervanv:
update and cleaned up some useless strings


translate/translate.json
```diff
@@ -94387,6 +94402,7 @@
     },
     {
       "en": "batchupload",
+      "nL": "batch upload",
```
si458, line 94405: If I remember this needs to stay the same as en because of a translate bug that affects p16filterevents, I'll check next week when I'm back :)

si458 (changes requested):

translate/translate.json
petervanv: accidentally used a capital letter

si458:
ive fixed the translate issue with option elements here https://github.com/Ylianst/MeshCentral/commit/31f2224a938dddd750e60a20e21f60193a0f82e6

what now happens is
1. if `notrans=1` is on the element, it wont translate both the value of option and its HTML text at all
2. if `notransval=1` is on the element, it wont translate the value of the option, but translate the HTML text instead
3. if neither `notrans` or `notransval` are set, it will translate both 👍

PR #6722, opened 2025-01-22
Update console info command

PTR-inc:
A little update to the info command, added the h option to make the numbers a bit more readable. Converts elapsed time in seconds to days hours minutes seconds and memory to Mb.

@si458 Do you know why all the try's are there? Maybe could be removed?


si458:
the `try { } catch(e) { }` is there to prevent the application crashing
say nodejs has a bug and `process.uptime()` cant be processed for whatever reason,
it would crash, so it just catches the crash, and carrys onto the next thing, and so on

PTR-inc:
I understand the try..catch functionality, but it seems a bit over the top here? Or is there a know issue with the process object? Just curious.

PR #6731, opened 2025-01-25
Move double serverWarnings functionality to common.js

PTR-inc:
While doing #6695 I saw that there were two 'addServerWarning' functions defined in meshcentral.js. Probably a residue of the app evolution, but thought this could be cleaned up to common.js so it can be called the same way from everywhere.

### DIFF
--- a/common.js
+++ b/common.js
@@ -420,3 +420,42 @@ module.exports.uniqueArray = function (a) {
     }
     return out;
 }
+
+// Add a server warning, warnings will be shown to the administrator on the web application
+/* for reference from one of the old functions
+var ServerWarnings = {
+    1: "",
+    2: "Missing WebDAV parameters.",
+    3: "Unrecognized configuration option \"{0}\".",
+    4: "WebSocket compression is disabled, this feature is broken in NodeJS v11.11 to v12.15 and v13.2",
+    5: "Unable to load Intel AMT TLS root certificate for default domain.",
+    6: "Unable to load Intel AMT TLS root certificate for domain {0}.",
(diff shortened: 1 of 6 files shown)


si458:
this cannot be merged as its all over the place with the code!

you shouldnt including `const common = require('./common.js');` in any of the files
they are passed through using `parent.common`

also `addServerWarning` is located the same thing on `parent` or `obj.parent` so its located in the `meshcentral.js`
it should not be moved otherwise it can cause issues

PTR-inc:
Ah yes, I thought the `const common = require('./common.js');` seemed the cleaner way of accessing common functions to me then the `parent.common` way. There are different ways common.js is referenced throughout the modules, `obj.common = require('./common.js');` being another one.

The serverwarning calls aren't called consistently throughout the modules: `addServerWarning/obj.addServerWarning` (meshcentral.js), `parent.addServerWarning` (db.js), `obj.parent.addServerWarning` (webserver.js) and `parent.parent.getServerWarnings` (meshuser.js), so it was an effort to clean it up a bit.

PTR-inc:
Oh and there are multiple conflict because of the latest accepted pr's I see.

DaanSelen:
Is this still relevant as the issue mentioned is completed?

PTR-inc:
I just checked and there are still the 2 functions in 'meshcentral.js' through` obj.addServerWarning` and `function addServerWarning`, so this could still be cleaned up to 1 function.

PTR-inc:
But I will close this one for now, as other stuff is more important to spend time on. I'll keep it on the ever expanding to-do list.

PR #6735, opened 2025-01-26
Note maximum password length in USB key dialog

interfect:
The password validation function for Intel AMT passwords enforces a maximum password length, but there's nothing in the UI to tell the user this or to indicate that a provided password is, specifically, too long.

This adds the maximum length to the UI.


si458:
thank you for this!
you forgot about the normal UI which needs updating too`default.handlebars`
also the will be translations which needs updating but i can sort those for you

si458:
thank you again, i fixed the translations and a few other places here too https://github.com/Ylianst/MeshCentral/commit/cac505e2cdc03bcce151af10af2965faa2e36e7a

PR #6788, opened 2025-02-18
Fix: Init webstate with empty 'object' instead of empty string

PTR-inc:
If a (new) user has no webstate saved yet the app crashes at login. This is because in #6763 the `var webstateJSON = JSON.parse(webstate);` stumbles if webstate is en empty string. By making the webstate default '{}' instead of '' this is fixed.


si458:
nice catch! sorry im poorly at the moment, so things keep slipping through my fingers

PTR-inc:
@si458 No worries, hope you get well!

PR #6790, opened 2025-02-18
Clone: Use structuredClone instead of parse/stringify

PTR-inc:
From node v17 onward there is the function structuredClone to deepcopy an object. Safer and maybe even faster.

### DIFF
--- a/db.js
+++ b/db.js
@@ -699,9 +699,15 @@ module.exports.CreateDB = function (parent, func) {
         return r;
     }
 
-    // Clone an object (TODO: Make this more efficient)
-    function Clone(v) { return JSON.parse(JSON.stringify(v)); }
-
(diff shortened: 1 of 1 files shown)


si458:
this post shows its actually slower - https://github.com/nodejs/node/issues/50320

PTR-inc:
Ah interesting! I got my impression from this discussion: [what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript](https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript).

The fastness was just my quick interpretation of it being a 1 call versus a 2 call. But the extra complexity of properly cloning a complex object costs cycles, obviously.

The most important aspect was the safety of cloning an object, regardless of the complexity, and it being the 'official' way of properly cloning a object.

PR #6812, opened 2025-02-25
fix evaluation of config.settings.agentlogdump

redneck-f25:
If set to the default value `false` in config.json the expression evaluates `true`


si458:
thank you can confirm ur findings! 👍

Issue #6429, opened 2024-10-04
RunCommand not working consistently in peering environment


kubamaruszczak:
**Describe the bug**
After this feature have been added https://github.com/Ylianst/MeshCentral/issues/6404 Send Commands seemed to work in peering environment. I started to use it and I still encounter issue when passing commands to agents connected to peer I'm executing commands from. When I execute below command on peer agent is connected to I receive `OK` in the output. However, command is not executed on the device and if I add `--reply` flag to the command it gets stucked. If I would do the same on the other peer it will work all fine. 

```
/home/ec2-user/.nvm/versions/node/v18.17.1/bin/node /home/ec2-user/meshcentral/node_modules/meshcentral/meshctrl --loginuser $NAME --loginkey $LoginTokenKey --url ws://localhost:443 RunCommand --id 'node//vCgyHmO$lqTsvaMfSRoHs20mUbg4FDNQu20Twafg20FnSvYjEvCNFZcEEYbqiK@p' --run 'touch /tmp/mesh1' --reply
```

**To Reproduce**
1. connect device to peer1
2. connect to the peer1 and execute RunCommand shell command with --reply flag (in that case it will get stucked and command won't be executed)
3. connect to peer2 and execute RunCommand shell command with --reply flag (in that case mesh_id will be returned and command will be executed)

**Expected behavior**
RunCommand should work no matter which host is used to execute command on a device

**Server Software:**
 - OS: Amazon Linux 2023
 - Version: 1.1.31
 - Node: 18.17.1

**Your config.json file**
```
{
  "settings": {
    "cert": "my.domain.url",
    "WANonly": true,
    "LANonly": false,
    "sessionKey": "MyReallySecretPassword1",
    "mongoDb": "mongodb://MeshCentralMongoDBDNS:27017/meshcentral?replicaSet=rs0",
    "mongoDbChangeStream": true,
    "port": 443,
    "redirPort": 80,
    "nightMode": 0,
    "tlsOffload": true,
    "maxInvalidLogin": {
      "time": 10,
      "count": 10,
      "coolofftime": 10
      },
    "rootCertCommonName": "xxx",
    "mpshighsecurity": true,
    "allowLoginToken": true,
    "authLog": "meshcentral-data/auth.log"
  },
  "domains": {
    "": {
      "title": "MeshCentral",
      "newAccounts": false,
      "userNameIsEmail": true,
      "amtAcmActivation": {
        "certs": {
          "mycertname": {
[...]

PR #6825, opened 2025-02-27
fix runcommands in a peering environment

ijustw0rkhere:
#6429


DaanSelen:
Thank you for your service!

PR #6828, opened 2025-02-27
Feature/sync device name to agentname translate de

### LINKED ISSUE #6808: Add option to prefer --agentName over hostname when syncing name to server
On embedded devices in our network the configured agent name (--agentName) changes on outer conditions while the hostname itself does not change. So we need have an option to prefer the agent name (if given as argument) over the hostname.

In this context it would also be nice to opt-in to be able to change the name an the server until then next connection. This makes ist easier to track changes while commissioning.

See pull request and the Screenshot below.

We could use a plugin to detect the circumstance and drop the records from the DB and let the device reconnect.

**Additional context**
Image

redneck-f25:
Resolves #6808 including translation to german (de).


si458:
can you fix conflicts and i will merge!
sorry, my bad, i should of merged ur request first before making changes

si458:
ok ill merge your other request instead #6809
you will just need to do a new PR with translation fixes after u updated ya repo with latest code 👍

Issue #6875, opened 2025-03-12
Resolve dependencies between plugins


redneck-f25:
**Is your feature request related to a problem? Please describe.**
If one plugin depends on the functionality of another, it would be nice if those dependencies would be resolved during download, installation, update and disabling.

This also helps to define a feature-set plugin, which has dependencies to all wanted plugins.

**Describe the solution you'd like**
as in `package.json`, see PR (draft)
```json
{
  "dependencies": {
    "plugin_short_name": "https://example.org/path/to/plugin/config.json"
  }
}
```

**Describe alternatives you've considered**
Install manually.

**Additional context**
For the development and testing, I created some plugin-repos without any functionality in a dedicated [GitHub-organisation](https://github.com/orgs/interim-mc-plugin-collection/repositories).

PR #6877, opened 2025-03-12
[pluginHandler] delete temp file when download is redirected

redneck-f25:
while working on #6875 i realized that the temp file for downloading is not deleted when the request is redirected. This is not a big problem since it's a zero-byte file. anyway it doesn't look good.


si458:
Thank you. Generally, plugins are provided as is without support/warranty 
However, PRs for fixes to the plugin systemt are always welcome to help improve the experience for plugins!

PR #6894, opened 2025-03-18
Allow allowedOrigin to be configured on container startup

madeddie:
We might want to connect this to HOSTNAME or the REVERSE_PROXY settings, since they'll most likely be similarly used.
For now this will at least allow for an easy managing of meshcentral with docker-compose and some reverse proxy.

### DIFF
--- a/docker/readme.md
+++ b/docker/readme.md
@@ -47,7 +47,8 @@ LOCALSESSIONRECORDING=false
 MINIFY=true
 # set this value to add extra arguments to meshcentral on startup (e.g --debug ldap)
 ARGS=
-# set to the hostname (or comma-separated list of hostnames) meshcentral will be reachable on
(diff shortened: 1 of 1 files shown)


si458:
just a mad question, allowedOrigin can also be an array, will be be safe for docker?
`allowedOrigin: ["my.domain.com","othermy.domain.com","random.domain.com"]`

si458:
@madeddie ah right sure no worries! 
in theory the setting up of config.json only happens on FIRST startup, then never again afterwards
because it detects you have a config.json already so uses that instead
i do know people would like to change the environment variables on the fly and it update,
but 1 thing at a time fokes haha

madeddie:
Alright, fixed the logic and readme to allow usage of all forms

Issue #6912, opened 2025-03-21
otplib is unmaintained, perhaps switch to: otpauth?


DaanSelen:
Regarding OTPlib I am going to make a PR for bumping that to its latest version, which will be 5 years old. And unmaintained... Perhaps an alternative: https://github.com/hectorm/otpauth

Proof of unmaintainability:
https://github.com/yeojz/otplib/issues/704
https://github.com/yeojz/otplib/issues/693
https://github.com/yeojz/otplib/issues/425
https://github.com/yeojz/otplib/pull/409

Perhaps a new library?
https://github.com/hectorm/otpauth
https://github.com/hectorm/otpauth/issues/599

si458:
ok so to begin with ive upgraded the otplib to 12.0.1 so we are at least up to date at much as possible
the was only 1 change that was needed as its now done within the otplb itself
i will look into alternative librarys next week when i get chance 👍

DaanSelen:
> ok so to begin with ive upgraded the otplib to 12.0.1 so we are at least up to date at much as possible the was only 1 change that was needed as its now done within the otplb itself i will look into alternative librarys next week when i get chance 👍

It has no priority just one of the things on the unending list of to-do's

DaanSelen:
Bump

DaanSelen:
Maybe I don't know how it all specifically works within JS but I assume old (and not updated) = bad.

DaanSelen:
Can be closed I guess

PR #6913, opened 2025-03-21
[chore] bump otplib to 12.0.1 (5 years ago)

DaanSelen:
@si458 please test if this version is compatible, I don't think its good to keep non-latest or outdated libraries in use.

Related issue: https://github.com/Ylianst/MeshCentral/issues/6912

### DIFF
--- a/meshcentral.js
+++ b/meshcentral.js
@@ -4299,7 +4299,7 @@ function mainStart() {
         if (config.settings.no2factorauth !== true) {
             // Setup YubiKey OTP if configured
             if (yubikey == true) { modules.push('yubikeyotp@0.2.0'); } // Add YubiKey OTP support
-            if (allsspi == false) { modules.push('otplib@10.2.3'); } // Google Authenticator support (v10 supports older NodeJS versions).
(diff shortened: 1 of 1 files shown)


si458:
we cant just upgrade straight to v12 because when jumping to v11 you have to make changes to your code
https://github.com/yeojz/otplib/tree/master?tab=readme-ov-file#migrating-from-v11x
so its the same going to v10 to v12

DaanSelen:
Overall I think its better to upgrade to otpauth, since optlib does not look to be maintained, would you not agree?

DaanSelen:
> we cant just upgrade straight to v12 because you have to make changes to your code https://github.com/yeojz/otplib/tree/master?tab=readme-ov-file#migrating-from-v11x

And other than that this talks about version 11, we are on 10.

si458:
otplib has been downloaded over half a million times in the last month
Just because it's unmaintained doesn't mean it's not good at what it does!
I will think about it

DaanSelen:
> otplib has been downloaded over half a million times in the last week Just because it's unmaintained doesn't mean it's not good at what it does! I will think about it

https://npmtrends.com/otpauth-vs-otplib I think it could be better overall given the support.

DaanSelen:
Closing due to no simple bumping possible.

PR #6977, opened 2025-04-17
externalsignjob - External Code Signing Job

inzi:
# externalsignjob - External Code Signing Job

## This adds an option to config.settings called externalsignjob. 

This pull requests consists of changes to the codesigning.md file detailing the externalsignjob feature. 

## What this does

In meshcentral.js, when calling to code sign the agent, mesh central provides a callback. 

This enhancement attaches the signingArguments to the callback function used for code signing. After MeshCentral has completed its update to the agent (such as setting values from domain, etc.) and signed the agent, callExternalSignJob is called. if externalsignjob is not set, it simply returns.  If it is set, it spawns the value in externalsignjob and passes the path to the file for the externalsignjob to process. 

This doesn't have to be used for code signing. For example, it could be used to calculate MD5s of the agent and posted to a website.

## Why this is needed

Public code signing now requires a hardware token. This allows external processes to do codesigning using the administrator's tool of choice. 

Additionally, signing could be done more securely by having the externalsignjob use an internal or cloud server to securely complete the code signing process for the agent.

### DIFF
--- a/meshcentral.js
+++ b/meshcentral.js
@@ -3411,6 +3411,7 @@ function CreateMeshCentralServer(config, args) {
                         if (err == null) {
                             // Agent was signed succesfuly
                             console.log(obj.common.format('Code signed {0}.', agentSignedFunc.objx.meshAgentsArchitectureNumbers[agentSignedFunc.archid].localname));
+                            obj.callExternalSignJob(agentSignedFunc.signingArguments);
                         } else {
                             // Failed to sign agent
                             addServerWarning('Failed to sign \"' + agentSignedFunc.objx.meshAgentsArchitectureNumbers[agentSignedFunc.archid].localname + '\": ' + err, 22, [agentSignedFunc.objx.meshAgentsArchitectureNumbers[agentSignedFunc.archid].localname, err]);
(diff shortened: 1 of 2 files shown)


meshcentral.js
```diff
@@ -3411,6 +3411,7 @@ function CreateMeshCentralServer(config, args) {
                         if (err == null) {
                             // Agent was signed succesfuly
                             console.log(obj.common.format('Code signed {0}.', agentSignedFunc.objx.meshAgentsArchitectureNumbers[agentSignedFunc.archid].localname));
+                            obj.callExternalSignJob(agentSignedFunc.signingArguments);
```
si458, line 3414: i think this should be moved out of a successful signing,
because meshcentral MIGHT not be able to sign using its own built in method,
but could succeed with the external method
(ive seen this a few times where meshcentral cant sign some agents because they cant contact the timestamp server for example)

meshcentral.js
```diff
@@ -3479,16 +3483,39 @@ function CreateMeshCentralServer(config, args) {
...
         }
 
         if (--pendingOperations === 0) { func(); }
     }
 
+    obj.callExternalSignJob = function (signingArguments) {
+        if (obj.config.settings && !obj.config.settings.externalsignjob) {
+            return;
+        }
+        console.log('############# EXTERNAL SIGNING JOB CALLED ##############');
```
si458, line 3503: its better to use `obj.debug('main', "message goes here")`

meshcentral.js
```diff
@@ -3479,16 +3483,39 @@ function CreateMeshCentralServer(config, args) {
...
+        const { spawnSync } = require('child_process');
+
+        const signResult = spawnSync('"' + obj.config.settings.externalsignjob + '"', ['"' + signingArguments.out + '"'], {
+            encoding: 'utf-8',
+            shell: true,
+            stdio: 'inherit'
+        }); 
+
+        if (signResult.error || signResult.status !== 0) {
+            console.error("External signing failed for file: " + signingArguments.out);
```
si458, line 3514: its better to use obj.debug('main', "message goes here") AND `console.error`

meshcentral.js
inzi: Good point - I'll modify

meshcentral.js
inzi: Will update

meshcentral.js
inzi: Will update

PR #7075, opened 2025-06-02
Dockerfix 2

DaanSelen:
This fixes the weird nullification of the installed NPM modules.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,7 @@
 ### STAGE 1 BUILDING.
 FROM node:lts-alpine3.21 AS builder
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,6 +1,8 @@
 ### STAGE 1 BUILDING.
 FROM node:lts-alpine3.21 AS builder
 
+LABEL org.opencontainers.image.description="MeshCentral Docker Image"
```
si458, line 4: this isnt needed as i think it already sets this from the github repo
https://github.com/Ylianst/MeshCentral/actions/runs/15396375196/job/43318108279
image
image

DaanSelen: image

Where does this come from then?

Issue #6857, opened 2025-03-06
Send keys "hidden". For passwords or confidental Information.


JSuenram:
**Is your feature request related to a problem? Please describe.**
Sometimes when you connect to a remote System is neccessarry to enter a password.
When using a Passwort Manager you often just copy and insert it with the "type" feature.
However, in this moment the passwort is shown in clear text. When doing a screencast or recording this is not really well protected.

**Describe the solution you'd like**
Maybe an option to display it like "*" when using the "type" feature.

**Additional context**
https://github.com/Ylianst/MeshCentral/discussions/6830

JSuenram:
Working on this now....

PR #7082, opened 2025-06-04
Add Feature to send keystrokes as hidden field. #6857

JSuenram:
In #6857 there is a feature requested to have an option to send key strokes "invisible".
This should fix it, but please review first....

### DIFF
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -10278,63 +10278,94 @@
             }
         }
 
-        // Remote desktop typing
-        function showDeskType(text) {
-            if (xxdialogMode || desktop == null || desktop.State != 3) return;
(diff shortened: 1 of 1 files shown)


si458:
You have only edited the dault.handlebars which is the old style ui (classic).

You need to also edit the default3.handlebars which is the new style ui (bootstrap)

You can access it by using `https://mesh.mydomain.com/?sitestyle=3`

Also sorry to be picky,
Please can u make any comments in the code English? `// or /* */`

si458:
@JSuenram happy to merge but can you please address my comment above? sorry to be a pain

JSuenram:
Reworking translations....

PR #7092, opened 2025-06-07
Ukrainian localization fixes

Liub0myr:
This PR makes the Ukrainian localization at least somewhat usable.

Please don't use a machine translator. It's better to have some elements in English than to turn the UI into a "*Guess what is meant*" contest.


si458:
hi, thank you for the translations, but the translations arent machine translated,
they have been contributed by users over the years

Liub0myr:
Hi, thanks for the quick response. I thought this was a machine translation using Google Translate in the "Resource Translator Tool".

si458:
@Liub0myr part of it is translated using google translate,
and i from time to time when we add new things in, HAVE to use it to at least get the translations started!
then we rely on the community members like yourself to help correct it for us as we dont all speak 101 languages haha
but thank you again tho to correctly translate it for us!
if you find an more in the list, dont be afraid to edit them and submit another PR too!
ill have fire 24 hours and merge this in tomorrow to give you time to check the rest over for us!
thank you again tho to help us with the translations ❤️

PR #7125, opened 2025-06-27
Fix mobile page load failure and Ukrainian UI text improvements

Liub0myr:
Thanks for the update 1.1.46. Please don't forget to generate a new `agents/modules_meshcore/coretranslations.json` as the changes from commit #7092 didn't affect agent translation


si458:
agent GUI translations are actually handled by [agent-translations.json](https://github.com/Ylianst/MeshCentral/blob/master/agents/agent-translations.json)
i dont no what `coretranslations.json` is used for?

also new agent translations are ONLY applied to new downloads of the meshagent and not installed one

Liub0myr:
Agent localization uses not only [agent-translations.json](https://github.com/Ylianst/MeshCentral/blob/master/agents/agent-translations.json), but also [coretranslations.json](https://github.com/Ylianst/MeshCentral/blob/master/agents/modules_meshcore/coretranslations.json), which was last changed 10 months ago by commit [3a78cb8](3a78cb83c4b42738eff7e7391cdeca4adda27b6f). 

I also thought that the problem was with outdated agents, so I deleted the `signedagents` directory from `meshcentral-data` and restarted the server, but the problem disappeared only after manually editing `node_modules/meshcentral/agents/modules_meshcore/coretranslations.json`.

If you run [translate.js](https://github.com/Ylianst/MeshCentral/blob/master/translate/translate.js) with a specific language, it runs the `translateEx` function, which doesn't modify the `.json` files. Try running it with the `translateall` parameter, then `translateAllInJson` will run and the problem will be solved.

Liub0myr:
`node translate.js translate uk translate.json translate.json` not working
`node translate.js translateall translate.json uk` working

PR #7149, opened 2025-07-11
Added a “Hide Input” checkbox 

JSuenram:
Summary

Added a “Hide Input” checkbox for remote typing in the Bootstrap 5 interface, allowing users to switch between visible text and password mode

Updated the typing logic to read from the password input when masking is enabled, sending the masked text to the remote system

Introduced the same password-masking option in the sharing interface so that shared sessions can securely send hidden text

Modified the corresponding handler to support the new masked input field when sending keystrokes


si458:
you have edited the default.handlebars file which is the classic ui
you need to also edit the default3.handlebars file which is the new modern ui

also the text is out of alignment so github thinks you have made alot of changes?

PR #7162, opened 2025-07-16
Add backgroundContainer class for modal content in night mode

rscRoman:
Introduces the .backgroundContainer class to improve styling of modal content in night mode in modern ui by setting text color. Updates relevant divs in default3.handlebars to use this class for hardware keys and login events sections. Also adjusts the width of login event entries for better display.

### DIFF
--- a/public/styles/style-bootstrap.css
+++ b/public/styles/style-bootstrap.css
@@ -3315,6 +3315,10 @@ nav .lbbuttonsel2 {
     background-color: black !important;
 }
 
+.night .modal-content .backgroundContainer {
+    color: #000;
+}
+
 .table td,
 .table th {
     padding: 0.05rem !important; /* Decrease padding to reduce row height */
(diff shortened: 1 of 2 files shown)


si458:
is this good to be added?

rscRoman:
Thanks for the feedback!
I'm definitely open to adjustments if needed.
I named the class `.backgroundContainer`, but I’m not entirely sure if that’s the best naming - happy to change it if there's a better suggestion.

I’ve added a screenshot to the PR to show the visual improvement in night mode.
Let me know if you'd prefer a different approach or if anything should be revised further.

si458:
ahh right ok! its a styling issue! 
the is bound to be a few bugs here and there with the nightmode!
dont forget u can also change the theme too in the new modern ui
so have u changed the theme to a different one and checked if it still works?
for example the `Darkly` theme is black from the word go in day or night mode

rscRoman:
Yes, I’ve tested it with several themes, including `Darkly`.
From what I’ve seen, the background in those sections stays in a similar light tone across themes, so the updated text color remains readable and the styling works well throughout.

si458:
great thank you! happily merge it then 👍

PR #7164, opened 2025-07-17
Add initial support for custom stylesheets

### LINKED ISSUE #7144: Custom Stylesheet
Presently when perfoming customization of MeshCentral one can override files by placing a similar named file in `meshcentral-web`. This is a useful feature but risks one ending up with an outdated file over time so anything to minimize large scale copying is desired.

Its generally accepted to have a simple custom stylesheet that can override other css elements in many projects. This is commonly called `custom.css`.

This would be a fairly minor change by doing the following:

1. Add a blank `custom.css` file to `public/styles`.
2. CSS order of precedence when all else is equal is that later entries override earlier entries. This means adding a line to load the `custom.css` as late as possible in the `view` files
3. One can then just create their own `custom.css` file under `meshcentral-web/public/styles` to partially override the stylesheet without need to duplicate a whole stylesheet.

**Additional context**
This is a good first PR for someone looking for an easy one to practice on.

Related to #4903

bundabrg:
Allow an override file `custom.css` that can partially override the stylesheet used for the web interface.

# Goals
-  Allow one to partially override the web UI without risking changes to files becoming stale each time a new MeshCentral update happens. This can be done through the regular method of placing a new `custom.css` into the correct path under `meshcentral-web` or `meshcentral-web-<domain>`

# Issues
- MeshCentral is in the process of changing to a more modern UI using bootstrap. The existing classic UI is in a semi-frozen state and all future development seems to be towards the modern UI. Supporting both is desired but with ideally with minimal changes.
- There are a lot of hardcoded inline styles. These are not compatible with a custom stylesheet unless one were to force it using `!important`. A better option is to move it out of inline and into the imported stylesheet and to also make good use of css-vars for theming. This is considered out of scope here  as one can still use `custom.css` and override the views where needed so behaviour is no worse than without this change but allows future changes to build upon it.

# Todo
- [ ] Update all views where it is useful
- [x] Update Documentation

Closes #7144


si458:
this does sound intresting! is this PR complete?

bundabrg:
It should be complete. Its not perfect as there are plenty of views that have inline styles but it does work well for the two default views and the bootstrap mobile view (since its the same), invite view and a couple of others, so its in 'good enough' state. 

I didn't want to go through and update all the views to pull inline style entries out into their stylesheets, that can always be a separate thing and quite possibly is happening already in the revamped frontend.

Should be no breaking changes as its essentially just adding a 0 length extra stylesheet to several views. 

EDIT: I did have one question. A bunch of the views have an attribute on many style links `keeplink=1`. There's no reference to using this anywhere in the codebase and its not AFAIK a valid html attribute. I ignored it when adding a reference to `custom.css`

si458:
i have 0% idea what keeplink does? (will ask @Ylianst and even asked AI its not got a clue either?)
ive even searched the whole sourcecode too and node_modules but found nothing?
so i think its safe to not include it :)

EDIT: found the original commit where it was added - https://github.com/Ylianst/MeshCentral/commit/1e7a39c6312ece002d1e7d0c87b0ae31b61e2a46
i think it has to do with the custom websitecompiler app that was used for minify and intel amt to shrink down in size
but i dont think it matter anymore

PR #7192, opened 2025-07-30
Update index.md - Embedding Options - for devicename & devicername

tobias9931:
for Commit cb9bc58

### DIFF
--- a/docs/docs/meshcentral/index.md
+++ b/docs/docs/meshcentral/index.md
@@ -778,13 +778,15 @@ As illustrated above, we see the business site using the token key to generate a
 
 ### Embedding Options
 
-There are multiple options available for user to explicitly choose the features that will be loaded from MeshCentral to the business website. The argument in the in the URL can dictate which web page should display and how. The three embedding URL arguments are Viewmode, Hide and Node. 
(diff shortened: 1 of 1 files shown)


si458:
the is actually a load of url arguments which you could add

1. viewmode - which page/view of the device you want to access - (always required!)
2. gotonode - go directly to a node with the provided ID
3. gotodevicename - go directly to a node with its name (nickname, NOT OS name)
4. gotodevicername - go directly to a node with its OS name (generally same as name/nickname)
5. gotodeviceip - go directly to a device with a certain IP
(this wont work for devices with external ip addreses, its meant for internal ips and amt for example)
6. gotomesh - go directly to the mesh general page of the device group

2,3,4,5,6 SHOULD be used with 1 and viewmode can be gotten by checking the url of whatever panel/view you are in now

also ALL VALUES ARE CASE SENSITIVE!

also you make a bobo with text for gotodevicername in its value hehe

tobias9931:
I will add the missing values

si458:
@tobias9931 plz can u resolve conflict and ill merge 👍

tobias9931:
new in #7479

PR #7227, opened 2025-08-11
refac(docker): half the Docker image size due to deferment of compilation

DaanSelen:
Sounds fancy doesn't it?


DaanSelen:
Not quite how I like it yet.

DaanSelen:
This image should do it. It now prepackages `modern-syslog`. However it cannot be added to the `package.json` because then the `npm install` tries to reinstall (and fails) with python (which is not there)

DaanSelen:
Perhaps thinking of it, if it gets approved. Squash commit it (that should remove my internal bickering).

DaanSelen:
Have people been able to test this?

si458:
sorry merged in wrong order, can u fix the conflicts plz?

DaanSelen:
Ofc! Please test is as well. And check the methods which I am using regarding the JS parts

PR #7232, opened 2025-08-14
encode nodeid parameter in connectTunnel url

kubamaruszczak:
Targets issue with special characters in nodeid which was discussed in https://github.com/Ylianst/MeshCentral/discussions/7229


si458:
thanks! can indeed confirm that fixes the issue!
its sending the `$$` over the websocket correctly (ID MUST be in quotes `'LONG$$HERE'`)
but it didnt send correctly as part of the URL and was missing a `$` symbol!

PR #7251, opened 2025-08-28
Add MkDocs configuration and tasks for local documentation setup 

silversword411:
via vscode run and debug


si458:
is this good to add? as will indeed add it 👍

silversword411:
Should be, just makes it easier to launch mkdocs locally via vscode run and debug when testing locally

Issue #7274, opened 2025-09-16
GetNodeRightsCache not invalidated properly


HuFlungDu:
**Describe the bug**
`GetNodeRightsCache`:
 https://github.com/Ylianst/MeshCentral/blob/b66948132a28c63f47c921b127194ffe141820bd/webserver.js#L9162

is not properly invalidated on permissions change. This is mostly not an issue in normal usage, as it is cleared every 10 seconds, which is normal human timeframes, but it can became a problem when interacting with the server programmatically.

**To Reproduce**
For example, have 2 users, each with a websocket connected, user 1 has admin permissions on a device, user 2 has no permissions on that device. User 2 tries to run commands on the device, but fails (this fills `GetNodeRightsCache`) and complains to user 1. User 1 adds user 2 to the device, and user 2 immediately sends a "runcommands" request through their websocket. The command will execute on the device (does not consult `GetNodeRightsCache`), but the msg it returns with the return data gets routed through:
https://github.com/Ylianst/MeshCentral/blob/b66948132a28c63f47c921b127194ffe141820bd/webserver.js#L9426

which will consult `GetNodeRights`, which gets a cache hit and says user 2 does not have remote permissions on the device, and will therefor not forward the return data to user 2, and they will wait forever for a response (or timeout or try again, rerunning the command). This will continue until the users wait 10 seconds for the cache to auto-invalidate.

**Expected behavior**
During `adddeviceuser` and `addmeshuser`, the GetNodeRightsCache for that user/mesh/node combo should be invalidated, allowing user 2 to receive a response from the server immediately.

si458:
hmmm so `GetNodeRights` is used `EVERYWHERE`
so im the only thing we could do is completely remove the cache thing?
also the function only passes 3 values, none of which would help identify where it was called from

EDIT: or maybe ask your bots/automation/staff to SLOW THE FU*K DOWN AND CHILLACKS A LITTLE

HuFlungDu:
I created 2 PRs as options to explain what I'm getting at. We don't need to remove the caching or track from where the cache check originated, just need to invalidate the parts of the cache that are no longer valid once a device or mesh is updated.

PR #7279, opened 2025-09-17
Added cache invalidation functionality

HuFlungDu:
Invalidate GetNodeRightsCache entries when a user's permissions on a device or mesh.

This is a proposal for #7274.

In short, changing permissions can invalidate cache values, causing some users to not get permissions they should, or causing a user to continue having permissions for 10 seconds after they are removed from a device/mesh.

This version uses a hierarchal cache in order to allow cache invalidations for various scopes (user, mesh, device). The upside is that it allows us to empty the cache selectively based on what permissions are modified. Downside is the cache lookup is a bit more complicated (time should be about the same since there's no strcat happening), and cache invalidation of the entire user needs to iterate over all its children in order to do a correct cache count manipulation. This could be avoided if we were just fine with cache count being wrong and flushing slightly more often, but I didn't want to make that executive decision.

### DIFF
--- a/meshuser.js
+++ b/meshuser.js
@@ -5776,6 +5776,7 @@ module.exports.CreateMeshUser = function (parent, db, ws, req, args, domain, use
                         if (db.changeStream) { event.noact = 1; } // If DB change stream is active, don't use this event to change the user. Another event will come.
                         parent.parent.DispatchEvent(targets, obj, event);
                     }
+                    parent.InvalidateNodeCache(newuser, node.meshid, node._id)
                 }
             }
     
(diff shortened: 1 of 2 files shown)


si458:
if you happy for me to merge this then plz change to ready :)

HuFlungDu:
I guess I prefer this solution.

si458:
@HuFlungDu is this still good to go sorry? happy to merge!
was the any other pr that was needed to go with this one?

HuFlungDu:
The other PR is a secondary recommendation in case it was decided to be a better implementation.

I will run tests on this against the latest version of meshcentral and let you know if it is still good to merge.

si458:
yes plz do!
im going to ask Ylian to do a new release tomorrow/wednesday so if you can before then test!

HuFlungDu:
Sorry I was a little late on this.

Tested against 1.1.53 and I don't see any issues and it performs as intended.

DaanSelen:
Not stale I guess.

silversword411:
> Not stale I guess.

All depends, what's its TTL? LOL

PR #7280, opened 2025-09-17
Fix/invalidate node rights cache  flush

HuFlungDu:
Invalidate GetNodeRightsCache entries when a user's permissions on a device or mesh.

This is a proposal for https://github.com/Ylianst/MeshCentral/issues/7274.

In short, changing permissions can invalidate cache values, causing some users to not get permissions they should, or causing a user to continue having permissions for 10 seconds after they are removed from a device/mesh.

This version does not change the caching mechanism, instead in the case of a mesh change, it will dump the cache for all users. The alternative is to iterate over all cache entries, which seems like a bad idea. However, this will cause a bunch of cache misses every time a user has their permissions changed, which is not ideal either.


si458:
if you happy for me to merge this then plz change to ready :)

si458:
happy to merge if u can fix the conflict plz 👍

HuFlungDu:
This was an alternative to #7279. Since that was merged, this does not need to be. I'm closing this.

PR #7289, opened 2025-09-21
Geolocation and lock for android mobile device

mpedy:
Added lock option for mobile device and a geolocalization tab for getting position of android mobile device with a fork of MeshCentralAndroidAgent

CLIENT SIDE: https://github.com/Ylianst/MeshCentralAndroidAgent/pull/42


si458:
Unfortunately, we cannot accept this PR because the package.json has been modified with packages that won't work on Linux machines/docker

Also u say this works with a fork of the meshagent for Android rather than the official build

What might be better is u convert this into a plugin!
https://ylianst.github.io/MeshCentral/meshcentral/plugins/
Then people can install it if they want to use it

EDIT: ok, I had a think, we might accept it, but u need to fix the package.json, please, as this will break linux machines!
Also please can u verify the android app etc too with the Google play store

mpedy:
I restored package.json and package-lock.json as they were previously. For Android, I'm going to check it asap thank you!

si458:
Thanks for fixing that!

If possible, please can u also apply ur changes to the new modern ui? `default3.handlebars`

We aren't really making any major changes to the old ui anymore and we are trying to push people towards the new ui!

You can use the new ui by clicking the diamond (top right) and click the bottom 3 lines icon OR by adding `?sitestyle=3` into the url bar

mpedy:
I think there is a typo in default3.handlebars at row 9343: `setModalContent('xxAddAgent', "Device Action", x)` should be `setModalContent('xxAddAgent', "Device Action", xxdialogTag)`. This causes the time div of the actions to not be displayed in the dialog mode

si458:
thank you @mpedy! indeed the modern ui had a bug! have pushed your fix here and checked it works https://github.com/Ylianst/MeshCentral/commit/4295901fc1eb10edac101766754811ec179fc50a
but also it just goes to show you how many people ACTUALLY use android agent, its very few as this was NEVER spotted in about a year of trialing the new modern ui

si458:
plz can u resolve these conflicts and we might be merging it!?

si458:
just another one, we already use openlayers v4.2.0
would it be possible for your to use this instead of leaflet?
ideally i would like to replace openlayers with leaflet but the is so much openlayers already does that would need replacing

mpedy:
@si458 I have ported map from leaflet to openlayer. Please consider to remove Content-Security-Policy from webserver.js if that's okay for you: there is an open route for *.openstreetmap. Regarding Custom CSS and JS tags, I placed them under webserver.js with no extra config file modification for using leaflet but, if it's not necessary anymore, consider only last commit.

si458:
@mpedy we already add to the CSP IF you have geolocation enabled
`const geourl = (domain.geolocation ? ' *.openstreetmap.org' : '');`

`'Content-Security-Policy': "default-src 'none'; font-src 'self' fonts.gstatic.com data:; script-src 'self' 'unsafe-inline' " + extraScriptSrc + "; connect-src 'self'" + geourl + selfurl + "; img-src 'self' blob: data:" + geourl + " data:; style-src 'self' 'unsafe-inline' fonts.googleapis.com; frame-src 'self' blob: mcrouter:" + extraFrameSrc + "; media-src 'self'; form-action 'self' " + duoSrc + "; manifest-src 'self'"`

pofa2007:
> thank you @mpedy! indeed the modern ui had a bug! have pushed your fix here and checked it works [4295901](https://github.com/Ylianst/MeshCentral/commit/4295901fc1eb10edac101766754811ec179fc50a) but also it just goes to show you how many people ACTUALLY use android agent, its very few as this was NEVER spotted in about a year of trialing the new modern ui

we are about to add 600 tablets to a selfhosted one, will it be a good test ? :)

si458:
@pofa2007 we havent updated the android agent to include these feature yet, this is why we havent merged it yet
still some way a way

pofa2007:
@si458 we found on old Android releases the current agent still works, on the last Android release it does not .
is there an estimated date to an apk update ?

thanks and regards

PR #7310, opened 2025-09-30
manageAllDeviceGroups - Added group support (members of)

stephannn:
Hi,
I did two modifications:

1. manageAllDeviceGroups
When adding a group to the array manageAllDeviceGroups like:
`"manageAllDeviceGroups": [ "ugrp//RAsh5nG6h0kk$SadJut@cJsiQekjwisy1IYqntP68VWNG@2zvn6r32hMJAC4RLYj" ]`
also the members of the group are super user and can access all devices in the specified domain.

2. Mapping between users and groups
I noticed that when I restart MeshCentral, the relation between user and its groups are lost. When logging in, the connection is only in the memory, but does not recover after a restart. The function just loops through the groups that still have the information about the user matching and restores the connection. This is in relation with bullet point 1, because even when a user is a member of a group, the super user role was lost in case of a restart.

### DIFF
--- a/meshcentral-config-schema.json
+++ b/meshcentral-config-schema.json
@@ -686,7 +686,7 @@
           "items": {
             "type": "string"
           },
-          "description": "Users in this list are allowed to see and manage all device groups within their domain. For example ['user//billybob','user//fred'] would allow billybob and fred to manage all device groups from all users in their domain."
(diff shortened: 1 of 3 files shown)


si458:
absolute bob on! 👍 
thank you for also updating the schema file too! 🎉

si458:
@stephannn this PR has broke a setup for a few people!
https://github.com/Ylianst/MeshCentral/issues/7371#issuecomment-3439414221

t0g3pii:
> @stephannn this PR has broke a setup for a few people! [#7371 (comment)](https://github.com/Ylianst/MeshCentral/issues/7371#issuecomment-3439414221)

*for almost everyone (even with fresh install)

PR #7319, opened 2025-10-02
autoAcceptIfLocked

stephannn:
Hi,
I actually just copied the pull from @si458 (https://github.com/Ylianst/MeshCentral/pull/6759) to make his work also available if the Desktop is locked. I only added it for the Desktop, not terminal, not file share, because in my opinion I does not make sense. 

Besides, I noticed that when my device is locked, I updated the meshcore.js, the locked status was not detected. Saying, I had to lock it again

### DIFF
--- a/db.js
+++ b/db.js
@@ -1656,12 +1656,6 @@ module.exports.CreateDB = function (parent, func) {
                     func(err, performTypedRecordDecrypt(docs));
                 });
             }
-            obj.GetNodeByComputerName = function (domain, rname, func) {
-                sqlDbQuery('SELECT doc FROM main WHERE type = $1 AND domain = $2 AND JSON_EXTRACT(doc, "$.rname") = $3', 
-                    ['node', domain, rname], function (err, docs) {
-                        func(err, performTypedRecordDecrypt(docs));
-                    });
-            };
             obj.Remove = function (id, func) { sqlDbQuery('DELETE FROM main WHERE id = $1', [id], func); };
             obj.RemoveAll = function (func) { sqlDbQuery('DELETE FROM main', null, func); };
             obj.RemoveAllOfType = function (type, func) { sqlDbQuery('DELETE FROM main WHERE type = $1', [type], func); };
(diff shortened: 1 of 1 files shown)


si458:
this is amazing thank you!
also you do need to apply it as well for the files and terminal, but you can sort of copy what you have done already!i

stephannn:
I was thinking about the Files and Terminal and I disagree with you ;). I tell you why. When the screen is locked, you can connect but you still have no access to any files unless you know the password. However, when the Screen is locked (the user is away) you would basically have access to local files and processes, even if maybe it requires usually a user consent. So I think there would only be two ways. Leave it as it is, or create another property for file and terminal.

Edit: 
Basically with these properties:
```
"userConsentFlags": {
					"desktopnotify": true,
					"terminalnotify": false,
					"filenotify": false,
					"desktopprompt": true,
					"terminalprompt": false,
					"fileprompt": false,
					"desktopprivacybar": true
			 },
```
you are still able to do a file or terminal access when the user shouldn't be informed anyways.

si458:
@stephannn unfortunately, most of the people who asked for this feature as an extra to the one i did, have consent enabled for everything!
so the new option would need to also need to work for the terminal/files as well

but anyways its still a good start! i will just add the missing bit for terminal/files for you unless u fancy looking into it :)

stephannn:
Then I would suggest splitting IfLocked and IfNoUser also into these three areas (terminal, files, desktop). It will be minimally more code but you can fine granular your security settings. If you like, I can do that and do a new request

si458:
If you have time, then I'm always happy for for PRs make my life easier, haha

Only thing, plz keep the old values (including ya new one) but make them apply to all the options.

Then the new options can be fine-tuned, as u say to each feature of meshcentral!

Then we can tell people they can now finetune those options to desktop/terminal/files, etc!

(It's actually a good idea now that you mension it, and im thinking about it!)

stephannn:
@si458 , I did the modifications. Please tell me if that is fine. Personally, I would be happier just with Desktop, Terminal and Files instead a forth one for all options, but I also don't mind

si458:
LGTM 👍

PR #7337, opened 2025-10-11
Update plugins page Back Button

The-Dev-Ryan:
Just a simple update to the Back Button on the plugins page "p19"


si458:
do you have any before/after pics? as it looks like you have removed some text `Plugins`
so we might need to re-run the translations etc

The-Dev-Ryan:
Hey, I totally forgot to add the word Plugins back in for the main version. (I removed page titles from my own Version :) 


**New:**
image

**Current:**
image

> do you have any before/after pics? as it looks like you have removed some text `Plugins` so we might need to re-run the translations etc

Issue #6990, opened 2025-04-22
Option to Remove Duplicate Devices in MeshCentral


Sinwinz:
Good afternoon,
I kindly ask to add a feature that allows removing duplicate devices from MeshCentral. In large networks, this becomes a significant issue. Parsing duplicates from JSON and deleting them via Node.js doesn't work reliably.

si458:
this is very difficult to achieve as how do we know whats a duplicate device?
what makes a duplicate device identifable for an existing device?

the issue is actually **why** did the device get duplicated?

and the answer is very simple, it gets duplicated
because the meshagent uses MAC addresses on the device to make each device unique along with other things.
BUT the year is 2025 and devices use VPN, and they are being turned on and off all the times,
and unfortunately this make the MAC addresses change
so when meshagent then restarts/starts, the mac addresses have changed, and when it creates its meshid,
and then compares to whats stored in its `.db` file but again unfortunately is different 
so meshagent goes `no worries ill use the new meshid`

**how can i fix this behaviour going forward you ask?**
you should add `"agentConfig": ["skipmaccheck=1"]` into your `domains` section in your `config.json`
then on any new installs of meshagent, it adds this value into the `.msh` file and tells meshagent,
if the mac address changes, dont worry about it, keep the same meshid inside the `.db` file

**how can u fix existing setups you ask?**
you simply add `skipmaccheck=1` into the existing `.msh` file which should be located next to your binary files or .exes
and then restart the meshagent service

**how can i fix it without connecting into every single computer you ask?**
you can (in theory) run `msh set skipmaccheck "1"` from the group actions `run command`
then run `service restart` to restart the meshagent service

**can i get back my original first meshid for the system?**
unfortunately this isnt possible, unless you HAPPEN to know the exact mac address of the ORIGINAL network/vpn interfaces

also do note, bluetooth mac address can effect the change as well, and even using windows `private mac address` feature has the same effect! as every time the device connects to the wifi, it will change the wifi cards mac address!

Sinwinz:
"Can you please advise how to remove devices that have already been duplicated? I can parse the devices by their last online time, but I’m unable to delete them via script."

si458:
@Sinwinz ask AI maybe? also how do you name your devices? do u customise the names?

the only thing i can think of is use the API to get all devices,
then bundle the devices together based on the OS hostname,
then check if they online or offline
if offline delete that one?

i would have to test it im afraid

tobias9931:
[...]

PR #7339, opened 2025-10-13
Duplicated Devices

stephannn:
Hi,
I made a new pull request about duplicated nodes with the same name in MeshCentral. When the property preventDuplicateDevices is set to true and an agent tries to register with an existing name again, the old node gets removed and the new one added.
 Please let me know what you think

Edit: https://github.com/Ylianst/MeshCentral/issues/6990

### DIFF
--- a/db.js
+++ b/db.js
@@ -1657,7 +1657,7 @@ module.exports.CreateDB = function (parent, func) {
                 });
             }
             obj.GetNodeByComputerName = function (domain, rname, func) {
-                sqlDbQuery('SELECT doc FROM main WHERE type = $1 AND domain = $2 AND JSON_EXTRACT(doc, "$.rname") = $3 ORDER BY lastbootuptime', 
+                sqlDbQuery('SELECT doc FROM main WHERE type = $1 AND domain = $2 AND JSON_EXTRACT(doc, "$.rname") = $3', 
                     ['node', domain, rname], function (err, docs) {
                         func(err, performTypedRecordDecrypt(docs));
                     });
@@ -2220,7 +2220,7 @@ module.exports.CreateDB = function (parent, func) {
             obj.GetAllIdsOfType = function (ids, domain, type, func) { sqlDbQuery('SELECT doc FROM main WHERE (id = ANY ($1)) AND domain = $2 AND type = $3', [ids, domain, type], function (err, docs) { func(err, performTypedRecordDecrypt(docs)); }); }
             obj.GetUserWithEmail = function (domain, email, func) { sqlDbQuery('SELECT doc FROM main WHERE domain = $1 AND extra = $2', [domain, 'email/' + email], function (err, docs) { func(err, performTypedRecordDecrypt(docs)); }); }
             obj.GetUserWithVerifiedEmail = function (domain, email, func) { sqlDbQuery('SELECT doc FROM main WHERE domain = $1 AND extra = $2', [domain, 'email/' + email], function (err, docs) { func(err, performTypedRecordDecrypt(docs)); }); }
-            obj.GetNodeByComputerName = function (domain, rname, func) { sqlDbQuery('SELECT doc FROM main WHERE type = "$1" AND domain = $2 AND doc->>\'rname\' = $3 ORDER BY lastbootuptime', ['node', domain, rname], function (err, docs) { func(err, performTypedRecordDecrypt(docs)); });};
+            obj.GetNodeByComputerName = function (domain, rname, func) { sqlDbQuery('SELECT doc FROM main WHERE type = $1 AND domain = $2 AND doc->>\'rname\' = $3', ['node', domain, rname], function (err, docs) { func(err, performTypedRecordDecrypt(docs)); });};
             obj.Remove = function (id, func) { sqlDbQuery('DELETE FROM main WHERE id = $1', [id], func); };
             obj.RemoveAll = function (func) { sqlDbQuery('DELETE FROM main', null, func); };
             obj.RemoveAllOfType = function (type, func) { sqlDbQuery('DELETE FROM main WHERE type = $1', [type], func); };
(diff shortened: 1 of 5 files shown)


DaanSelen:
I propose a git commit squash

si458:
We discussed this at the meeting, we are going to have a think.
Only reason being someone brought up the issue of what happens to all the old devices history?
So its power status, the events etc, they would all get deleted :(
So just a rethink for the moment!

Edit: maybe we could do a nodeid replace across the database!?

stephannn:
> We discussed this at the meeting, we are going to have a think. Only reason being someone brought up the issue of what happens to all the old devices history? So its power status, the events etc, they would all get deleted :( So just a rethink for the moment!
> 
> Edit: maybe we could do a nodeid replace across the database!?

it was also about some plugins?!?

I am very new to MeshCentral, but I noticed, now the burner, it has not a real relation db schema. Every information is just stored in a JSON in a doc. So to do a replace, every JSON has to be analyzed. Tomorrow I will take a look again at the DB.

Besides, it is just an option that can be chosen, it is not forced. But yes, I am welcomed for ideas and I will take a look what I can do.

A further question. I had to change the main function to async, because I had trouble receiving the value from the DB without. In case no one has seen it ;)

si458:
@stephannn yes the plugins was a sudden brain wave I had in my mind.

For example the script task plugin has its own database 

But it stores information like nodeid ran this command

This command is run on against that node

If the node suddenly got deleted

The plugin would be like WTF where's the node gone?

But indeed your unique case as u explained about the AD made a valid point! So the hostnane would always be unique etc so its easy to clean up etc and u wasn't too bothered about history etc!!!

Im thinking of maybe merging it but adding a disclaimer in the schema and also displaying a warning in the `my server` page so if they enable it they know that old information will be permanently deleted when the device reconnects etc with a new node id etc :) (brain wave im just having now haha)

stephannn:
@si458 , I took a look again and I think I can add an update for the table power to replace the nodeid with the new one. However, I still have two question. I think when an agent registers, it does not transfer der BIOS ID to check if the new node with the same hostname is really the same device that is going to be replaced. This information will just be in the sinode doc later on.
The other point is, I think I can also add a function that moves old duplicate devices to a different mesh when configured, instead of deleting. Not sure what do you think

DaanSelen:
What is the current status of:
Condition -> Action?

stephannn:
Sorry for letting the topic fade away a bit. I made following process (no real process language) to check if a device exists:
image

It has basically two variables: AvoidDuplicates and MoveDuplicates
If the first is set to true, the process will start, otherwise, nothing will change to the current setup. The second variable will move a duplicate device to another mesh if specified. So that maybe only admins still can see the device, but other IT agents wont.
I also thought about a BIOS Serial check, but Simon explained me that this is more an enterprise feature. Consumer hardware does not have a serial and it will be null. However, this would also require a change in the agent because currently on the first contact, the BIOS SN is not transmitted. In my opinion I still would love this, because even when you are in a Windows domain and can only have unique names, sometimes a device just resets the name of another device. The old device is of course not a member of the domain anymore, but at least you still would see this in MeshCentral.

Besides, a function to move a client in the meshagent.js seems to be missing.

Maybe we can discuss this in the next monthly meeting.

silversword411:
For all those TPM2 machines out there aka win11+
```ps
Get-TpmEndorsementKeyInfo
```
Will give you an unchangeable tracking SN.

You don't think that's abused by people wanting to track you...do you?

stephannn:
Another idea we had in the monthly meeting:
The client initially sends TPM, BIOS-SN, MAC and on the server site these values are evaluated based on existence and priority. Also we have to check nicknames, when a client name was changed in the view

What I don't understand, why there exists duplicates out of nothing. the node-ID on the client in the MeshAgent.db should be the same?!?

DaanSelen:
Ongoing @si458 should we increase PR stale marks?

DaanSelen:
@si458 not stale?

silversword411:
I found an edge case...then didn't document the reason. Bah

DaanSelen:
> I found an edge case...then didn't document the reason. Bah

Explain!

stephannn:
I also opened a FR on the agent site: https://github.com/Ylianst/MeshAgent/issues/341

Suncatcher:
I don't know if I observe this same issue, but this is what I observe:
image
The same device appears twice in the list. On top of that, if I rename the device to the human-friendly name, it then resets to the old machinable name, and my renamed device stays always offline, which I think evidence of the same bug

Suncatcher:
@si458 does it looks like this issue, or should I file a separate bug?

si458:
@Suncatcher you need to check the `details` tab of each device and check the `network` mac addresses
if they are different this is why you have multiple devices, because if a mac address changes (random wifi mac, private mac, vpn, etc) when the meshagent restarts it will get a new device id and create a duplicate device
to fix this you need to add `skipmaccheck=1` into the `.msh` next to `meshagent.exe` then restart meshagent
https://github.com/Ylianst/MeshAgent#msh-format

Suncatcher:
> to fix this you need to add skipmaccheck=1 into the .msh next to meshagent.exe then restart meshagent

@si458 the remote OS is Debian 13. Where to add this on Linux?

si458:
@Suncatcher same place, next to the meshagent binary, inside the `.msh` file
normally somewhere like `/usr/local/mesh_services/meshagent/meshagent`
you can also check this by running `info` on the console tab of a device and check the `Application Location`

Suncatcher:
> @Suncatcher same place, next to the meshagent binary, inside the `.msh` file normally somewhere like `/usr/local/mesh_services/meshagent/meshagent` you can also check this by running `info` on the console tab of a device and check the `Application Location`

ok, this helped. It worth making this setting default, as many devices nowadays use random MAC generation on each boot. We cannot rely on MAC staticness anymore, most MACs are dynamic now

si458:
@Suncatcher unfortunately that's not possible without new meshagents being made.
You can howvee add the option into your config.json under domains
`"agentConfig": ["skipmaccheck=1"]`
This then allows any new installs of meshagent to already have it included by default

Also computers and servers mac address don't really change ad they have lan cards where as it is only really wifi and Bluetooth ones that change.

Also remember how long meshcentral has been around for, 10 years or more! So before random mac address became the new norm..

stephannn:
I uploaded a script written in PowerShell to move or remove duplicate devices based on the device serial number:
https://github.com/stephannn/MeshCentralPowerShellScripts

stephannn:
I think the PR/FR is not possible. On the initial contact only the host name and a generated mesh ID are transmitted (obj.agentInfo). No more information about mac or anything.
I haven't check yet the agent and what does the command ""agentConfig": [ "skipmaccheck=1" ], " does exactly, but in my environment, running for 6 months with 180 devices, I never had a duplicate device that was not reinstalled.

si458:
`agentConfig` is a server side value which ADDS values into the `.msh` file BUT you must REDOWNLOAD the exe for them to be included everytime u edit the `agentConfig`
`skipmaccheck` does as it says on the tin, it SKIPS checking mac addresses and uses whatever meshid it already has,
because years ago (and to this day) the meshid is generated from the mac addresses of devices, and on startup of the meshagent it verifies those mac addresses by generating the meshid again and checking they match, if they dont, you get a new meshid!
this was BEFORE the likes of VPNs took off which oversally have their own mac addresses etc and can be there 1 minute and then gone the next!

Issue #7340, opened 2025-10-13
Remove Device name and Group name edit button from anywhere except from the general tab


panreyes:
**Is your feature request related to a problem? Please describe.**
I like clean UIs and simple UX. I prefer to only have the "Edit device/device group name" only in the General tab instead of in every tab.

**Describe the solution you'd like**
Only have the "Edit device/device group name" only in the General tab instead of in every tab.

**Describe alternatives you've considered**
Changing it in my fork or in custom.js, but I consider this change might be good for everyone.

**Additional context**
I will quickly provide a small PR to fix this :)

PR #7341, opened 2025-10-13
Remove Device name and Group name edit button from anywhere except from the General tab

panreyes:
Implements this feature request:
https://github.com/Ylianst/MeshCentral/issues/7340

### DIFF
--- a/meshcentral-config-schema.json
+++ b/meshcentral-config-schema.json
@@ -1922,6 +1922,11 @@
             "default": false,
             "description": "When enabled, this will show the notes panel in the device view"
           },
+          "collapseGroups": {
+            "type": "boolean",
+            "default": false,
+            "description": "When enabled, device groups will be collapsed by default in the devices tab."
+          },
           "userSessionsSort": {
             "type": "string",
             "default": "SessionId",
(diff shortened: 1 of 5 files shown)


si458:
this is better as an argument/variable in config.json
then the admin (you) can decide if it should only be within the general tab 
or elsewhere as well
will ask the community tho for their feedback later today!

edit: can you show a screenshot of before/after?

panreyes:
Sorry for the delay, I didn't see the edit this morning.

Before:
image
image
image
etc...

After:
image
image
image
etc...

The idea is to help reduce a bit the "cognitive overload", to make the new users have a less overwhelming on-boarding.

Currently, most of my customizations in MeshCenter are about removing elements that my users will not actually use. I'm kind of doing what Microsoft did with the task bar, so it might not be the best of the ideas 😅

MarkStephan89:
> this is better as an argument/variable in config.json then the admin (you) can decide if it should only be within the general tab or elsewhere as well will ask the community tho for their feedback later today!
> 
> edit: can you show a screenshot of before/after?

This would be nice, because I find it better to have the Buttons not only in the General tab ✌️
Config.json 👍👍👍

PR #7345, opened 2025-10-14
Japanese localization fix.

kouhei-ioroi:
Hello.
I've been very grateful to MeshCentral for their help and will always be grateful for this project.
In this PR, I've made significant improvements to the Japanese translation.
In particular, the current Japanese translation contains some Korean (Hangul) characters, which made it even more difficult to understand than the English.
This will greatly improve usability for Japanese speakers.
There may be some translations that are not accurate. I plan to continue making similar contributions if I find any areas that need improvement in the future.
Thank you.


si458:
thank you for your contribution!
nobody speaks every language!
so its nice when the community help us fix the translations!

PR #7355, opened 2025-10-17
fix: honor settings.autobackup.backuppath everywhere

wdlut:
make obj.backuppath single source of truth

Deleting old backups failed when using a non-default folder because `settings.autobackup.backuppath` from `config.json` was not respected consistently.

- Set `obj.backuppath` from `config.json` at startup (when configured)

Follow-up: The lookups of parent.config.settings.autobackup in `db.js` are now redundant and should be removed the next time `db.js` is touched.

### DIFF
--- a/meshcentral.js
+++ b/meshcentral.js
@@ -116,9 +116,12 @@ function CreateMeshCentralServer(config, args) {
     });
 
     // Look to see if data and/or file path is specified
-    if (obj.config.settings && (typeof obj.config.settings.datapath == 'string')) { obj.datapath = obj.config.settings.datapath; }
-    if (obj.config.settings && (typeof obj.config.settings.filespath == 'string')) { obj.filespath = obj.config.settings.filespath; }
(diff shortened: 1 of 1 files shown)


PTR-inc:
Please elaborate on your fix. I did a quick check, but don't see the error.

`obj.backuppath` is properly defaulted, defined and checked and db.js only uses `obj.backuppath` through `parent.backuppath` for the backup folder path, see specifically the function `removeExpiredBackupfiles` in db.js.

As for removing the lookups, that is not an option, at least, not the way you state it. They are all lookups for other config options.

wdlut:
Sorry. I'll remove the pull request.
The bug was already fixed  03.11.2024: SHA-1: e58d659fa94a74e6bbbef5198f85ec10e8a3993e
* Fix archiver error, add backup options and SQLite maintenance (#6487)
Due to merge conflicts parts of my code was old.

PR #7360, opened 2025-10-18
Update login token modal and button behavior

The-Dev-Ryan:
Removes orange background from username and password fields in the login token modal, updates modal title to 'Login Token Created', disables the OK button, and changes the Cancel button text to 'Close'. Also added a 'return false' statement to "account_createLoginTokenEx" to keep the modal open upon token creation.

### DIFF
--- a/views/default3.handlebars
+++ b/views/default3.handlebars
@@ -4557,10 +4557,11 @@
                     var x = "Take note of this username and password, the password cannot be shown again." + '<br /><br />';
                     x += addHtmlValue("Name", EscapeHtml(message.name))
                     if (message.expire != 0) { x += addHtmlValue("Expire", EscapeHtml(printDateTime(new Date(message.expire)))); }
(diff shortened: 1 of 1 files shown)


si458:
because you have added a string `"Close"` you need to run the translate so its extractes the new string for translations
`cd translate`
`node translate.js` <- let it install missing modules
`node translate.js extractall`
this should really ALWAYS be done when editing the handlebars files
then inspect the `translate.json` file

edit: also `Login Token Created` will need to be translated too
hint: anything with `" "` gets translated so use this for strings to be translated and `' '` for everything else

The-Dev-Ryan:
> because you have added a string `"Close"` you need to run the translate so its extractes the new string for translations `cd translate` `node translate.js` <- let it install missing modules `node translate.js extractall` this should really ALWAYS be done when editing the handlebars files then inspect the `translate.json` file
> 
> edit: also `Login Token Created` will need to be translated too hint: anything with `" "` gets translated so use this for strings to be translated and `' '` for everything else

Thanks for that mate, I had not looked into the translations side of this :)

Tech-Werkz:
is this already fixed ? 
I'm still encountering now the Create Token Issues
its not continuing after selection of Token name and Duration

PR #7391, opened 2025-10-28
Openid Client Proxy

stephannn:
When the server is run behind a network proxy, the openid-client is not using the system proxy.
The proxy can be set by the docker compose file like:

```
environment:
      - HTTPS_PROXY=http://myproxy.org:9090
      - HTTP_PROXY=http://myproxy.org:9090
      - NODE_TLS_REJECT_UNAUTHORIZED=0
```

NODE_TLS_REJECT_UNAUTHORIZED is important for the message:
`ERR: (node:344) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.`

### DIFF
--- a/webserver.js
+++ b/webserver.js
@@ -8081,23 +8081,71 @@ module.exports.CreateWebServer = function (parent, db, args, certificates, doneF
             }
 
             // Callback function must be able to grab info from API's using the access token, would prefer to use the token here.
-            function oidcCallback(tokenset, profile, verified) {
+            function oidcCallback(tokenset, profile, done) {
+                // Handle case where done might not be the third parameter
+                if (typeof done !== 'function') {
+                    // OpenID Connect strategy calls with (tokenset, done) instead of (tokenset, profile, done)
+                    if (typeof profile === 'function') {
+                        done = profile;
+                        profile = null;
+                    } else {
+                        parent.debug('error', 'OIDC: Unable to find callback function in parameters');
+                        return;
+                    }
+                }
+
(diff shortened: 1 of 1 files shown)


si458:
thank you for this, but this will break existing setups because you are trying to require the http proxy module
but most people wont be using a proxy so the module isnt install and nodejs will crash
we need to add a catch first to check if we can use it or not

stephannn:
@si458 , thanks, you are right. I thought the packages were updates because with the latest release I did not have to modify the package.json. Now when I check the code I see:
`
const hasExistingProxy = process.env['HTTP_PROXY'] || process.env['HTTPS_PROXY'] || process.env['http_proxy'] || process.env['https_proxy'];
if (((config.settings.plugins != null) && (config.settings.plugins.proxy != null)) || (hasExistingProxy)) { modules.push('https-proxy-agent@7.0.6'); } // Required for HTTP/HTTPS proxy support`

I assume I just copy part of the code?

si458:
similar yes, im gunna work on this now as i have some free time
strange one, do you set `npmproxy` in your config.json for NPM to install packages?
or do you just set HTTP/S_PROXY environment and it installs modules?

si458:
thank you for all your help!
i added my version here instead https://github.com/Ylianst/MeshCentral/commit/1bd06ccded89bd48e2264d18117066bcd8f35bbe
it uses a global object instead so it can be referenced elsewhere if we have missed places!

PR #7395, opened 2025-10-29
Remove "Try the new MeshCentral UI" message when using the modern UI

kambereBr:
- Remove unnecessary **"Try the new MeshCentral UI"** message in modern UI
- Show Terms & Privacy in modern UI
- Fix footer styling across screen sizes

Screenshots
Screenshot from 2025-10-29 11-29-10

Screenshot from 2025-10-29 10-49-18

### DIFF
--- a/public/styles/style-bootstrap.css
+++ b/public/styles/style-bootstrap.css
@@ -472,16 +472,13 @@ body {
 }
 
 #footer {
-    -ms-grid-column: 2;
-    -ms-grid-row: 4;
-    grid-area: footer;
-    clear: both;
-    overflow: auto;
+    position: fixed;
+    bottom: 0;
+    left: 0;
     width: 100%;
-    text-align: center;
     background-color: #113962;
-    padding-top: 5px;
-    padding-bottom: 5px;
+    padding: 5px 0;
+    text-align: center;
 }
 
 .fulldesk #footer {
--- a/views/default3.handlebars
+++ b/views/default3.handlebars
@@ -1664,7 +1664,7 @@
         </div>
         <div id="footer">
             <div class="footer1">{{{footer}}}</div>
-            <div class="footer2" style="display:none;">
+            <div class="footer2">
                 <div class="row">
                     <div class="col-md-6 d-flex gap-2">
                     </div>


kambereBr:
> Good idea! Though it might be best as a config option — not everyone would want this shown by default.

@The-Dev-Ryan  there’s already a config option **"showModernUIToggle": true** that determines whether to show or hide the message for the end user.

The-Dev-Ryan:
> > Good idea! Though it might be best as a config option — not everyone would want this shown by default.
> 
> @The-Dev-Ryan there’s already a config option **"showModernUIToggle": true** that determines whether to show or hide the message for the end user.

Sorry mate, I meant adding that as an option for footer2.

kambereBr:
> > > Good idea! Though it might be best as a config option — not everyone would want this shown by default.
> > 
> > 
> > @The-Dev-Ryan there’s already a config option **"showModernUIToggle": true** that determines whether to show or hide the message for the end user.
> 
> Sorry mate, I meant adding that as an option for footer2.

Oooh, got it! Doesn’t the **"Toggle Footer Bar"** option already do that, hide or show the entire footer bar?

The-Dev-Ryan:
> > > > Good idea! Though it might be best as a config option — not everyone would want this shown by default.
> > > 
> > > 
> > > @The-Dev-Ryan there’s already a config option **"showModernUIToggle": true** that determines whether to show or hide the message for the end user.
> > 
> > 
> > Sorry mate, I meant adding that as an option for footer2.
> 
> Oooh, got it! Doesn’t the **"Toggle Footer Bar"** option already do that, hide or show the entire footer bar?

Yes, that option toggles the entire footer, but your changes cause the Terms and Privacy links to always display. I think it would be better as a configurable option, since some users may not want those links and might prefer to use the footer for custom text.

kambereBr:
> > > > > Good idea! Though it might be best as a config option — not everyone would want this shown by default.
> > > > 
> > > > 
> > > > @The-Dev-Ryan there’s already a config option **"showModernUIToggle": true** that determines whether to show or hide the message for the end user.
> > > 
> > > 
> > > Sorry mate, I meant adding that as an option for footer2.
> > 
> > 
> > Oooh, got it! Doesn’t the **"Toggle Footer Bar"** option already do that, hide or show the entire footer bar?
> 
> Yes, that option toggles the entire footer, but your changes cause the Terms and Privacy links to always display. I think it would be better as a configurable option, since some users may not want those links and might prefer to use the footer for custom text.

Okay, if I’m not mistaken, the **Terms and Privacy** links were always shown by default in the Classic UI, so in the Modern UI this is probably a regression introduced during its implementation.

That said, your idea makes sense and could definitely be discussed and implemented if @si458  and the community agree.

si458:
thank you @kambereBr indeed the terms & privacy should be shown by default as in the old one 
and also the fix for the footer styling!

@The-Dev-Ryan we are trying to make sure the modern ui is SIMILAR to the old ui before we make the default switch to it, 
but also @kambereBr is right, you can hide it by setting `"hide": 4` under `domains` in your config.json and this will hide the footer
image

The-Dev-Ryan:
> thank you @kambereBr indeed the terms & privacy should be shown by default as in the old one and also the fix for the footer styling!
> 
> @The-Dev-Ryan we are trying to make sure the modern ui is SIMILAR to the old ui before we make the default switch to it, but also @kambereBr is right, you can hide it by setting `"hide": 4` under `domains` in your config.json and this will hide the footer image

This is not what i meant, I meant having the ability to have the custom footer note without having to force the terms/  privacy links mate.

si458:
@The-Dev-Ryan ah right ok! Yeh sadly at the moment it's not possible but I do like ur idea tho hiding it! So yeh we can maybe add a config.json value in for this :)
I would recommend u open a feature request on github issue just explaining what u want
Then u can have a crack at it and also link the issue etc!
If not ill try have a go this week for u! Just a little tied up at moment with other stuff

PR #7398, opened 2025-10-30
Refactor: Rebuilt Masthead User Dropdown for Cleaner Structure and Modern Behavior

The-Dev-Ryan:
### 📱 Mobile View  
IMG_3453 IMG_3451

---

### 💻 Tablet View
image image1

---

### 🧭 Light/Dark - Updated Leftbar Size
image5 image4

---

### 🪄 Full UI
image6image7

---

### ⚙️ Mobile Permissions Control View
image3  image8

**Toggle Light/Dark** Mode button is Disabled if Light/Dark is forced from config.json 
**Toggle Classic/Modern** is disabled if  "siteStyle": 3, is set in config.json
**Toggle Footer** is disabled if  "hide": 4, is set in config.json


silversword411:
After a little jiggery and pokery lots of little tweaks on space usage. 

TheDevRyan is the baws. Awesome!!!!

views/default3.handlebars
```diff
@@ -2112,120 +2206,13 @@
             }
             args.hide = hide;
             QV('uiViewButton5', !(args.hide & 4)); // Hide the footer toggle button if footer is hidden anyway.
+            QV('toggleFooterMenuItem', !(args.hide & 4)); // Hide the footer toggle menu item if footer is hidden anyway.
+            QV('toggleNightMenuItem', ((features2 & 0x00100000) == 0) && ((features2 & 0x00200000) == 0)); // Hide night mode toggle if permanent night (1) or day (2) mode is set
+            QV('toggleModernUIMenuItem', ((features2 & 0x80000000) == 0)); // Show only if allowed and not forced siteStyle=3
```
si458: i think this shouldnt be forced.
i spoke to a gent today who asked if it was possible to set the web ui to moden ui by default
but then let the staff return to the classic ui if they wanted
with this your are FORCING the style regardless and the staff have no way to return to the classic ui if they choice

silversword411: A theme style should be a default you can set in the server config (and at some point we'll change the schema/default from classic to modern unless already defined in config). 

Then each use should be able to select in the menu their desired theme and it be remembered for their login.

Melo-Professional:
When "siteStyle": 3 is set, the UI button to change classic/modern does nothing. So makes sense for me to remove the button from the UI.
When "siteStyle": 2 or 1 is set, the button actually works.

This PR hides that button ONLY IF "siteStyle": 3 is set.

si458:
@Melo-Professional yes but unfortunately as i explained
i spoke to a gent yday with his meshcentral server and staff of over 100,
and he wanted the staff to by default use the modern ui `siteStyle: 3` 
BUT allow them to switch back to modern ui if they wanted using the switched
this PR would remove the option to switch back which is NOT OK

HOWEVER i do agree with @The-Dev-Ryan what we should have is maybe another config.json option which FORCES the sitestyle and disables the switcher (unless they know about the url `?sitestyle=2`)
maybe something like 
`sitestyle: 3` <- keep this to choose a default style
`forcesitestyle: 3` <- this would make sure everybody uses this style regardless of what they have

another thing the gent also mensioned too was a config.json option to selete a default THEME for the modern ui 
(the color switcher under my accounts)


EDIT: code below from webserver.js
```
var uiViewMode = 'default';
var webstateJSON = JSON.parse(webstate);
if (req.query.sitestyle != null) {
  if (req.query.sitestyle == 3) { uiViewMode = 'default3'; }
} else if (domain.sitestyle == 3) {
  uiViewMode = 'default3';
} else if (webstateJSON && webstateJSON.uiViewMode == 3) {
  uiViewMode = 'default3';
}
```
so it actually checkes the web url first and uses that,
then checks sitestyle in domains and uses that,
then checks the users preference which is stored in db and then switches to use that if set
so i have just realised i think the is a bug here as i think it should be user, then url then domain

The-Dev-Ryan:
Yeah for instance if you have it set as sitestyle 3 currently, and go to use the switcher it does nothing, I always assumed this was your intended function, so the idea was to remove the none working button.

si458:
@The-Dev-Ryan forgive me! im having one of those weeks and its only wednesday!
i just had a lightbulb moment💡 and when facepalm 🤦 
i will look into the bug ASAP!

The-Dev-Ryan:
> @The-Dev-Ryan forgive me! im having one of those weeks and its only wednesday!
> 
> i just had a lightbulb moment💡 and when facepalm 🤦 
> 
> i will look into the bug ASAP! 

No worries, I never thought to check if it was a bug as reading the docs when setting style3 it just makes it seem like when you do it forces it. All good mate. 🙂

si458:
LGTM 👍

PR #7406, opened 2025-11-02
Add mobile bootstrap CSS and mobile view template

The-Dev-Ryan:
Introduces a new stylesheet for mobile bootstrap styling and a new Handlebars template for the mobile default view. This enhances mobile UI support and layout customization.

For testing, Just place both in 

meshcentral-web/views - rename default3-mobile.handlebars > default-mobile.handlebars
meshcentral-web/public/styles - Insert style-mobile-bootstrap.css

Then use Classic UI and check it on Mobile.

### DIFF
--- a/public/styles/style-mobile-bootstrap.css
+++ b/public/styles/style-mobile-bootstrap.css
@@ -1,0 +1,3192 @@
+.masthead-container {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    padding: 8px 16px;
+    background: linear-gradient(to bottom, #104893 0%, #113962 100%);
+    min-height: 50px;
+    color: #c8c8c8;
+    position: relative;
+    width: 100%;
+    box-sizing: border-box;
+    padding-top: calc(8px + env(safe-area-inset-top));
+    padding-left: calc(16px + env(safe-area-inset-left));
+    padding-right: calc(16px + env(safe-area-inset-right));
+    height: calc(50px + env(safe-area-inset-top));
+}
+
+#masthead {
+    background-size: 18rem 2.5rem !important;
+    background-position: center left;
+    background-repeat: no-repeat !important;
+    padding-right: 20px !important;
+    background-image: url(../logo.png);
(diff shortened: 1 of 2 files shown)


marclaporte:
@kambereBr @Ndaboom Any thoughts?

The-Dev-Ryan:
> @kambereBr @Ndaboom Any thoughts?

Don't think we are going this direction as it's an extra 2 files to maintain. 🙂

si458:
we will ask the community at next meeting what they think is best 👍

silversword411:
I probably need to check the files in this PR to see if it's fixed these mobile issues I just noticed in modern:

- When in classic hamburger doesn't have a UI switcher
- file viewer buttons go off screen and can't be interacted with. Should be scrollable left/right.
- file sort is missing and there's just a view style item, can't change sorting.
- the permanent top items needs to dynamically slide or toggle or something, takes up 1/3 of the screen constantly.

The-Dev-Ryan:
> I probably need to check the files in this PR to see if it's fixed these mobile issues I just noticed in modern:
> 
> - When in classic hamburger doesn't have a UI switcher
> - file viewer buttons go off screen and can't be interacted with. Should be scrollable left/right.
> - file sort is missing and there's just a view style item, can't change sorting.
> - the permanent top items needs to dynamically slide or toggle or something, takes up 1/3 of the screen constantly.

Hey mate, the files in this PR are not related to the original Modern UI, have a test/play and you will see 🙂

silversword411:
I reserve the right to blame myself completely for not testing right 😂

silversword411:
....weird. I'm having to F5 refresh when switching from desktop to mobile views using browser developer console.

The-Dev-Ryan:
> ....weird. I'm having to F5 refresh when switching from desktop to mobile views using browser developer console.

You really need to use an actual phone, it does not switch once the page is already rendered. It's a new handlebar. 

Devtools will make things look different to how they are on a real phone.

PR #7451, opened 2025-11-26
fix: Dropdown Menu (Updated Permissions Checks)

The-Dev-Ryan:
This Fixes a small bug, when switching from Fixed Width/Top Bar back to LeftBar that these 3 was failing the Auth Check and not showing.

Not a major issue until next release a simple refresh will fix it for now.


si458:
its always the little things you spot!

PR #7481, opened 2025-12-09
Fix stray character typo in GrabKeyInput key handler

saeedrizwi:
This PR fixes a small typo in the GrabKeyInput function where an extra stray character (c) was present after the onkeypress assignment.


si458:
thank you so much for spotting this!
its only a minor thing but from the looks of git, its been in their for the last 4 years haha
i have done a seperate commit as you need to also run the minify tool to create a minify version also
https://github.com/Ylianst/MeshCentral/commit/f542b61cd41ffddcf2df10af96cefd58282a2506
but thank you again so much for helping the community and the project!

PR #7496, opened 2025-12-18
fix: issue with installed web folder

DaanSelen:
Replace echo message and adjust folder movement logic.


DaanSelen:
This only occurs when the web folder IS a persistent volume AND there is something in there already.

PR #7528, opened 2026-01-02
OIDC - Custom Scopes

stephannn:
Currently the operator is only > 1, but should be >= 1 because otherwise the custom scopes only work when it has more then 2 attributes, but not just 1


si458:
changed to `> 0` instead because if you dont specify a custom scope then the length will be 0

PR #7559, opened 2026-01-17
Create the string for Linux install/uninstall with AgentPort instead of Port

Professor-Heavy:
While every single other agent install uses the AgentPort, the string generated for Linux installation does not, which generates a bad link if AgentPort is specified.

I'm **very** unfamiliar with this type of project, so I haven't been able to verify the changes. I wanted to set up a PR to try and fix it regardless.


si458:
this isnt needed, because the client downloads the binary and a msh file from the web ui which would be hosted on the main port (80/443)
then inside the msh file is the real MeshServer url with the agentport that is required for the agent to connect too

also you have an error in the commit as well which would of broken the web ui `if(agentprox)`

PR #7562, opened 2026-01-19
Fix text overlap for long filename in the filelist

### LINKED ISSUE #7532: Selection window resize
Running commands from a file on a server to a remotely connected device shows a small window, and long script names are displayed on two different lines. Is it possible to resize the selection window or make it resizable?

Image

fz0000:
Fixes #7532. Works well on both Classic UI and Modern UI.

Only tested on Files page and the Run Commands dialog. Please help to test other areas that I may not have covered.

By the way, I found that `<div class="filelist"` is also present in the `public/commander*.htm` files. I’m not sure whether they need to be updated so I didn’t modify them.

files_classic
run_classic
files_modern
run_modern

### DIFF
--- a/views/default-mobile.handlebars
+++ b/views/default-mobile.handlebars
@@ -2779,7 +2779,9 @@
 
                 // Filter out files
                 if ((f.t == 3) && (fileFilter != '') && (f.nx.toLowerCase().endsWith(fileFilter) == false)) { continue; }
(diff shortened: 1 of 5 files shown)


si458:
This is why i love the community! 
They step up and help out when im a idiot or run out of time!
Thank u so much! I will check it all tomorrow and merge!

stephannn:
@fz0000 , that's really great! Just had this on my youtube feed: https://www.youtube.com/watch?v=1fZTOjd_bOQ&t=153s

I case you a bored and looking for another challenge ;), I personally would love a top bar, maybe even with extension type that can also be used to sort by clicking on the column name
image

PR #7566, opened 2026-01-20
Authentication Strategy - support auth with github enterprise instances

ricsc:
As documented here: 

https://github.com/cfsghost/passport-github/pull/17/commits/5283b9c1a4a7d0efaa74ba3809d19218ef0d7e85

passport-github2 allows to specify custom urls. 

I have sucessfully tested the change with my github enterprise instance.

### DIFF
--- a/meshcentral-config-schema.json
+++ b/meshcentral-config-schema.json
@@ -3314,6 +3314,26 @@
                     "type": "string",
                     "description": "GitHub client secret."
                   },
+                  "authorizationurl": {
+                    "type": "string",
+                    "format": "uri",
+                    "description": "The URL used to obtain an authorization grant from the user."
+                  },
+                  "tokenurl": {
+                    "type": "string",
+                    "format": "uri",
+                    "description": "The URL used to exchange a code for an access token."
+                  },
+                  "userprofileurl": {
+                    "type": "string",
+                    "format": "uri",
+                    "description": "The URL used to obtain user profile information."
+                  },
+                  "useremailurl": {
+                    "type": "string",
+                    "format": "uri",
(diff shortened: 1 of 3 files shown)


si458:
if you use OIDC setup instead (which you should be doing and not using third party modules like the github one)
then you can already set the values in question
https://ylianst.github.io/MeshCentral/meshcentral/openidConnectStrategy/

you also need to setup your server for oidc too
https://docs.github.com/en/enterprise-cloud@latest/admin/managing-iam/configuring-authentication-for-enterprise-managed-users/configuring-oidc-for-enterprise-managed-users

si458:
ok ignore my above comment!
you need the `userProfileURL` and `userEmailURL` endpoints which we dont set anywhere for oidc!
nice catch!

ricsc:
> if you use OIDC setup instead (which you should be doing and not using third party modules like the github one) then you can already set the values in question https://ylianst.github.io/MeshCentral/meshcentral/openidConnectStrategy/
> 
> you also need to setup your server for oidc too https://docs.github.com/en/enterprise-cloud@latest/admin/managing-iam/configuring-authentication-for-enterprise-managed-users/configuring-oidc-for-enterprise-managed-users

Unfortunately oidc does not work for me as I cannot change the enterprise setup. I'm working with a company provided enterprise envirnment. I'm only able to change configuration inside my projects.

si458:
@ricsc yes its ok, i quickly googled and looked at docs then our source code and you are correct!
those few extra endpoints are not OIDC standards, so the current OIDC wouldnt work
but adding them to the existing github auth method would be perfectly acceptable!
thank you again for your contribution ❤️

PR #7569, opened 2026-01-21
Add ports in docker for setup without editing config file

diogo24m:
The current setup doesn't allow to change ports without editing the config file, this changes allows different setups for different environments

Make ports configurable
* ports are no longer hardcoded, added `PORT` and `REDIR_PORT` environment variables in `compose.yml`
* `entrypoint.sh` updates `config.json` with the values of `PORT` and `REDIR_PORT`
* defaults are used if variables are not set (`443` for `PORT`, `80` for `REDIR_PORT`)

### DIFF
--- a/docker/Dockerfile-debian
+++ b/docker/Dockerfile-debian
@@ -227,7 +227,8 @@ RUN cd meshcentral \
     && npm cache clean --force
 
 # Expose needed ports
-EXPOSE 80 443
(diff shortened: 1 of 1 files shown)


si458:
This is why I love our community! ❤️
We all come today and do pull requests to make each other's lives easier!
Thank you! Merging shortly!

Issue #7579, opened 2026-01-26
MeshCentral reports wrong glibc version of arm64 agent


DaanSelen:
https://github.com/Ylianst/MeshCentral/blob/4923c71742552178b506f9926a40582651367a11/meshcentral.js#L3240

Shows statically when using Arm64/aarch64 agents that the version is locked to Glibc from: 2016-08-04

From what I saw at my own compiled MeshAgent version, the hard requirements would be GLIBC_2.38
However the UI does not change.

DaanSelen:
Seems like its hardcoded. Regardless of actual GlibC version.

cxvqo:
```
root@nx:/usr/local/mesh_services/meshagent# ldd meshagent
	linux-vdso.so.1 (0x0000007f9e10f000)
	libpthread.so.0 => /lib/aarch64-linux-gnu/libpthread.so.0 (0x0000007f9e097000)
	libutil.so.1 => /lib/aarch64-linux-gnu/libutil.so.1 (0x0000007f9e084000)
	libm.so.6 => /lib/aarch64-linux-gnu/libm.so.6 (0x0000007f9dfcb000)
	libdl.so.2 => /lib/aarch64-linux-gnu/libdl.so.2 (0x0000007f9dfb6000)
	librt.so.1 => /lib/aarch64-linux-gnu/librt.so.1 (0x0000007f9df9f000)
	libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000007f9de46000)
	/lib/ld-linux-aarch64.so.1 (0x0000007f9e0e3000)
root@nx:/usr/local/mesh_services/meshagent# ldd --version
ldd (Ubuntu GLIBC 2.27-9ubuntu1.6) 2.27
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
root@nx:/usr/local/mesh_services/meshagent# 
```
seems fine running on 2.27? maybe you incorrectly find these hard requirements? maybe the minimum version is actually 2.24?

DaanSelen:
> ```
> root@nx:/usr/local/mesh_services/meshagent# ldd meshagent
> 	linux-vdso.so.1 (0x0000007f9e10f000)
[...]

PR #7581, opened 2026-01-27
chore: remove hardcoded glibc version (its not that version)

DaanSelen:
fix: https://github.com/Ylianst/MeshCentral/issues/7579


si458:
changing it in the JS doesnt change it in the web ui, only the backend
you need to also change it within all the handlebar files,
then you also need to do the translate and translate the string also in `translate/translate.json`

DaanSelen:
Fair enough well spotted

Simon Smith ***@***.***> schreef op 27 januari 2026 19:36:57 CET:

DaanSelen:
@si458 do you think this covers it?

PR #7585, opened 2026-01-29
Azure UPN Display

stephannn:
Analog to the UPN feature for on-premises devices I tried something similar for Azure. 

However, I have some thoughts about it:

1) I have to recursive loop through the reg key: HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\IdentityStore\LogonCache\, however JS seems not to support recursive calls. My PowerShell Code looked like that which I used for a previous project:
```
function Get-LogonCachedUser {
	
    $userObj = @()
    $userObjTmp = @()

    Get-ChildItem -Path "Registry::\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\IdentityStore\LogonCache\*" -Recurse | ForEach-Object {

            $userTmp = Get-ItemProperty -Path "Registry::\$($_.Name)"
            if($userTmp) {
                $userObjTmp += ( [pscustomobject]@{
		            UPN = $userTmp.IdentityName
		            NTAccount = $userTmp.SAMName
		            SID = $userTmp.Sid
                    SAM = $userTmp.SAMName
	            })
            }
   
	    }

    $userObjTmp | Group-Object UPN | ForEach-Object {
  
        $userObj += ( [pscustomobject]@{
		                UPN = $_.Name
		                NTAccount = $_.Group.NTAccount | Select-Object -Unique
		                SID = $_.Group.Sid | Select-Object -Unique
                        SAM = $_.Group.SAM | Select-Object -Unique
	                })
    }

     return $userObj | Select-Object UPN, NTAccount, SAM, SID -Unique 

}
```

2) I had to add another helper function getJoinState. In my case, my user was created in the on-premise world, but is synced to Azure. Because of that my user still has a domain, even when the device itself is cloud only.
3) I only added the code to the section "// Setup logged in user monitoring"

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -214,20 +214,14 @@ function getJoinState() {
 
     // 1 Azure AD / Entra ID
     try {
-        const joinInfo = require('win-registry').QueryKey(
-        require('win-registry').HKEY.LocalMachine,
-        'SYSTEM\\CurrentControlSet\\Control\\CloudDomainJoin\\JoinInfo'
-        );
+        const joinInfo = require('win-registry').QueryKey(require('win-registry').HKEY.LocalMachine, 'SYSTEM\\CurrentControlSet\\Control\\CloudDomainJoin\\JoinInfo');
         isAzureAD = Array.isArray(joinInfo.subkeys) && joinInfo.subkeys.length > 0;
     } catch (e) {}
 
     // 2 On-prem AD
     try {
-        const tcpip = require('win-registry').QueryKey(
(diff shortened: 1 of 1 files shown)


si458:
this is brilliant and been very helpful, ive found a few problems/issues :(

for example, 2 users logon to 1 device, both are azuread users,
userObj returns 2 users,
but later when you get the upn, you are just getting the first index in the array, when the could be multiple ones
and then you might end up showing a different users upn and not the currently logged in users upn

also why is the sid and sam in an array when they could be just text?

PR #7598, opened 2026-02-03
feat: filter out Raspberry Pi devices for its boot_mode

DaanSelen:


DaanSelen:
Had to revert some wrong local changes.

PR #7599, opened 2026-02-04
Add allowedFramingOrigins config for CSP frame-ancestors whitelist

JonBons:
- Add settings.allowedFramingOrigins and domain-level override
- Implement Content-Security-Policy frame-ancestors with domain whitelist, always includes self if set
- Adjust X-Frame-Options when external origins are allowed
- Enable iframe feature flag when allowedFramingOrigins is set
- Support array or comma-separated string config

Enables more secure embedding in parent portals (e.g. TacticalRMM) while blocking clickjacking from arbitrary sites. Use of parameter disables X-Frame-Options and uses CSP frame-ancestors when only specific origins should frame the site.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -53,6 +53,7 @@
         "allevents",
         "allowaccountreset",
         "allowframing",
+        "allowedframingorigins",
         "allowfullscreen",
         "allowhighqualitydesktop",
         "allowsavingdevicecredentials",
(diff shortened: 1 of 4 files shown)


si458:
plz can you revolve the conflicts 👍

JonBons:
> plz can you revolve the conflicts 👍

Done, let me know if you need anything else!

si458:
@JonBons no all good! i had to ask AI what this PR was as i didnt understand it!
im losing my brain recently with things!
but the PR is great and i understand what it does exactly now!

PR #7625, opened 2026-02-16
Build revision

stephannn:
Sometimes it is nice to have the patch level (winver) also as information, in special when the agent stopped working ;)

image

Build Number is not the correct wording, it is Update Build Revision (UBR), but sounds awful ?!? Feel free to change it


si458:
this is great, but 1 tiny (hassle) change,
**you forgot to do the translations**
1. `cd translate` <- go into the translate folder
2. `node translate.js` <- this will install extra packages that are needed for the translate
3. `node translate.js extractall` <- this extracts the new translation
4. open `translate.json` and find the `"en": "Build Number"`
5. add each translation string in e.g: `"fr": "Numéro de build"` <- also try alphabetically
6. OPTIONAL: `node translate.js translateall` <- this will translate them all then restart meshcentral and check ur language
7. submit the missing `translate.json` to the PR 👍

stephannn:
@si458 , thanks for the good guide. I hope I did it correctly. I just copied it into an AI for translation

si458:
@stephannn im afraid the file is broken :(

```
      "de": "Web-Socket kann nicht verbunden werden",
      "en": "Unable to connect web socket",
      "es": "No se puede conectar al web socket",
      "fi": "Verkkopistoketta ei voida yhdistää",
      "fr": "Impossible de se connecter à la web socket",
      "hi": "�
```
might be better by the opening the file with a text editor,
then just let the AI work out the lines etc manually with its auto complete,
this is what i normally do, i dont give the whole file to AI as it will break the AI haha

si458:
no that broke again :( 
just delete the file from the PR and ill sort it out 👍

stephannn:
nop, that is not the problem. It was the way how I exported it from the container, but 600 KB are missing now... I think I will just edit the original file.

Edit:
it seems that extractall command did not work as expected

PR #7627, opened 2026-02-17
UPN / loggedOnUser - domain clients

stephannn:
The UPN was not correct when the domain name and the DNS name are not the same, Same for user logon name and SAM name.  Unfortunately, there seems not be a good way to extract the information from WMI or registry. The only way seems to be dsregcmd /status

With this approach the information is still incorrect when logging in with the SAM, but when using the UPN the result looks better.


DaanSelen:
@si458 this looks to be handy!

PR #7649, opened 2026-02-24
File URI

stephannn:
Hi,
I wasn't aware about the GoTo button in Files, so I thought I had to go through the click path.

With this change, you do a double click on the space next to the path:
image
and it will change to a path box:
image


stephannn:
to be discussed in the next meeting

PR #7656, opened 2026-03-01
fix: add 404.html w/ relevant redirects

userandaname:
does right as it says on the tin; re-directs any existing links to where they should be on the new domain for compatibility!

### DIFF
--- a/404.html
+++ b/404.html
@@ -4,11 +4,12 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <script>
-        const newDomain = "https://docs.meshcentral.com";
-        const destination = newDomain + window.location.pathname + window.location.search;
-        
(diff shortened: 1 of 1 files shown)


si458:
oh did i not do it correctly?
i created an index.html page and from what i can see, it works here
https://ylianst.github.io/MeshCentral/ diverts to https://docs.meshcentral.com

userandaname:
> oh did i not do it correctly? i created an index.html page and from what i can see, it works here https://ylianst.github.io/MeshCentral/ diverts to https://docs.meshcentral.com

yup, that redirect does work, but it'll only work for the root; anything else, e.g https://ylianst.github.io/MeshCentral/meshcentral/config, will result in a 404, which this fixes!

si458:
@userandaname thank you so much!
Ylian had a struggle to get it working, then i eventually got the pages to build,
now all moved and was just cleaning up things but i forgot about the 404 stuff!

PR #7660, opened 2026-03-04
Postgres maintenance

stephannn:
Hi,

I noticed that the variables events and statsevents for expiration had no effect on my postgres DB. The change is actually just a copy of the MariaDB setting


si458:
thank you, one of those over sights ❤️

PR #7667, opened 2026-03-06
Plugin permissions update

ryanblenis:
Adds plugin permissions framework and settings for plugin developers so that server administrators can pick and choose which users/groups have access to plugins within certain meshes or nodes (an example plugin - RegEdit - is waiting for a version to be released with this feature to be used as an example).
Permissions modal has been implemented outside the Q*() functions within common-0.0.1.js (which I believe may be affecting other plugins ability to be clicked/used within the new "ModernUI").

image

Also adds a new dropdown feature on the plugin admin page to purge the require cache and "reload" a plugins files without rebooting the server.

NOTE: Tested database functionality in NeDB and MongoDB. Other db.js updates for other database types was attempted but untested as I don't have many of these DB's available or know of people using them in production. Help testing or bug reports from users using these databases with plugins enabled are appreciated.

### DIFF
--- a/db.js
+++ b/db.js
@@ -809,6 +809,7 @@ module.exports.CreateDB = function (parent, func) {
                         CREATE TABLE power (id INTEGER PRIMARY KEY, time TIMESTAMP, nodeid CHAR(255), doc JSON);
                         CREATE TABLE smbios (id CHAR(255) PRIMARY KEY, time TIMESTAMP, expire TIMESTAMP, doc JSON);
                         CREATE TABLE plugin (id INTEGER PRIMARY KEY, doc JSON);
+                        CREATE TABLE pluginpermissions (id VARCHAR(255) PRIMARY KEY, doc JSON);
                         CREATE INDEX ndxtypedomainextra ON main (type, domain, extra);
                         CREATE INDEX ndxextra ON main (extra);
                         CREATE INDEX ndxextraex ON main (extraex);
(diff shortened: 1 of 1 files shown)


si458:
looks amazing and happy to merge!
please can you resolve the conflict
as `pluginHandler.js` looks like a complete replacement when it should only have changes in there?

ryanblenis:
Dang line endings didn't show locally. Should be fixed up now. Thank you.

stephannn:
hi @ryanblenis , @si458 ,
I think in the file default3.handlebars are two brackets in line 21543 and 21544 too much:
image

@ryanblenis , btw. tables are created now, top.

si458:
@stephannn indeed correct thank you so much! https://github.com/Ylianst/MeshCentral/commit/e76f6d54f218c88c0884c774b7a1784139cbac9d
all fixed now 👍

si458:
@ryanblenis added translations for you! https://github.com/Ylianst/MeshCentral/commit/946c1a2c23ab0ae8680879ece49440912772bd34
to do translations always make sure any text is inside `" "`
then run `cd translate && node translate extractall`
then edit translate.json and submit that file too 👍 
e.g: 
```
var abc = "Plugin Permissions"; 
var xyz = '<b>' + "Bold Text" + '</b>';
```

ryanblenis:
Thank you. I had no idea how translations worked and completely forgot about them. My bad guys. Thank you for the fixes to make sure that works

Tech-Werkz:
571353317-be55b9d6-7a87-4f8e-b127-f7abc4a00c89
tried and currently not working on MySQL Database
also tried using classic ui but still it's just loading

si458:
@Tech-Werkz the plugins permission is new!
The plugins themselves have to support the plugin permissions module
So the event log plugin (i guess based on title) will have to be updated
But @ryanblenis can confirm this for us

ryanblenis:
1. @si458 is correct that the plugins have to have support for permissions in order to use this feature;
2. @Tech-Werkz However you should not indefinitely see the "loading" in this instance. You should see a message that there are no defined permissions for that plugin, so I'll take a look and spin up a dev environment with mysql to see what's happening there

PR #7595, opened 2026-02-03
Add xterm-addon-image support for inline images (Sixel protocol)

Paulo-Borszcz:
## Feature Overview
Adds inline image support to MeshCentral's terminal using the official @xterm/addon-image addon, enabling SIXEL graphics protocol.

## What This Adds
- SIXEL graphics protocol support for terminal inline images
- Enables CLI tools like `chafa`, `lsix`, `timg` to render images at full quality instead of using Unicode block characters

## Changes
- Added `@xterm/addon-image` v0.9.0 addon files
- Updated `xterm.js` to v5.5.0 (required for addon-image compatibility - older version was missing required APIs)
- Updated `xterm-addon-fit` to v0.10.0
- Modified all 8 terminal views to load and initialize ImageAddon
- Added new addon to minification list in translate.js

### Modified Views
- xterm.handlebars
- ssh.handlebars
- default.handlebars
- default3.handlebars
- default-mobile.handlebars
- sharing.handlebars
- sharing-mobile.handlebars
- player.handlebars

## Tested
- SIXEL image rendering verified working with test sequences
- Terminal functionality unchanged for normal text operations

## Backward Compatibility
Fully backward compatible - no breaking changes. The addon only activates when SIXEL/IIP sequences are received, otherwise terminal behaves exactly as before.

Fixes #7591

Issue #7682, opened 2026-03-15
Can't connect to Terminal in v1.1.57: allowProposedApi in xterm-min.js


jesta192:
**Describe the bug**
In MeshCentral `v1.1.57`, nothing happens when I click the **Connect** button in the **Terminal** tab. In the Chrome browser Debug console I see the following error immediately after clicking **Connect**:

> Uncaught Error: You must set the allowProposedApi option to true to use proposed API

The problem does not occur in `v1.1.56`; the **Terminal** connects properly in that version.

**To Reproduce**
1. Update MeshCentral Docker image to `v1.1.57` (I used `ghcr.io/ylianst/meshcentral:latest`) and pull/deploy
2. Open MeshCentral and navigate to a Device's **Terminal** tab
3. Click the **Connect** button to initiate a **Terminal** session, and observe nothing happens
4. Open the Chrome browser Debug console and find the above error in the log

**Expected behavior**
**Terminal** should connect upon clicking the button as it did in `v1.1.56`, and no error should be logged in the console.

**Screenshots**
> <img width="957" height="407" alt="Image" src="https://github.com/user-attachments/assets/02a2bb98-7a29-4721-a487-0a3d97350349" />

**Server Software (please complete the following information):**
 - OS: Raspbian/Raspberry Pi OS "Debian GNU/Linux 11 (bullseye)"
 - Virtualization: Docker Compose
 - Network: WAN using NPM over DDNS
 - Version: 1.1.57

**Client Device (please complete the following information):**
 - Device: Laptop
 - OS: Windows 11 Home
 - Network: Remote over WAN
 - Browser: Google Chrome
 - MeshCentralRouter Version: n/a

**Remote Device (please complete the following information):**
 - Device: Raspberry Pi 4+
 - OS: Raspbian/Raspberry Pi OS "Debian GNU/Linux 11 (bullseye)"
 - Network: Remote over WAN
 - Current Core Version: Nov 21 2022, 2775716666

**Additional context**
Manually setting the image tag to `ghcr.io/ylianst/meshcentral:1.1.56` and pulling/re-deploying that previous version seems to resolve the problem. I guess it has something to do with adding images to the Terminal in #7595 but that's just a hunch?

**Your sanitized config.json file**  
```
{
  "$schema": "https://raw.githubusercontent.com/Ylianst/MeshCentral/master/meshcentral-config-schema.json",
  "settings": {
    "plugins": {
      "enabled": false
    },
[...]

PR #7683, opened 2026-03-15
Fixed regression in #7595, image-addon uses experimental apis.

cconkrig:
This PR fixes a regression where browser terminal sessions stopped opening after `xterm-addon-image` support was added in #7595.

Desktop sessions continued to work, but terminal-capable views failed immediately in the browser with:
>Uncaught Error: You must set the allowProposedApi option to true to use proposed API

That failure occurs because xterm-addon-image uses xterm proposed (experimental) APIs, while several MeshCentral terminal views were still creating terminals with:
```
new Terminal()
```
instead of:
```
new Terminal({ allowProposedApi: true })
```

This change updates the relevant browser-side terminal constructors so any view that loads xterm-addon-image creates xterm with:
```
allowProposedApi: true
```

There are really only two options here due to the regression:

- Enable allowProposedApi anywhere we load xterm-addon-image
- Revert PR #7595 and restore terminal functionality until the image addon stops depending on experimental/proposed APIs

This has been tested and has restored functionality lost after #7595 was implemented.

### DIFF
--- a/views/default-mobile.handlebars
+++ b/views/default-mobile.handlebars
@@ -5555,7 +5555,7 @@
 
                 // Setup the terminal with auto-fit
                 if (xterm != null) { xterm.dispose(); }
-                xterm = new Terminal();
+                xterm = new Terminal({ allowProposedApi: true });
                 xtermfit = new FitAddon.FitAddon();
                 xtermimage = new ImageAddon.ImageAddon();
                 if (xtermfit) { xterm.loadAddon(xtermfit); }
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -11035,7 +11035,7 @@
                         if (xterm != null) { xterm.dispose(); }
                         xtermfit = new FitAddon.FitAddon();
                         xtermimage = new ImageAddon.ImageAddon();
-                        xterm = new Terminal();
+                        xterm = new Terminal({ allowProposedApi: true });
                         if (xtermfit) { xterm.loadAddon(xtermfit); }
                         xterm.loadAddon(xtermimage);
                         xterm.open(Q('termarea3xdiv')); // termarea3x
--- a/views/default3.handlebars
+++ b/views/default3.handlebars
@@ -12035,7 +12035,7 @@
                         if (xterm != null) { xterm.dispose(); }
                         xtermfit = new FitAddon.FitAddon();
                         xtermimage = new ImageAddon.ImageAddon();
-                        xterm = new Terminal();
+                        xterm = new Terminal({ allowProposedApi: true });
                         if (xtermfit) { xterm.loadAddon(xtermfit); }
                         xterm.loadAddon(xtermimage);
                         xterm.open(Q('termarea3xdiv')); // termarea3x
--- a/views/player.handlebars
+++ b/views/player.handlebars
@@ -1104,7 +1104,7 @@
             // Setup the terminal
             if (term != null) { term.dispose(); }
             termimage = new ImageAddon.ImageAddon();
-            term = new Terminal();
+            term = new Terminal({ allowProposedApi: true });
             term.loadAddon(termimage);
             term.open(Q('XTermParent'));
             term.resize(80, 25);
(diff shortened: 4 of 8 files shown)


si458:
Can you just explain what remote os and web browser u are using?

I've yet to experience or see this problem yet

Is it only maybe certain OS types or versions?

si458:
@cconkrig is your remote machine up to date etc? as cant replicate the issue here same with another gent too
https://github.com/Ylianst/MeshCentral/issues/7682#issuecomment-4061896639

cconkrig:
- MeshCentral server is Ubuntu 22.04 LTS (updates itself each week automatically)
- Docker is not in use. 
- Guest systems are a wide variety of Windows 10 or Windows 11 and Linux is everything from Arch, Ubuntu, to Debian, none of them work because it's a server side issue with the xterm javascript. Some are updated, some are a month or two behind from what I have checked. 

I had patched my installation so I could access systems again so I had to revert to test this.
I stopped Meshcentral service, moved node_modules folder to node_modules_bak, installed Meshcentral 1.1.56. Rebooted the server, cleared browser cache, confirmed issue was not present in 1.1.56.
Ran the upgrade again in the web interface to 1.1.57 which broke terminal connections again with the same error as before.
Cleared browser cache again, tried again to open terminal, same error.

I am using Librewolf browser.

Checked the MeshCentral server and it only had a couple updates.  Ran updates which updated:
      1 linux-modules-extra-5.15.0-173-generic
      1 linux-modules-5.15.0-173-generic
      1 linux-image-5.15.0-173-generic
      1 linux-headers-5.15.0-173-generic
      1 linux-headers-5.15.0-173

Rebooted the server, and terminal is working now. I do not know why any of those packages would fix something like this given it was related to proposed apis in xterm and specifically the image add-on. Browser was last updated back on March 11th. I have had this issue for a while as I typically update pretty quick after new releases drop; just had time last night to dig into why since I was needing to access a headless remote system and couldn't get into it. That lead me down the path of digging into why this last update broke it.

I'll close this for now given it's working. But I'll leave this here so it can be indexed by search engines as I am sure others will run into this issue if they are not doing a fresh install. If it resurfaces, I have a pipeline that I can run now to patch it manually for me. I may build one that rips out the xterm image add-on if it continues to cause problems for me.

Appreciate you checking on it!

si458:
@cconkrig amazingly thats the exact same thing the other gent did! Only thing I can think is the was a bug with the kernel maybe? But very odd an upgrade and reboot fixed the issue?

cconkrig:
I had to step out for a bit when I returned, I tried to get into a system and it's come back without me doing anything. I can no longer connect to any terminal session windows or linux without the allowProposedApi option error coming up in Brave and Librewolf. I guess I didn't get the cache cleared good enough.

si458:
@cconkrig can you tried unminified? `?minify=0`
if it still happens, ill just merge the PR, doent appear to do any harm!

edit: tried it here with brave in windows, nope, cant replicate the problem
it must be something unique to your setup or the remote machines maybe?

si458:
adding as it cant do any harm!

silversword411:
Something isn't right here...this needs further investigation.

Why are my Spidey senses going off...and I'm thinking about this: https://youtu.be/aoag03mSuXQ

si458:
@silversword411 the gent is correct
https://xtermjs.org/docs/api/terminal/interfaces/iterminaloptions/#optional-allowproposedapi
The feature is off by default

cconkrig:
@si458, apologizes, I have been sick. I tested changing the config.json file to minify: false this morning and restarted the service, it did not resolve the issue even in a private browser window. I made sure it was actually downloading the files and not using local browser cache in developer tools network tab. The same error is triggered:
image


I agree with @silversword411 about the something is not right here but it is something this image plugin is trying to do. I usually not a fan of enabling experimental apis as they can break them with later updates. As far as risk profile, if someone gains access to that plugin or xterm js repo (think supply chain attack earlier), then all bets are all the table anyway, you won't need experimental apis enabled.  I wish I could figure out why it doesn't behave the same for you but I can revert and do it as long as I get my cache cleaned well enough.

si458:
@cconkrig but whats weird is i can't replicate the issue???
This is why we think it could be a browser cache issue.
But the was no harm is setting it as the default is indeed false, so turning it on isnt a big deal :)

PR #7686, opened 2026-03-16
Upn overview

stephannn:
I took Simons code and places it also in the getUserShortStr function for the general overview. The width of the user column is a bit broken , too small:
image


si458:
brilliant thank you! 
yes sadly the columns are fixed sizes at the moment
i am looking at making them resizeable in the modern ui in the future tho 👍

PR #7693, opened 2026-03-17
Plugin permission ModernUI fixes and name collision issue fix

ryanblenis:
1. Fixes an error in default3 where plugin permissions switch cases were placed under "events" versus "actions" keeping modernUI from loading the plugin permissions
2. Fix an issue where if a user or group name was a substring of another user/group name it would not add the second user correctly


si458:
Erm line diff issues again i think for default3, as no way is the 23k+ changes

ryanblenis:
> Erm line diff issues again i think for default3, as no way is the 23k+ changes

Fixed. I have no idea how that keeps happening. Especially since I did those changes in 2 commits and the first commit didn't cause it, but the second did somehow.

si458:
@ryanblenis hey don't worry about it! It always happens to me sometimes when ive pushed a commit ive then had to fix the formatting cuz stuff outta line!

PR #7721, opened 2026-03-28
UPN Column Width - List View

stephannn:
not sure if this is the correct way to wider the user column


si458:
Correct but u forgot the classic ui in default.handlebars

stephannn:
Not sure, some say so, other so ;)

Still don't know if this is the correct way

PR #7736, opened 2026-04-02
Make the "Expand All"/"Collapse All" button larger

### LINKED ISSUE #5360: Make the "Expand All"/"Collapse All" button larger
The "Expand All"/"Collapse All" functionality  - as implemented in #2593 - is a real life saver when there are lots of machines.
But the size of the button poses some real challenges, it is way too easy to miss it and click either on the "Select All" button, or the empty space on the left.
In my browser it is a 9x11 px area, that is teeny-tiny. :)

I would suggest a bigger button.
The normal size button as suggested in the original issue by nzalev is a better approach. The rationale for this is that the "Select All" button is used way less than the "Collapse All" button, so it should get at least the same amount of space if not more.
At least this is true for any user that has more than a couple dozen of machines or a dozen groups at least. For those with fewer machines/groups the screen is already as tidy as it gets.

If one would like to really save screen estate (I don't really think this is needed, but nevertheless), there could be made two buttons, the first for the "Collapse All" functionality with the current icon "[ ↕ ]", and the "Select All" button with an asterisk "[ * ]" with the correct overlay.
This way the "Collapse All" would get some additional padding so that it could be more easily clicked on, and the overall width of this area would still get smaller.

Thank you for considering this and I'm open on any discussion about this.

tylla:
Cosmetic changes in the HTML to make the "Expand all"/"Collapse all" button larger and easier to click
Fixes #5360

### DIFF
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -324,8 +324,8 @@
                     <tr>
                         <td class=h1></td>
                         <td id=devListToolbar class=style14 style="display:none">
-                            <img style="cursor:pointer;margin-top:4px;display:none" title="Collapse All" id="CollapseAllButton" src="images/icon-collapse.png" width=9 height=11 onclick="cmexpandaction(2)" />
-                            <img style="cursor:pointer;margin-top:4px;display:none" title="Expand All" id="ExpandAllButton" src="images/icon-expand.png" width=9 height=11 onclick="cmexpandaction(1)" />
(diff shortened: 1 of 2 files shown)


si458:
you forgot to edit the default3.handlebars which is the new modern ui
you cn test the new modern ui by adding `?sitetsyle=3` in the url bar

tylla:
Thanks for the reminder, it totally escaped me. I will try to find some time and change it.

tylla:
I've added the needed changes to the same PR. Albeit I'm not that confident with this one whether it is as it should be.
It definitely looks OK, and is more easy to use, but I have some doubt with the HTML. Please take a look at the code if it's acceptable with inline styles.

PR #7738, opened 2026-04-07
feat: add TLSOffload variable to docker

DaanSelen:


si458:
you forgot to add the new values into the readme.md in the docker folder :)

DaanSelen:
> you forgot to add the new values into the readme.md in the docker folder :)

Will do that today

docker/compose.yaml
```diff
@@ -1,10 +1,11 @@
 services:
   meshcentral:
-    image: ghcr.io/ylianst/meshcentral:latest
+    image: meshcentral:latest
     environment:
-      - DYNAMIC_CONFIG=false # Show the option but disable it by default, for safety.
+      - DYNAMIC_CONFIG=true # Show the option but disable it by default, for safety.
```
si458: i think the value conflicks with the comment haha

docker/compose.yaml
```diff
@@ -1,10 +1,11 @@
 services:
   meshcentral:
-    image: ghcr.io/ylianst/meshcentral:latest
+    image: meshcentral:latest
```
si458: you changed the value here but not in the readme?
can you maybe keep the ghcr.io as the default one for now
but in the readme add a comment or something to say you can now use the docker hub image
and link to the image too so people are aware?

si458 (changes requested):

docker/compose.yaml
DaanSelen: Yeah this was a mistake. Local testing change which got comitted

si458 (approved):

PR #7551, opened 2026-01-13
OIDC - Azure roles

stephannn:
The PR is to use Azure App Roles. There is also a discussion about the topic at: https://github.com/Ylianst/MeshCentral/discussions/7320

I also updated the meshcentral-config-schema and sample-config-advanced about the options fix https://github.com/Ylianst/MeshCentral/commit/be28612d1a1ace3a855632c6fa427b2380061881 

Also I am not sure if App Roles is just an Azure thing or if other provider also offer it

PR #7763, opened 2026-04-22
Fix for azure groups and roles sync

marcinl26:
Fix for azure groups and roles sync  [7551](https://github.com/Ylianst/MeshCentral/pull/7551#issuecomment-4297251949) in webserver.js

### DIFF
--- a/webserver.js
+++ b/webserver.js
@@ -8501,26 +8501,26 @@ module.exports.CreateWebServer = function (parent, db, args, certificates, doneF
                     if((Array.isArray(strategy.custom.authorities) && strategy.custom.authorities.filter(x => x.trim().length > 0).length > 0) == false || strategy.custom.authorities.includes('groups')) { 
                         getGroups(user.preset, tokenset).then((groups) => {
                             user = Object.assign(user, { 'groups': groups });
-                            //done(null, user);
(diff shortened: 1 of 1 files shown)


stephannn:
Hi @marcinl26 ,
I have a question about the PR. The idea was following:
`if((Array.isArray(strategy.custom.authorities) && strategy.custom.authorities.filter(x => x.trim().length > 0).length > 0) == false || strategy.custom.authorities.includes('groups')) { `
is used for the default or if the authorities value contains groups, so that basically it uses groups.

`if(strategy.custom.authorities.includes('roles')){`
Runs when authorities contains _roles_ . (Maybe there is an error because I didn't check for null value error in authorities)

But the idea was basically that you can have three situation
1. just groups
2. just roles
3. groups + roles

```
const strategy = {};
strategy.custom ??= {};
strategy.custom.authorities ??= [''];

if((Array.isArray(strategy.custom.authorities) && strategy.custom.authorities.filter(x => x.trim().length > 0).length > 0) == false || strategy.custom.authorities.includes('groups')) {
  console.log("groups");
}
if(strategy.custom.authorities.includes('roles')){
  console.log("roles");
}
```

PR #7775, opened 2026-04-28
Azure Ggroups and roles again

stephannn:
Hi,
after I broke the Azure Groups by my Role PR this PR https://github.com/Ylianst/MeshCentral/pull/7763 broke the roles ;)
(I never see the async functions and think they will just run line by line, my bad)

This change does following check:
1. Check if groups
1b. Check if roles also are used
2. Check roles if no groups were used before

@marcinl26 @si458 
what do you think?

### DIFF
--- a/webserver.js
+++ b/webserver.js
@@ -8500,16 +8500,11 @@ module.exports.CreateWebServer = function (parent, db, args, certificates, doneF
                 if (strategy.groups && typeof user.preset == 'string') {
                     if((Array.isArray(strategy.custom.authorities) && strategy.custom.authorities.filter(x => x.trim().length > 0).length > 0) == false || strategy.custom.authorities.includes('groups')) { 
                         getGroups(user.preset, tokenset).then((groups) => {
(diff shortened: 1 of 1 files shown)


DaanSelen (approved):

marcinl26:
@stephannn you should review your code again because here 
image
there is group object returned from getGroups function and it isn't used for user object. So user object doesn't have groups. I've checked in my test environment with debug logs (with added these debug echoes)
image

and this is what I got (redacted sensitive data):
image
So groups are missing in user object, user is authenticated but without any group assigned.
in my config.json I have set "oidc"."custom"."authorities" = "groups"
When I put "oidc"."custom"."authorities" = "groups roles" or only "roles", only roles working fine because roles are from JWT auth token, so no extra call is needed from MS Graph API.

stephannn:
Hi @marcinl26 ,
thanks for reviewing and explaining it. Really appreciate it. 
I changed the PR

Issue #5310, opened 2023-08-24
ntfy and --resetaccount


Nothing2Do:
Hello, long time, no see !
I've tryed to implement ntfy as 2 factor (as anything else I've tried don't work), and it don't work (as expected) but when I use "node ~/node_modules/meshcentral --cert `<DNS>` --domain `<DNS>` --pass `'<PW>'` --resetaccount `user//<ID>`" (stopped before and relaunched after), it still ask for a ntfy second factor.

Usually, "node --resetaccount" disable 2 factor and it don't disable ntfy ...
version used 1.1.10

Nothing2Do:
sidenote, access can be restored if you disable "messaging"  in your config.json (and use "_messaging"). but, anyway, it's a bug

si458:
just going to look into this for you, please can you share your `config.json` (hide sensitive stuff obvs)

si458:
i have been able to get it to work no problems here but it was a learning curve!?
```
  "messaging": {
    "ntfy": true
  }
```
1. login with an account without messaging setup
2. click your username at the top right hand side
3. click `manage messaging`
4. select ntfy and enter your topic you have created
5. you will be sent a verification code, enter it and click ok
6. then logout
7. now try logging back in
8. you will be greaeted with a box saying 'TOKEN' IGNORE THIS AND CLICK THE BLUE BUBBLE ICON BELOW IT
9. it will then send you a code, enter the code into a new box that appears
10. you are now logged in!

si458:
the other issue is ntfy isnt a 2fa factor, its a messaging factor!
so i dont think `--resetaccount` will work, maybe `--resetaccount` needs to also remove the messaging factors?

Nothing2Do:
I'm sorry for the waiting time. ntfy isn't 2fa, but anyway, it ask for a second factor using the ntfy icon (https://<DNS>/images/login/2fa-messaging-48.png) with the mail icon too (I suppose I can add some other 2fa).
my config.json, 
{
  "$schema": "http://info.meshcentral.com/downloads/meshcentral-config-schema.json",
  "settings": {
    "cert": "help.nothing2do.fr,hell.nothing2do.fr",
    "_WANonly": true,
    "_LANonly": true,
    "_sessionKey": "MyReallySecretPassword1",
    "port": 443,
    "_aliasPort": 442,
    "redirPort": 80,
    "_redirAliasPort": 80,
    "compression": true,
    "wscompression": true,
[...]

Issue #3455, opened 2022-01-13
--Resetaccount not clearing 2 factor


UFOTyler:
I recently enabled minimum password requirements in the config.json, this caused a forced password change for users. When a user that had 2 factor enabled tried resetting at the login screen it would crash the page. This seems to be a bug in its self, but to get around it I logged in to reset the users password and clear the 2 factor. But when they try to login it still asks for 2 factor.

So then I SSH'd in to the server, stopped it, and used: 

```
sudo node node_modules/meshcentral --resetaccount user//testuser --domain Default --pass Password123
```

It returns "Done." but when I enter those credentials it still asks for a token. From my understanding both attempts should've cleared 2 factor. 

We are using mongodb if that makes a difference. other subdomain servers that didn't have anyone using 2 factor seemed to be able to login and add two factor. I don't know why it asked for password change from certain users and not others, but it seemed to break something with resetting the accounts.

I was able to successfully get in by adding phone numbers to the accounts then choosing the phone symbol for the 2 factor code. then change the password inside the accounts. So as of now I got everyone up and running but it took some trial and error.

Thanks.

Ylianst:
Thanks for the report. Checking this now.

Also, for the "crash the page" problem, if you can go in "meshcentral-data" and look for "mesherrors.txt" and send back any errors in that file, I will take a look at it.

Ylianst:
I noticed the SMS and Device Push 2FA where not cleared when using `--resetaccount` so I just put in a fix for that. I will also note that `--resetaccount` is a last resort and requires that the server be reset when that command is done.

Hopefully that does the trick.

Ylianst:
MeshCentral v0.9.69 is out with this fix.

UFOTyler:
Hey Ylian, sorry for the delay.

I checked the mesherrors.txt, and this error seems to be relevant to the issue:

-------- 1/13/2022, 9:05:25 PM ---- 0.9.62 --------

internal/modules/cjs/loader.js:818
  throw err;
  ^

Error: Cannot find module 'wildleek'
Require stack:
- /home/mesh/meshcentral/node_modules/meshcentral/webserver.js
- /home/mesh/meshcentral/node_modules/meshcentral/meshcentral.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:815:15)
    at Function.Module._load (internal/modules/cjs/loader.js:667:27)
    at Module.require (internal/modules/cjs/loader.js:887:19)
[...]

PR #7797, opened 2026-05-16
Fix --resetaccount not clearing Duo 2FA (missing delete keyword)

adamdewolf:
Follow-up to commit 1399ce35 ("forgot 2 extra otpduo to clear!"), which added otpduo cleanup in two places. The meshuser.js side correctly used `delete chguser.otpduo;`. The meshcentral.js --resetaccount path got `user.otpduo;` instead of `delete user.otpduo;` -- the `delete` keyword was lost. As a bare expression statement it has no effect.

Effect: an administrator running --resetaccount to unlock a user that has Duo 2FA configured will see all other 2FA fields cleared, but the Duo record on the user document survives. On next login the server still requires Duo authentication, defeating the account-recovery purpose.

Related historical reports of --resetaccount not clearing specific 2FA mechanisms: #3455 (original "not clearing 2 factor"), #5310 (ntfy variant), #5331 (messaging cleanup PR).


si458:
ffs thank you! i must of been in a rush when i did the commit!
we are all only human!

PR #7788, opened 2026-05-12
Add missing index on eventids.fkid

iez0:
Fix #7787 : slow delete of old events

Issue #7787, opened 2026-05-12
Postgres maintenance : slow delete of old events due to missing index


iez0:
The commit 8d416760f713edb3488516689fc8429e3519d63d (pull request #7660) enable maintenance for postgresql, deleting old events.
**This delete is very slow due to a missing index.**
The query is : 
`DELETE FROM events WHERE time < $1`

`eventids.fkid` is a foreign key to `events.id` with `ON DELETE CASCADE`.
There is an index on `events.id`, but not on `eventids.fkid`.

Here is a small delete of 10 lines : 
```
# explain (analyse,buffers,timing) delete from events where id in (select id from events where time < '2023-01-01' LIMIT 10 );
                                                                               QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Delete on events  (cost=2.78..86.88 rows=0 width=0) (actual time=0.411..0.414 rows=0.00 loops=1)
   Buffers: shared hit=66 read=3 dirtied=3
   ->  Nested Loop  (cost=2.78..86.88 rows=10 width=34) (actual time=0.237..0.297 rows=10.00 loops=1)
         Buffers: shared hit=46 read=3
         ->  HashAggregate  (cost=2.36..2.46 rows=10 width=32) (actual time=0.185..0.193 rows=10.00 loops=1)
               Group Key: "ANY_subquery".id
               Batches: 1  Memory Usage: 32kB
               Buffers: shared hit=6 read=3
               ->  Subquery Scan on "ANY_subquery"  (cost=0.42..2.33 rows=10 width=32) (actual time=0.123..0.165 rows=10.00 loops=1)
                     Buffers: shared hit=6 read=3
                     ->  Limit  (cost=0.42..2.23 rows=10 width=4) (actual time=0.113..0.149 rows=10.00 loops=1)
                           Buffers: shared hit=6 read=3
                           ->  Index Scan using ndxeventstime on events events_1  (cost=0.42..4338.90 rows=24010 width=4) (actual time=0.111..0.144 rows=10.00 loops=1)
                                 Index Cond: ("time" < '2023-01-01 00:00:00'::timestamp without time zone)
                                 Index Searches: 1
                                 Buffers: shared hit=6 read=3
         ->  Index Scan using events_pkey on events  (cost=0.42..8.44 rows=1 width=10) (actual time=0.009..0.009 rows=1.00 loops=10)
               Index Cond: (id = "ANY_subquery".id)
[...]

Issue #7246, opened 2025-08-25
PostgreSQL extremely large


b8two:
I found my PostgreSQL extremely large at 6GB. 
I thought this was due to my configuration of 90 days for events, however there was much more than 90 days of data in the tables when I queried them.

I was able to remove the old power events quickly.
When I tried to delete events, I would get time outs and high CPU usage. I tried to reduce the time span to the last hour of data with about 100 rows but that would take almost 80 seconds to return the command.

I used an AI bot to identify the issue and the soution was to have an index on the Foreign Key.

`CREATE INDEX idx_fkid ON eventids(fkid);
`

This worked and I was able to remove the rows quickly and drop the database to under 0.8 GB

I'd like above improvement to be added for PostgreSQL and the power/event key limits (delete) code to be checked against PostgreSQL backend.

si458:
i think we already do this?
```
sqlDbExec('CREATE INDEX ndxeventstime ON events(time)', null, function (err, response) { });
sqlDbExec('CREATE INDEX ndxeventsusername ON events(domain, userid, time)', null, function (err, response) { });
sqlDbExec('CREATE INDEX ndxeventsdomainnodeidtime ON events(domain, nodeid, time)', null, function (err, response) { });
sqlDbExec('CREATE INDEX ndxeventids ON eventids(target)', null, function (err, response) { });
```

b8two:
@si458 what you have quoted only creates an index for **target** on the **eventids** table. 

I have indicated that an index is needed on **fkid** on the **eventids** table. 

It is needed on **fkid** because of the constraint created against it.
`fk_eventid   =  FOREIGN KEY (fkid) REFERENCES events(id) ON UPDATE RESTRICT ON DELETE CASCADE
`

I'm unsure if you need an index on the target column (not sure how this is used) but I doubt that this has a negative impact.
`sqlDbExec('CREATE INDEX ndxeventids ON eventids(target)', null, function (err, response) { });
`

I observed a noticable difference in performance with an index on the fkid column. 
(at least for old data removal with a large dataset.)

This was the fix I applied to resolve my problem. 
I hope this info helps anyone else that may have deployed MeshCentral with PostgreSQL.

PR #7799, opened 2026-05-16
Fix RemoveAllNodeEvents callers passing only nodeid (silent no-op cleanup)

adamdewolf:
_Loosely related (not duplicates): #7787 / #7788 / #7246 are about the time-based event-retention sweep being slow on large Postgres event tables. The per-device cleanup fixed here is a separate code path, but it contributes to the same observed symptom (event tables on SQL backends growing larger than expected) since every device removal since the bug landed has been skipping its event history._

`RemoveAllNodeEvents(domain, nodeid)` in every db.js backend starts with `if ((domain == null) || (nodeid == null)) return;` so a caller that passes only one argument leaves `nodeid` undefined and the early-return swallows the whole call. As a result, every device-removal path that should also clear the device's event history has been silently leaving those rows behind on every backend (NeDB/MongoDB TTL-expires them eventually; SQLite, PostgreSQL, MariaDB/MySQL, and AceBase keep them until manual cleanup).

Four callers fixed here:

  db.js:244           // removeInactiveDevices
  meshcentral.js:1101 // test/diagnostic-agent cleanup
  meshuser.js:2855    // user-triggered manual device removal
  meshagent.js:101    // agent self-removal on close (temporary/recovery flags)

For each, the domain string is already in scope -- `node.domain` on the node-iteration paths and `domain.id` in meshagent (parameter at line 18 of CreateMeshAgent).

- [ ] **_`A fifth caller exists at meshipkvm.js:244 with the same shape, but it is deliberately left for a follow-up. The IP-KVM auto-remove subsystem (Raritan / WebPowerSwitch, gated on the per-mesh auto-remove flag) is niche enough that I could not personally verify it and would rather not patch a path I cannot test.`_**

== How this was verified ==

Tested locally with a small harness using Node's built-in `node --test` (no new dependencies). Kept local to my fork rather than introducing a `test/` directory in MeshCentral; happy to send the harness as a separate PR if useful.

1. Arity scan: walks every .js file in the repo, finds every `RemoveAllNodeEvents(...)` call (paren-depth-aware, comment- and string-aware), and asserts each has exactly 2 args. Run on master, the test reports 5 buggy call sites. After this fix, only the intentionally-deferred meshipkvm.js:244 site remains -- the test makes that one easy to revisit.

2. Function-contract test: pins the canonical shape -- null/undefined domain or nodeid short-circuits (the F-DB-08 shape), valid pair reaches the datastore exactly once with the right query, and empty string '' (the default-domain id) is passed through (the guard is `== null`, not `!domain`, so '' is not nullish).

### DIFF
--- a/db.js
+++ b/db.js
@@ -241,7 +241,7 @@ module.exports.CreateDB = function (parent, func) {
                                     obj.Remove('si' + node._id);                          // Remove system information
                                     obj.Remove('al' + node._id);                          // Remove error log last time
                                     if (obj.RemoveSMBIOS) { obj.RemoveSMBIOS(node._id); } // Remove SMBios data
-                                    obj.RemoveAllNodeEvents(node._id);                    // Remove all events for this node
+                                    obj.RemoveAllNodeEvents(node.domain, node._id);       // Remove all events for this node
                                     obj.removeAllPowerEventsForNode(node._id);            // Remove all power events for this node
                                     if (typeof node.pmt == 'string') { obj.Remove('pmt_' + node.pmt); } // Remove Push Messaging Token
                                     obj.Get('ra' + node._id, function (err, nodes) {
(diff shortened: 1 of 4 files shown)


si458:
thank you again so much!

PR #7800, opened 2026-05-17
Software inventory optional

stephannn:
The PR is to enable or disable the software inventory feature

### DIFF
--- a/meshcentral-config-schema.json
+++ b/meshcentral-config-schema.json
@@ -1453,6 +1453,11 @@
               }
             }
           },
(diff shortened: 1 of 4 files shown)


si458:
Beating me to it as normal!
I was going to finish off things this week for this haha

stephannn:
maybe you can do it nicer by enabling it for groups separately :P

si458:
@stephannn it's all in the works!
I wanted to also merge the registry pr which has permissions and then copy from there but need the pr conflict fixing first before I can merge

Issue #5347, opened 2023-09-13
how to maximize the file download speed?


jalil1408:
meshcentral download speed of remote computer files is very slow, it is about 30KB/s with single connection, while direct download speed from the server is about 1MB/s with multiple connections.
how to force mechcentral to use all of the available speed with multiple connections?

si458:
currently this isnt possible, as the filetransfer uses websockets, so its a single connection, 
we would have to re-write the whole filetransfer stuff to do multiple uploads/downloads over websocket
sorry

SomeGuru:
> currently this isnt possible, as the filetransfer uses websockets, so its a single connection, we would have to re-write the whole filetransfer stuff to do multiple uploads/downloads over websocket sorry

Boy would it be awesome to have, or even if we could do a file check similar to those T******  things open source people love to share on. I could see multiple nodes helping to upload a file to a location given the technology could be implemented. Imagine uploading a file to agent x at 30 minutes, but then when you upload to agent p it only takes 3 minutes because agent A-I+x are helping to push the file towards agent p. 

Just some food for thought... File uploads and downloads seem to be the Achilles heel of the operations, but not a complete solution killer...

-SomeGuru

Issue #3550, opened 2022-01-27
File transfer should be improved


jacotec:
While remote desktop works like a charm, file transfer could get a bit more love in my opinion.

Compared to Teamviewer or Anydesk it's quite complicated to download a complete folder with many files from the remote PC (unless I've not overseen something). The only way is to click on every single file to save it in the browser.

So, a "total commander view" like in Teamviewer would be helpful. Or drag-and-drop from the "files" window which also covers whole folders. While I can drag files from an Explorer window into the "files" tab of the remote session, dragging files of the remote PC out of the files tab and into an Explorer window on my computer does not work.

Also I see that the file transfer is incredible slow. The MC server runs on a server in a datacenter with 1GBit/s connection. I'm connected here with 100/40 MBit. Uploading a 3MB file to "MyFiles" of the server is just a fraction of a second. Uploading the same 3MB file to a remote computer via "Files" takes around 30 seconds! And the remote PC is in the same office at the same 100/40 internet connection.

As we very often need to transfer files to and from remote computers this is a showstopper right not to replace Teamviewer.

Edit: I found that the issue with file transfer speed is related to WebRTC. I'll open a separate issue for this

Ylianst:
I have not looked at other software, but for downloading all files in a folder, you can hit the "Select All" button followed by "Zip" and have the remote agent compress all the file. You can then download that zip.

![image](https://user-images.githubusercontent.com/1319013/151319089-7298c5d5-55da-4b65-b545-90b306901ad1.png)

The "Files" panel does support drag & drop in, but not out. I think drag & drop out is only supported in Chome, I would have to check how they do it.

Your feedback on file transfer with WebRTC will be interesting, I have not tried tried to do performance tests.

jacotec:
In Teamviewer it looks like this:

![image](https://user-images.githubusercontent.com/3321655/151330173-08ad7042-1087-47bd-b024-b61dcbe782c8.png)

I'd not say to copy other designs, but this is very convenient. You can select multiple files and folders on each side and then hit "Send" or "Receive" to transfer them right from one side to the other.

Especially the ability to also browse the full dirs on the controlling computer without the need to do an "official browser download" is convenient.

Regarding the WebRTC issue I've opened #3551 . 50 seconds vs. 8 seconds for the same 2.8MB file ;-)

Ylianst:
[...]

PR #7816, opened 2026-05-20
meshdevicefile: only pause source socket when HTTP write buffer fills

oslyak:
## Problem

`meshdevicefile.js` currently issues an unconditional `pause()` on the agent WebSocket for every binary frame, then resumes from the `res.write()` callback. This is correct backpressure handling but forces stop-and-wait on every chunk — each frame must wait for the HTTP write callback before the next one can flow from the agent.

Combined with the small chunk size the agent emits and any non-trivial RTT between agent and server, this caps device file download throughput at around **2 Mbit/s on real WAN agents** regardless of available bandwidth.

This is the root cause behind recurring slow-file-transfer reports such as #5347 and #3550.

## Fix

Standard Node.js streaming pattern: only call `pause()` when `res.write()` returns `false` (the HTTP write buffer actually is full), and resume on the next `drain` event. Otherwise frames flow continuously and TCP handles overload through its own backpressure.

```js
} else {
    try {
        if (this.res.write(data) === false) {
            var sock = this._socket;
            sock.pause();
            this.res.once('drain', function () { try { sock.resume(); } catch (ex) { } });
        }
    } catch (ex) { }
}
```

## Measurement

Production deployment, 29 random Win10/Win11 agents across distinct customer networks (10 MB random file, fetched via `/devicefile.ashx`):

| Metric | Before | After | Change |
|---|---|---|---|
| Median | 250 KB/s (~2 Mbit/s) | **2.71 MB/s (~21.7 Mbit/s)** | **~11×** |
| Mean | similar | 2.53 MB/s (~20 Mbit/s) | |
| P25–P75 | – | 2.60 – 2.74 MB/s (tight) | very stable |
| Best | – | 3.40 MB/s | |
| Errors | – | 0/29 | |

A 60 MB transfer drops from ~4 minutes to ~22 seconds for affected clients.

## Notes

- Agents that emit very small WS frames with their own host-side throttling (e.g. older Windows 7 MeshAgent builds from early 2025) do not benefit as much — there the agent itself is the bottleneck.
- No changes to the agent. Server-only.
- Affects only the binary-relay branch (HTTP download of agent files); the JSON/control path is untouched.

### DIFF
--- a/meshdevicefile.js
+++ b/meshdevicefile.js
@@ -224,10 +224,18 @@ module.exports.CreateMeshDeviceFile = function (parent, ws, res, req, domain, us
                         }
                     }
                 } else {
-                    var unpause = function unpauseFunc(err) { try { unpauseFunc.s.resume(); } catch (ex) { } }
-                    unpause.s = this._socket;
-                    this._socket.pause();
-                    try { this.res.write(data, unpause); } catch (ex) { }
+                    // Only apply backpressure when the HTTP write buffer actually fills,
+                    // instead of pausing/resuming the source socket on every chunk.
+                    // The previous unconditional pause+resume per chunk forced stop-and-wait
+                    // on every frame, capping device file download throughput around 2 Mbit/s
+                    // for any client whose RTT to the server is non-trivial.
+                    try {
+                        if (this.res.write(data) === false) {
+                            var sock = this._socket;
+                            sock.pause();
+                            this.res.once('drain', function () { try { sock.resume(); } catch (ex) { } });
+                        }
+                    } catch (ex) { }
                 }
             });
 


si458:
@oslyak unfortunatly this has now broken file downloads so im going to have to revert it
this is the problem i was having in the past, everytime i changed this line
whats happening is the meshagent is just sending the file to the meshcentral server TOO QUICKLY 
and the server doesnt no what to
but its also the other way around when you upload
you send the file too quickly to meshcentral and its forwarding it to the meshagent but the meshagent doesnt no what to do with it

EDIT:
my test example was a 500mb file from 3 machines over a 1 GB leased (up/down) line remotely and meshcentral runnin gin the datacenter and my laptop only on a 300/50 fttp line
when i started the download it just stopped then failed
then clicked again and it got further then stopped then failed
and so on

EDIT2:
ok so the commit seems to work for linux devices here for me but not windows so more testing is needed

edit3: you can see from my debugging and screenshot, it does send the data but then randomly the other side just disconnects the websocket
image

PR #7824, opened 2026-05-24
Fix win-wmi module resource cleanup

PTR-inc:
While working on a fix for the win-volumes module which uses wmi, saw that the win-wmi.js module doesn't cleanup up wmi/com resources. SafeArrayAcces need a SafeArrayUnaccess to decrease the lock so it can be released. Same with the com objects, check first before release. And the wmi_handlers object gets entries added, but not deleted.
Also added a service check if wmi is running at all.

@si458 These were added to win-wmi-fixed, which is only used by a few queries. I guess this one will replace win-wmi eventually?


si458:
ideally i wanted to fully remove the `win-wmi` as it was embedded in the meshagent and didnt really need to be as it was windows only,
HOWEVER the is a few embedded modules which use the `win-wmi` module so it cant be removed
the `-fixed` version is fixed for a few issues i discovered and should be used instead

PTR-inc:
Ok, the changes I made are in de -fixed version. And I have some more updates to win-wmi-fixed, but those are more functional, will make a separate pr for that. 

- The embedded win-wmi could be replaced with the -fixed one in the polyfills. But that would make updates harder as they can only be done through an agent update.
- Replace every reference to win-wmi with win-wmi-fixed/win-wmi2. Start up with the embedded one and use the new one in the serverside modules.
- Remove the embedded win-wmi from meshagent as you said and always use the serverside one. The references in meshagent to win-wmi seem removable at first glance (use registry for the 3 lookups instead of wmi), but that would take some testing and seems more risky than the other options.

@si458 What do you think?

si458:
The is a few places that need the win-wmi module and they are in the gui if i remember? So we can't just simply remove it.
This is why i copied it to the server, called it something else and then referenced what I could with the new version

PTR-inc:
The references I found are in ilibduktape_scriptcontainer.c (os.name property), the embedded win-registry.js and embedded util-language.js. The os.name one is always triggered at agent start and thus loads the embedded win-wmi. That’s I think why it can’t be replaced with a serversside one. All the wmi queries are replaceable with registry lookups. The os.name function already does it if the wmi query fails.

But maybe I missed some. With gui you mean the installer?

But for now a replacement of all the actively used wmi queries to the -fixed version is the easiest way I guess.

si458:
U might be correct and its only that 1 place!
But yes the plan was to give from win-wmi to the -fixed version where possible as the fixed version fixed arrays and loops etc which was broken with the embedded one

Side note: I'm would ideally like to move ALL modules from being embedded into the server side then it's more flexible and quicker fixes etc but again the is a few modules that are needed within the agent itself so it's not possible as of yet

PTR-inc:
I'm already working on enhancements on the agent side, so I could look further into the windows embedded scripts if that's an option for you.

Issue #5746, opened 2024-01-28
BitLocker recovery key for offline devices


JSkolnik:
As a technical team, we would welcome a Bitlocker recovery key stored in the dashboard. The use case is for the case when the computer performs a BIOS flash incorrectly and asks the user to enter a key that he does not know, because the HW vendor has enabled encryption by default with the key stored in the TPM. Or when replacing the motherboard. These are cases where the computer is offline and it is not possible to read the key from powershell or sysinfo.
Would it be possible for the agent to load the recovery key and store it in the database, then display it in the Meshcentral dashboard? The display can be on demand after entering the password.

si458:
how does this look?
you can hover over the string and click to display a message with the key OR click the little key icon for same thing
ALSO the little key will ONLY show IF it knows a recoveryPassword, it not the key is hidden, but will still show `Fully Encrypted`
just carry on working on doing the pop up box tomorrow 👍  
![image](https://github.com/Ylianst/MeshCentral/assets/765314/00b727e8-345a-46c9-baa6-dca8b3a4786e)
![image](https://github.com/Ylianst/MeshCentral/assets/765314/04cc6ef0-a111-4773-b38e-75cf1a982cb3)

si458:
how does this look? im TERRIBLE at making GUI interfaces!?
![image](https://github.com/Ylianst/MeshCentral/assets/765314/b9c54932-96f0-4e5a-b341-b78def738d50)

JSkolnik:
Cool!
Perhaps the identifier for which the recovery code is displayed should also be displayed.
Good job!

si458:
what do you mean identifier sorry? i currently only get the drive letter, status, protectionstatus, password ?
this is the information that gets outputted for the recovery key
```
PS C:\Users\ASUS> manage-bde -protectors -get C: -Type recoverypassword
BitLocker Drive Encryption: Configuration Tool version 10.0.22621
Copyright (C) 2013 Microsoft Corporation. All rights reserved.

Volume C: [OS]
Key Protectors of Type Numerical Password

    Numerical Password:
      ID: {CC8EEA4F-9DCF-XXXX-XXXX-XXXX-XXXX-XXXX}
      Password:
        505109-477147-XXXXXX-XXXXXX-XXXXXXX-XXXXXXX-XXXX
```

JSkolnik:
It is this parameter:
Numerical Password:
[...]

PR #7827, opened 2026-05-25
Module win-volumes: fix bitlocker loop & make language independent

PTR-inc:
- DriveLetter could be null, extra check
- Now use regexp to find the id & recoverypassword, language independent


PTR-inc:
Forgot to link https://github.com/Ylianst/MeshCentral/issues/5746#issuecomment-3753545919 and credit @aelfwine88 as the inspiration for the languagefix

Issue #7819, opened 2026-05-21
FR: Disable User Shell


stephannn:
Because of the PWSH 7 request I noticed that you can also start a terminal session in the users context:

<img width="241" height="342" alt="Image" src="https://github.com/user-attachments/assets/67958e5c-3e42-46e7-a745-097ecdd0d41e" />

I have to check how we can enable/disable this feature

stephannn:
Not stale

stephannn:
not stale

PR #7828, opened 2026-05-25
Disable user shell

stephannn:
PR is about this FR: https://github.com/Ylianst/MeshCentral/issues/7819

Maybe someone feels like testing it:
image

This allows by enabling the contextMenu option in the config to define what terminal sessions can be used. All, just admin, or user. I personally think it is not a good idea that you can easily perform terminal commands in the behalf of the user, when on the other hand you must have admin rights for trouble shooting on that machine. (I know, why a scheduled task you can still execute things in the users context)

Edit:
I check the linuxShell option and it seems to be focusing on another request


stephannn:
Not stale, still able to run a terminal as user and use users tokens to do other stuff outside of the client:
image

PR #7830, opened 2026-05-26
Sanitize SVG custom icon uploads

Ndaboom:
- Clean SVG content from potentially malicious js scripts
- validate uploaded file type, size, and dimensions on the server
- limit uploaded icons to 10 MB, and limit dimension to 64 x 64 pixels
- update the UI to show the supported formats and file size limit, and dimension

### DIFF
--- a/public/js/ui-components.js
+++ b/public/js/ui-components.js
@@ -184,6 +184,9 @@ class ModernCard {
 // - `normalizePreviewUrl` for domain/path normalization
 // - `iconKey`, `label`, `currentValue` for per-instance identity and content
 // The component owns input/file/preview UI; persistence and status updates stay in page logic.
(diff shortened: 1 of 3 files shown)


si458:
We don't want to fully block svg icons as they are brilliant! 
We just want to block the JS that runs within them, check the screenshot i sent u :)

Ndaboom:
> We don't want to fully block svg icons as they are brilliant! We just want to block the JS that runs within them, check the screenshot i sent u :)

Great, applying changes

Ndaboom:
> We don't want to fully block svg icons as they are brilliant! We just want to block the JS that runs within them, check the screenshot i sent u :)

I've updated accordingly. Not using directly DOMpurify as on the screenshot to avoid installing a new dependency, but i've used instead a custom simpler function ```cleanSvg``` to clean the svg content.

PR #7837, opened 2026-05-28
fix: generate flat macOS agent package

### LINKED ISSUE #5653: macOS 14.2.1 agent not automatically starting on boot / reboot.
**Describe the bug**
Fresh install of meshcentral server and client. macOS agent installed on macOS Sonoma.

Reboot Mac and agent doesn't start. 

/Library/LaunchAgent/meshagent_osx64_LaunchAgent.plist
/Library/LaunchDaemon/meshagent_osx64_LaunchDaemon.plist

The [fix listed here](https://github.com/Ylianst/MeshCentral/issues/4822#issuecomment-1542789876) doesn't match my agent names nor does it work.

According [to this a fix has been merged](https://github.com/Ylianst/MeshCentral/issues/4849#issuecomment-1815388170) so this issue should be solved when 1.1.17 is released?

### LINKED ISSUE #6398: macOS 15 / Sequoia Installation Fails with "The installer could not install the software because there was no software found to install"
**Describe the bug**
When trying to install the agent on my Sequoia macs, I am getting the following error:
image

This is an Intel mac and I have tried both installers (x86 & Universal).  Installs on other, older macOS from this server works fine.

 - OS: Ubuntu
 - Virtualization: N/A
 - Network: AWS EC2
 - Version: [e.g. 1.0.43]
 - Node: [e.g. 18.4.0]

AJV20:
## Summary
- Replace generated macOS MeshAgent bundle `.mpkg` downloads with a flat XAR `.pkg` inside the existing ZIP archive.
- Add a cross-platform Node.js package builder so Linux/Windows MeshCentral servers can still generate customized macOS installers without `pkgbuild`/`productbuild`.
- Keep the bundled uninstall script and update the UI copy so it no longer asks users to open the legacy `.mpkg` bundle contents.
- Avoid adding the old `xar` npm package; the flat package archive is written with Node built-ins instead.

## Why
Recent macOS releases reject the old bundle-style PackageMaker `.mpkg` format with errors such as “Legacy Installer Package” / “no software found to install”. A flat XAR package is the modern installer format accepted by Installer.app.

## Test Plan
- `node --check macosinstaller.js`
- `node --check webserver.js`
- Generated a test package with `createMacOSInstaller()` and verified the output starts with `xar!`.
- `xar -tf /tmp/MeshAgent-test.pkg`
- `pkgutil --expand /tmp/MeshAgent-test.pkg /tmp/MeshAgent-test-expanded`
- `installer -pkg /tmp/MeshAgent-test.pkg -showChoicesXML`
- `pkgutil --payload-files /tmp/MeshAgent-test.pkg`
- `npm audit --omit=dev` confirms there are no `xar`/`xmldom`-related findings.

Note: this validates package structure and Installer metadata. It does not claim to fix unrelated macOS 15/26 remote desktop capture issues tracked separately.

Fixes https://github.com/Ylianst/MeshCentral/issues/6398
Part Fixes https://github.com/Ylianst/MeshCentral/issues/5653


si458:
this is great, HOWEVER is the no way to use a different package than xar and included extra ones ?
the package is 10 years ago and could have vuneribilites in
we are looking into switching from archiver to @zip.js/zip.js anyways 
could we use that instead?

AJV20:
Thanks, good call. I removed the `xar` npm dependency entirely and replaced it with a small in-tree XAR writer that only uses Node built-ins (`zlib`/`crypto`/`fs`). That avoids pulling in the old `xar` transitive dependencies while still producing the flat `.pkg` format macOS Installer expects.

I also reran the package validation locally:
- `node --check macosinstaller.js`
- `node --check webserver.js`
- generated a test `.pkg` and verified `xar -tf`
- `pkgutil --expand`
- `installer -pkg ... -showChoicesXML`
- `pkgutil --payload-files`
- `npm audit --omit=dev` no longer reports any `xar`/`xmldom`-related findings

I left the existing outer ZIP creation alone for now so this PR stays focused on fixing the macOS package format; this change should not block a later `archiver` → `@zip.js/zip.js` cleanup.

AJV20:
Thanks for the review. I checked this path: the PR is not adding the old npm `xar` package (or any external XAR dependency). `macosinstaller.js` writes the small XAR container directly because Apple flat `.pkg` files are XAR-based; tools like `productbuild` produce the same outer package format.

`@zip.js/zip.js` would be a good candidate for ZIP/archive handling elsewhere, but a ZIP container would not be accepted by macOS Installer as a flat `.pkg`. So for this specific installer output I think the outer format still needs to be XAR, while avoiding the unmaintained `xar` npm package is exactly why this is implemented inline.

I also re-ran the targeted checks after your comment:
- `node --check macosinstaller.js`
- `git diff --check origin/master...HEAD`

If you prefer, I can adjust the comments/helper names to make it clearer that this does not depend on the external `xar` package.

si458:
Thank u for the change, can u explain the other 2 packages that are in package.json and being included? 
In an ideal world we include no packages and write everything by hand!

AJV20:
Thanks — I removed the other two npm dependencies from this PR as well.

What they were doing before:
- `cpio-fs` was only used to write the gzip-compressed `newc` cpio payload archives inside the flat pkg.
- `mkbom` was only used to write the package `Bom` file.

The branch now writes the cpio payload in-tree using Node built-ins, and uses macOS' native `mkbom` when it is available. On non-macOS hosts it falls back to a zero-byte `Bom` so the package builder still has no extra runtime npm dependency.

Verification on the updated head:
- `node --check macosinstaller.js`
- `node --check webserver.js`
- generated a test pkg
- `xar -tf /tmp/MeshAgent-nodeps-nativebom.pkg`
- `pkgutil --expand /tmp/MeshAgent-nodeps-nativebom.pkg /tmp/MeshAgent-nodeps-nativebom-expanded`
- `installer -pkg /tmp/MeshAgent-nodeps-nativebom.pkg -showChoicesXML`
- `lsbom /tmp/MeshAgent-nodeps-nativebom-expanded/internal.pkg/Bom`
- `gzip -dc /tmp/MeshAgent-nodeps-nativebom-expanded/internal.pkg/Payload | cpio -itv`
- `git diff --check upstream/master...HEAD`

AJV20:
Thanks — I addressed the dependency concern in the current head.

What changed:
- No new package dependencies are added for this PR now. `package.json` and `package-lock.json` are no longer part of the PR diff.
- `macosinstaller.js` uses only Node built-ins (`fs`, `path`, `os`, `crypto`, `zlib`, `child_process`, `util`) to write the flat package pieces.
- The `archiver` / `yauzl` entries you still see in `package.json` are pre-existing MeshCentral dependencies from `master`, not dependencies introduced by this PR.

I did not switch this to `@zip.js/zip.js` because the outer macOS flat `.pkg` container is XAR, not ZIP. A ZIP writer would not produce an installable flat package. The PR now includes a small XAR writer plus gzip-compressed odc cpio payload generation directly in this file, so the macOS package path does not need extra npm packages.

Verification on the current head:
- `node --check macosinstaller.js`
- `node --check webserver.js`
- `git diff --check upstream/master...HEAD`

si458:
YOU SIR ARE AMAZING, THANK YOU SO SO MUCH ❤️ 
~~WORKS A TREAT NOW!!! ❤️~~
please can you just confirm if you used AI ?
as we ask all PRs to put a disclaimer in saying if they did

edit: spoke to soon! rebooted my mac mini and it doesnt start on start up :(
it only starts AFTER the user has logged in the first time :(

AJV20:
Thanks for the reboot report — I pushed a follow-up at `c440f33`.

The flat package now installs the system LaunchDaemon with `KeepAlive` set to true instead of only relaunching after crashes. That should keep the daemon active after a cold boot before any user logs in, while the LaunchAgent remains available for the loginwindow/KVM path.

Verification run locally:
- `node --check macosinstaller.js`
- package smoke build with `createMacOSInstaller(...)`
- extracted the generated pkg with `xar` + `cpio` and verified `/Library/LaunchDaemons/meshagent.plist` contains `<key>KeepAlive</key><true/>` and no crash-only KeepAlive dict
- `git diff --check`

si458:
@AJV20 thank you for the message!
it turns out, IF you have `FileVault` enabled on your mac (which most have now by default, CHECK!!!)
ANY application wont run on startup, UNLESS the end user unlocks their device by entering there password.
after they do this, all onstartup apps are run
i tested this with rustdesk and few other remote support apps and they are the same too, no access UNLESS you unlock the device
but remember this is ONLY ON FIRST BOOT!
and lock screen after are perfectly fine!
if you disable `FileVault` you the app starts up!

PLEASE PLEASE PLEASE can you answe my question about AI?
was this PR generated by an AI?
we the community are happy the AI overlords can help us, BUT we just like to know they did!

AJV20:
Yes — AI assistance was used via Hermes Agent, using GPT-5.5, to help draft and iterate on this PR. The final code and package behavior were reviewed and verified by a human before being pushed.

johnnyq:
This really is amazing nice work @AJV20 

Now we just need Linux Wayland Support ;-)

SinisterSpatula:
Nice work!  This has been a long time problem and so happy that you took it on. 🫶

si458:
@johnnyq @SinisterSpatula it is good news indeed!
so we fixed the installer and running on reboot,
next up to fix the remote desktop 👊

PR #7843, opened 2026-05-28
win-wmi-fixed: Add timeouts to query() to prevent blocking and errorcodes

PTR-inc:
Default timeouts are infinite, so if wmi fails to respond in any way during a query, it would hang indefinitely.
Connect to wmi timeout is 2 minutes.
Per row timeout is 10 seconds.


PTR-inc:
Added errorcode translation for common errors

PR #7848, opened 2026-05-28
Fix typo: pass -> password key in bare-metal ACM activation log entry

adamdewolf:
The bare-metal ACM activation path at amtprovisioningserver.js:482 writes its log entry under the key 'pass', while every other credential-bearing logAmtActivation caller uses 'password':
  amtmanager.js:731 (runtime password change)
  amtmanager.js:2684 (CCM activation)
  amtmanager.js:2889 (TLS-ACM activation)
  certoperations.js:120 (ACM sign request)

Because the startup loader (meshcentral.js:3800) and the in-line consumer in this same file (amtprovisioningserver.js:635) both gate on typeof <obj>.password == 'string', bare-metal entries write to disk but never enter the in-memory amtPasswords cache. The multi-password fallback in amtmanager.js:577-579 therefore has no cached candidate to try for these devices.

Forward-only: existing 'pass'-keyed entries remain skipped by the loader.

Note: Verified by Claude Code (Claude Opus 4.7, 1M context) and independent second-pass review by OpenAI Codex.

### DIFF
--- a/amtprovisioningserver.js
+++ b/amtprovisioningserver.js
@@ -479,7 +479,22 @@ module.exports.CreateAmtProvisioningServer = function (parent, config) {
 
             // Save activation data to amtactivation.log
             var domain = parent.config.domains[dev.domainid];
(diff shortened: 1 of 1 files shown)


si458:
Nothing actually changed in this pr?
It just formatted the line so its readable?
This file is ok as is

adamdewolf:
pass: dev.aquired.pass -->   password: dev.aquired.pass, (new line 494)

si458:
OMG YOU ARE SO RIGHT, THANK YOU!!! ❤️

PR #7853, opened 2026-05-29
Software inventory permissons

stephannn:
Hi,
I asked copilot to duplicate the terminal or files permissions for the software inventory. So actually it is just a copy and paste of various lines of code and a new number. It works in style 1 and 3, but currently only in English. When it is OK so far I can try to generate the language file. Last time it broke :/

I also removed the previous change with the setting by the config (https://github.com/Ylianst/MeshCentral/pull/7800).

image


si458:
this is great!
give copilot a pat on the back!
bad news, im not going to merge it because ive spotted a bug with the installedapps and installedstoreapps
which is broken so ive had to change the functions inside meshuser.js
so you will need to change all this in a few days time
SORRY!

si458:
ok so ive finished moving/fixing the software to use a new api endpoint instead,
also means in future we can now cache or store the info in the database too!
but the conflicts need fixing and permissions re-adding :)

Tech-Werkz:
> ok so ive finished moving/fixing the software to use a new api endpoint instead,
> also means in future we can now cache or store the info in the database too!
> but the conflicts need fixing and permissions re-adding :)

+1 for this, coz if device offline, software tab is unusable

si458:
@Tech-Werkz plz open up a new issue/feature request to remind me!
The codes now in place ready for the offline support, but just not sure how to store it in the database but we can talk on ur issue about it

PR #7856, opened 2026-05-29
win-wmi-fixed: Make query() and queryAsync() more robust and failsafe

PTR-inc:
Adds i(ncludesystemprops) option to the wmi command:
By default the wmi query now ignores the systemproperties (All properties starting with __). This makes the query faster and system properties are normally not needed.

Adds s(how) option to the wmi command:
Gives a progress update regarding the total rows fetched. Try f.e. 'wmi cimv2 "select * from win32_comclass'

Adds prepareQuery():
Check the fields and tries to optimize the query or the fields argument.
'select * from...' is a costly operation because enumerateProperties is run on every row. With the * it will do a GetNames to fetch all property fields, every row. If the fields argument is used, the GetNames part is skipped giving another performance boost on larger queries.

Best way at this moment to use wmi queries in modules, is to do it like this:
query='select caption,processid from win32_process';
**Try to avoid using 'select * from' in your code. Only get what you need.**

### DIFF
--- a/agents/modules_meshcore/win-wmi-fixed.js
+++ b/agents/modules_meshcore/win-wmi-fixed.js
@@ -178,7 +178,7 @@ const QueryAsyncHandler =
             cx: 12, parms: 1, name: 'Release', func: function ()
             {
                 // console.log('Release: ' + this.refcount);
-                if (--this.refcount === 0) { console.log('Released');
+                if (--this.refcount === 0) { // console.log('Released');
                     destroy(this); }
                 return GM.CreateVariable(4).increment(this.refcount >>> 0, true);   // return new refCount
             }
@@ -187,7 +187,7 @@ const QueryAsyncHandler =
             cx: 13, parms: 3, name: 'Indicate', func: function (sink, count, arr)
             {
                 // console.log('Indicate: ' + count.Val);
-                if (this._abandoned || this.results === null) { return (GM.CreateVariable(4).increment(0, true)); }   // discard
+                if (this._abandoned) { return (GM.CreateVariable(4).increment(0, true)); }   // discard
                 if (this.sessionid) {
                     this.progress += count.Val;
                     var now = Date.now();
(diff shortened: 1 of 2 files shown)


agents/modules_meshcore/win-wmi-fixed.js
```diff
@@ -165,11 +167,17 @@ const QueryAsyncHandler =
             {
                 //console.log('Indicate: ' + count.Val);
                 if (!this.results) return GM.CreateVariable(4).increment(0, true);
-
+                if (this.sessionid) {
+                    this.progress += count.Val;
+                    var now = Date.now();
+                    if ((now - this.lastProg) > 2000) {     //max 1 msg/2s, otherwise it gets throttled and possibly lose the result msg
```
si458: dont know if this matters but the console tab is actually rate limited
if it gets 10 'console' messages within 1 second, it then discards every other msg after it

PTR-inc: That's why there is a rate limit of 1 msg per 2 secs in the loop

PTR-inc:
Last addition/improvement for now:

Separate the getAllNames part from enumerateProperties and cache PropertyNames if possible to reduce Name get hammering
Extra checks
Minor:
Cleanup, replace vars with more descriptive varnames
Add partialresult to response if error in wmi command
Add resultcount and timetaken to wmi command

PTR-inc:
@si458 I will consolidate the other PRs into this one, as discussed

PTR-inc:
Now everything combined from the other PRs (and a little bit more):
- default timeout of 120 seconds on queries. Functions can provide a different timeout
- default exclude system properties
- new default for wmi command: async, pretty and show progress. Options changed to reflect this (see wmi help)
- lots op optimizations, batch result processing instead of per row, name/field caching to avoid get-name hammering
- More potential leaks removed and try's added, fail gracefully especially with regards to COM release

PTR-inc:
@si458 Tested also on 32-bit windows 7 embedded (a pain if you forget to include some tools...) but it all works. Tried some wmi queries in the console, all work, also async with the gate on max 1 async query at a time.

Screenshot with some errors which have nothing to do with this PR, but with this w7 being very bare, without defender, and a battery.

image

Device details all filled in:
image
image

PR #7861, opened 2026-05-31
Software inventory permissons2

### LINKED ISSUE #7880: Granular permission for software uninstall without full admin rights
The new Software feature currently requires full device administrator rights to uninstall applications.
This makes it difficult to delegate software management tasks in a controlled way.

Add a dedicated permission for software uninstall, e.g.:

SOFTWARE_UNINSTALL

Use Case
In larger environments, support staff should be able to:

view installed software
uninstall applications

without having full administrative control over devices.
Benefit

better separation of responsibilities
improved security
more practical use of the Software feature

stephannn:
Hi,
another PR for the Software Inventory.

It is still not working I think because of an issue with the function routeCommandToNode. I think it is currently only for remote desktop and not for software  inventory or in future of the registry?!? That why I think it is failing for the user, but not for the admin.

Maybe someone has an idea

See old one: https://github.com/Ylianst/MeshCentral/pull/7853

Fixes #7880

(Translation still missing)

Edit:
Ok, when you add
`|| (rights & MESHRIGHT_SOFTWAREINVENTORY)`
in
`if ((node != null) && (mesh != null) && ((rights & MESHRIGHT_REMOTECONTROL) || (rights & MESHRIGHT_REMOTEVIEWONLY))) { // 8 is remote control permission, 256 is desktop read only`
and
`if ((node != null) && (mesh != null) && ((rights & MESHRIGHT_REMOTECONTROL) || (rights & MESHRIGHT_REMOTEVIEWONLY))) { // 8 is remote control permission`

to the function routeCommandToNode it works, but I do not really understand the function.

Edit2:
so the function routeCommandToNode is only for remote desktop and should not be used for other operations?!?! I removed it and adapted it like in the other cases


stephannn:
Just a question, how are the "Full Device Rights" calculated in the handlebars. Now I did just 4194304 + 57592 = 4251896, but how was the 57592 calculated?

si458:
I have no clue? Permissions have always been a weird one for me to understand? But it's to do with multiples and binary I believe

stephannn:
yeah, sometimes it is explained, sometimes it is not ^^ also could not find the 57592 anywhere else in the repository

Edit:
you can get the number by summing:

8 | Remote Control
16 | console
32 | Server Files
64 | Wake
128 | Notes
8192 | Limit Events
16384 | Chat
32768 | Uninstall

so maybe other permissions are just missing?
131072: Commands
262144: Reset / Off
524288: Sharing
1048576: Details
2097152: Relay

in the calculation?

si458:
not got time today to merge this sorry, so it will be in next months release :(
also you have a merge conflict too!

stephannn:
Probably I am blind, but I don't see the conflict :/

Github says it is about the switch case 'software' when I click on Resolve conflicts, it looks good to me?!?! I also compared it in notepad++ and don't see new code in the master in the switch case 'software'

si458:
shit! i merged the new registry tab which uses the same permissions that the software one does for this!
you are going to need to re-do the permissions with a higher number now :(

PR #7872, opened 2026-06-04
add maped network drive to drive list

vboyhn:
add maped network drive to drive list:
Untitled

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -2335,6 +2335,39 @@ function getSystemInformation(func) {
                 p.then(function (res)
                 {
                     results.hardware.windows.volumes = cleanGetBitLockerVolumeInfo(res);
+					// >>> BẮT ĐẦU CODE LẤY NETWORK DRIVES <<<
+                    try {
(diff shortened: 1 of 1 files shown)


PTR-inc:
I would suggest to move this to win-volumes.js, not put this in meshcore.js as this is windows specific. 

I think it can be taking up in de drive iteration already there. So no registry access needed.

si458:
@PTR-inc correct! Also would be amazing if we can find out storage size and space left etc too!

PTR-inc:
I also have a drive/bitlocker enumeration update waiting to submit after I finish the win-wmi today, so this could be combined. It also checks for networked drive status

si458:
@PTR-inc happy for you to create PR with this in the win-info.js instead 👍

si458:
added commit here instead to win-volumes.js https://github.com/Ylianst/MeshCentral/commit/bff94407e589a3f54a2eecce21c738449be7e454 
we can look at moving it later!
also only works for local users and not domain users :(

PR #7887, opened 2026-06-12
Add extra volume and bitlocker info, change sysinfo db format

PTR-inc:
- Adds encryptionmethod (AES-128, XTS-AES-128 etc) to bitlocker info
- Adds all statuses (Locked, encryption paused etc) to bitlocker info
- Adds lastknown recoverypassword option. Useful for Bitlocker To Go drives. Config.json option: bitlockerKeyRetentionDays. Default 0. If enabled, remembers bitlocker keys for the days entered. If a rembered drive is inserted, gives admins the option to see the recoverykey even if the drive is still locked. Accessible through the same keylink in de device details pane at the bitlocker info of the device.
- Adds all drivetypes as defined by windows (RAM disk, Network drive(< only system mappings, no user atm, coming soon)) to volumeinfo
- Adds admin checks to bitlocker and sysinfo console command, only admins see recoverycodes
- Fixes csv/JSON devicereports to show bitlocker info
- Remove bitlocker info from report without device details
- Moves the volume data description logic from the agent to the webviews. The agent just pushes the raw wmi codes instead of translating it to full text. The codes are stored in the db and translated in the webviews. With sysinfo updates from the nodes, the data is updated to the code format in the sysinfo object. Previous data (from disconnected nodes) is preserved until updated and displayed as is now. There is an optional migrate script to convert the legacy string format to the new code format. Run with the server off: node meshcentral.js --migratevolumeinfo
- Fix sessionid and add error feedback to bitlocker console command
- Optimise windows-volumes() drive iterationloop

image

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -2325,6 +2325,12 @@ function getSystemInformation(func) {
         replaceSpacesWithUnderscoresRec(results);
         var hasher = require('SHA384Stream').create();
 
+        var finalizeResults = function ()
+        {
+            results.hash = hasher.syncHash(JSON.stringify(results)).toString('hex');
+            func(results);
+        };
+
         // On Windows platforms, get volume information - Needs more testing.
         if (process.platform == 'win32')
         {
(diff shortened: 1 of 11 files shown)


PTR-inc:
Added the windows volumes and bitlocker info to meshctrl.js. Bitlocker info/cache only visible if you're admin.
The recoverykeys are related to locked volumes if possible, like in the webview.

PR #7888, opened 2026-06-12
Docker: fix DYNAMIC_CONFIG writing port/redirPort as JSON strings

### LINKED ISSUE #7886: Docker: DYNAMIC_CONFIG writes settings.port/redirPort as JSON strings, server silently falls back to port 443
### Describe the bug

When running the Docker image with `DYNAMIC_CONFIG=true` and a custom `PORT`, the entrypoint writes the port into `config.json` as a JSON **string**:

```json
"port": "8080",
```

MeshCentral rejects the non-numeric value and silently falls back to the default port 443. Same issue with `REDIR_PORT`. The mismatch is hard to spot because the entrypoint logs `Setting port... 8080` while the server ends up listening on 443.

### Root cause

In `docker/entrypoint.sh`, `PORT` and `REDIR_PORT` are injected with `jq --arg` (which always produces a string) without a `tonumber` cast:

```sh
jq --arg port "$PORT" '.settings.port = $port' ...
jq --arg redirport "$REDIR_PORT" '.settings.redirPort = $redirport' ...
```

Introduced with the env-based port support in #7569. Related in spirit to #7780, which fixed another jq quoting issue in the same script (postgres/mariadb section) but did not touch the port type.

### To Reproduce

```sh
docker run -e DYNAMIC_CONFIG=true -e PORT=8080 -e HOSTNAME=test.local meshcentral/meshcentral:1.1.59-mongodb-debian
```

Log shows `MeshCentral HTTPS server running on test.local:443.` and `config.json` contains `"port": "8080"`.

### Expected behavior

Server listens on 8080.

### Fix

Cast in the jq filter — verified working on 1.1.59:

```sh
'.settings.port = ($port|tonumber)'
'.settings.redirPort = ($redirport|tonumber)'
```

(or use `--argjson`). Still present on `master` and `1.2.0` as of today. Happy to open a PR if useful.

### Server Software / Version

Official Docker image `1.1.59-mongodb-debian`; affects any Docker/Kubernetes setup using `DYNAMIC_CONFIG` with a port other than 443.

duvoux:
Fixes #7886.

`jq --arg` always produces a JSON string, so with `DYNAMIC_CONFIG=true` the entrypoint wrote `"port": "8080"` into `config.json`. MeshCentral requires a number for `settings.port`/`settings.redirPort`:

```js
if (obj.args.port == null || typeof obj.args.port != 'number') { obj.args.port = 443; }
```

so the server silently fell back to 443/80 while the entrypoint logged the requested port.

This casts the values in the jq filter (`($port|tonumber)`), including the 443/80 default branches, matching the `--argjson` style already used for the boolean settings in the same script.

Tested on the generated config:
- `PORT=8080 REDIR_PORT=800` → `"port": 8080, "redirPort": 800` (numbers, pass the `typeof === 'number'` check);
- default branch → `"port": 443` (number);
- side effect of the cast: a non-numeric `PORT` now makes jq fail and leaves `config.json` untouched (the `&& mv` guard), instead of silently writing a string the server rejects.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -127,12 +127,12 @@ function dynamic_config() {
         echo "Setting port... $PORT"
 
         jq --arg port "$PORT" \
-            '.settings.port = $port' \
+            '.settings.port = ($port|tonumber)' \
             "$CONFIG_FILE" > temp_config.json && mv temp_config.json "$CONFIG_FILE"
     else
         echo "Invalid or no port, defaulting to '443', Value(s) given: ${PORT:-empty}"
         jq --arg port "443" \
-            '.settings.port = $port' \
+            '.settings.port = ($port|tonumber)' \
             "$CONFIG_FILE" > temp_config.json && mv temp_config.json "$CONFIG_FILE"
     fi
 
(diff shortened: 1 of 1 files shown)


si458:
thank you for pr!
im happy with the side effect as everyone who knows docker should know a number is not a string anyways!

Issue #7885, opened 2026-06-11
bootstrap requires access to fonts.googleapis.com


stephannn:
Hi,

noticed today in our restricted environment that the bootstraps:
_**/styles/themes/flatly/bootstrap-min.css**_
requires access to:
**_https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,400;0,700;1,400&display=swap_**

In our environment the internet access is limited and results in a 60 seconds waiting period.

This is a reminder to look into it, or maybe someone has already a solution.

si458:
sadly google want you to load the fonts from their server because they are dynamically generated

what we would need to do instead is edit every single bootstrap.css and bootstrap-min.css inside the themes folder
and replace the loading of the fonts from google and instead load them locally
but the downside would be it means we have to upload every single font thats uses into meshcentrals public repo
and if you check the url yourself, just the 1 file u mensioned above downloads 6 different font styles in woff2 format
6 x 27 = 162 possible different fonts

@Ndaboom @kambereBr can you have a look at this for me please?
see if the bootswatch themes can use local fonts instead of google ones

kambereBr:
> sadly google want you to load the fonts from their server because they are dynamically generated
> 
> what we would need to do instead is edit every single bootstrap.css and bootstrap-min.css inside the themes folder and replace the loading of the fonts from google and instead load them locally but the downside would be it means we have to upload every single font thats uses into meshcentrals public repo and if you check the url yourself, just the 1 file u mensioned above downloads 6 different font styles in woff2 format 6 x 27 = 162 possible different fonts
> 
> [@Ndaboom](https://github.com/Ndaboom) [@kambereBr](https://github.com/kambereBr) can you have a look at this for me please? see if the bootswatch themes can use local fonts instead of google ones

@si458 Sure, I'll take a look.

stephannn:
Hi @kambereBr ,
I just took a look and the official bootstrap css does not have google fonts included, but the theme ones like darkly, cosmo and cyborg do. There was also a discussion about it, some years ago but without a further result:
https://github.com/thomaspark/bootswatch/issues/1142

Also chatgpt has no real clue, except change the DNS, modify the CSS after each update, etc :D

Thanks for looking into it @kambereBr

si458:
@stephannn sadly your PR wont work because every theme has an import and each import is a different font, so we would need to download every single font from every single theme

[...]

PR #7892, opened 2026-06-13
Google Fonts - bootstrap

stephannn:
Hi,
I found another project with the same issue (https://github.com/Ylianst/MeshCentral/issues/7885):
https://github.com/epoupon/lms/commit/70e847ce3626a1cfd29acdc01f258e3175bfefa5

And because the bootstrap fork does not seem to get it fixed by the owner, I played around with the AI a bit to get the google API excluded from the bootstrap files. I did a test on our server and afterwards it loads normally and  I can see that the fonts getting loaded locally.

Not sure if this is a good idea, but how can you include https://fonts.googleapis.com/ into your CSS?!? :O

No problem when rejecting the PR ;)


si458:
this PR wont work
image
every theme has its own import from google fonts and is a different style
so we would need to include EVERY SINGLE FONT from ALL the themes

stephannn:
arghh, I found only three with the identical font, but did not expand my search filter.

I will try to find another way

PR #7896, opened 2026-06-14
Fixup sqlite setup

PTR-inc:
- Fix async chaining of sqliteSetOptions and setupFunctions
- Strengthen sqlite options check
- Fix improperly formatted pragma option incremental_vacuum
- Optimise pragma function to only get pragmas if debug is really enabled


si458:
sounds stupid, have u verified this is ok with existing sqlite setups?

PTR-inc:
Of course, all my mc’s run on sqlite

-------- Original Message --------

si458:
Thats all the proof I need haha

Issue #7436, opened 2025-11-18
Show the main IP address of a client


tylla:
**Getting the IP address of a client is really cumbersome**
Currently getting the IP address of a client is really painful, one has to comb through the Details tab's contents to find it.

**The IP address should be shown on the General tab**
The IP address(es) of a machine should be shown on the General tab either under the "Agent connected" label, or in the normal list somewhere around the "OS Name". It could be combined with the public IP address (the one that gets listed in the Devices list).
Something like "IP address: 192.168.1.132 (via 142.251.208.133)" should be displayed.

**Additional context**
There can be multiple IP addresses associated with a host (especially if it has IPv6 automatic addresses), but normally only the one with a default gateway set should be considered.

I'm really happy to answer/clear up any questions.

si458:
the problem you have with this, is what about devices with multiple interfaces and mutliple IP addresses?
how do we know which IP address it used?
u cant just assume it connected via an interface with default gateway ip
example,
meshcentral running on 192.168.88.99,
interface 1: 192.168.88.100/255.255.255.0 NO GATEWAY
interface 2: 192.168.66.22/255.255.255.0 GATEWAY 192.168.66.254
meshcentral would connect using interface 1
but this is more confusing when you use external IP addresses

other example, netbird or tailscale, meshcentral uses VPN interfaces (i have seen this with my own eyes from clients)

tylla:
Hmmm, you're right, the world is always way more complicated (and confusing) than one would assume... ;-)

I would guess as the simplest solution the "single address with the gateway" should handle most cases. (most = I wouldn't be brave enough to try to say a percentage)
For the more complex cases I could imagine the simple solution extended with a hover info bubble that holds all the addresses.

So something like this:
 ```
                        .------------------------------,
                        | 192.168.88.100 192.168.55.22 |
                        `----    ----------------------'
                              \/
IP address: 192.68.55.22 (via 192.168.88.100)
```
Am I overstretching my fantasy (and the possibilities/goodwill of the system/devs?

si458:
the other problem you have is if you use AMT devices or the local groups
the hostname option is ALREADY the devices local ip address as you need to know this to talk to AMT or a local device
also with AMT CIRA the hostname actually gets converted into a reverse dns
so in my case it shows the hostname as `123-123-123-123.static.dsl.as9105.com`

tylla:
[...]

PR #7898, opened 2026-06-15
Implement getProbableMainLanIp function

panreyes:
Implemented getProbableMainLanIp function in meshcore.js, related to #7436.

It still requires changes in default3.handlebars to display this field in the device list and in general.

Also escaped a few more strings.


si458:
you forgot about `default-mobile.handlebars` for EscapeHtml as this is the classic mobile ui 👍

panreyes:
Thanks! I added some EscapeHtmls there too, but I haven't tested it.

si458:
Alt text

panreyes:
I just tested it as far as I could (only with Windows devices) and it seems to work just fine.
I will probably review the 3 main handlebars in the near future to make completely sure there are no other fields that are not being escaped.

PR #7899, opened 2026-06-15
Remove unused listconfigfiles queries

PTR-inc:
Came across the stateless config management functions and saw this function is not used, as it was probably replaced with the default getalltype query it uses now a long time ago.
@si458 And optional help info regarding the commands to run meshcentral stateless. I placed it in a separate commit so you can decide if you want to include it or not.


PTR-inc:
..And forgot the output fix, it was late

PR #7904, opened 2026-06-16
Prevent linux-info.js from hanging the agent

PTR-inc:
Discovered this bug after the meshagent on one of my alpine servers hung. A trace showed it stopped in linux-info.js, at the clamav check. This server happened to have a disabled clamav package installed. The clamav processcheck hung and caused a 'timers.onElapsed() callback handler on '()' => waitExit() already in progress' exception in the console, with an unresponsive/hung agent as a result.

The child_process.execFile/waitExit() structure should be avoided if possible, as explained by Claude below. Originally I just added a timeout in the waitExit(4000), but the claude check brought up the bigger issue.
waitExit() is used quite a few times throughout the meshcore modules, so I think there is a todo to at least add the timeout and review and, where possible, remove them by making it async.

All functions in the linux-info.js, including the software inventory are async now with timeouts and there is a guard added in the 'sendPeriodicServerUpdate' function in meshcore.js to prevent overlapping runs, just to be safe.

**A bit of Claude:**
## The setup: one thread, one event loop

MeshAgent's JS runtime (duktape embedded in the native ILib "chain"/microstack) is **single-threaded with a single event loop**. That loop does roughly:

```
forever:
   poll all file descriptors (sockets, child stdout/stderr pipes, timers)
   for each thing that's ready: call its JS callback
```

Child-process I/O and exit notifications are delivered *through this loop* — the child's stdout arrives because the loop polled the pipe fd, and "the child exited" is detected the same way.

## The problem `waitExit()` solves

`child.waitExit()` offers **synchronous semantics**: the next line of JS doesn't run until the child is done, and `child.stdout.str` is fully populated.

```js
child.stdin.write('...\nexit\n');
child.waitExit();                 // <-- does not return until child exits
var out = child.stdout.str;       // <-- guaranteed complete here
```

But here's the catch: it **can't** implement that by just parking the thread (a plain `sleep`/wait). If the thread stops, the event loop stops, the child's pipe never gets read, the exit never gets noticed → permanent deadlock. The very mechanism that would tell you "the child exited" is the loop you just froze.

## The trick: a *nested* event loop

So `waitExit()` doesn't freeze the loop — it **runs another copy of the loop, in place**, on top of the current call stack:

```
waitExit():
   forever:
      poll fds            <-- same polling the top-level loop does
      dispatch ready callbacks   <-- timers fire, pipe data is read, etc.
      if (child has exited) return
      if (timeout elapsed) kill + return
```

This is the "nested / recursive event loop" technique (same idea as a browser's synchronous `alert()`, or `deasync` in Node). The actual implementation is in the native agent binary, not the JS you can see.

## In what sense it "blocks"

Two senses, and they're different:

1. **It blocks the calling code's stack.** Control is parked inside the `waitExit()` frame — which is itself buried under whatever triggered the current dispatch (a timer callback, a server command, etc.). The **top-level loop iteration cannot unwind and advance** until `waitExit()` returns. Anything waiting for the stack to return to the top is stalled.

2. **It does *not* block dispatch.** The nested loop is still pumping — so timers still fire, other pipes still get serviced, callbacks still run. The loop is "blocked" at the top level but "alive" inside the nest.

That second property is the dangerous one.

## Why that produced your bug

```
top-level loop
  └─ fires timer A  →  callback collects AV info
        └─ child.waitExit()        ← nested loop starts, pumping
              └─ (8s pass; command hung)
              └─ timer B's deadline passes → nested loop dispatches it
                    └─ timer B's callback calls child.waitExit()  ← SECOND one
                          └─ guard trips: "waitExit() already in progress"  ✗
```

A timer fired *inside* the nested pump, and its callback tried to start a **second** nested `waitExit()`. MeshAgent refuses to stack them (unbounded nesting would corrupt the chain's re-entrant state / risk the C stack), so it throws — and that exception surfaces as `timers.onElapsed() callback handler on '()' => waitExit() already in progress`.

And the *other* failure mode you saw — "the agent doesn't continue" — is the degenerate case: if the child **never exits and there's no timeout**, the nested loop spins forever. The stack never unwinds, so the top-level loop never resumes. The agent is wedged inside `waitExit()`.

## Why making it async fixes it completely

The async version (`runShellAsync`) never opens a nested loop:

```js
child.on('exit', function () { cb(...); });   // delivered by the TOP-level loop
child._to = setTimeout(...);                    // also a normal top-level event
// function returns immediately — stack unwinds
```

The `exit` and timeout are just ordinary events handled by the **one** top-level loop. The call stack unwinds right away, so there's never a second loop to nest into, and a hung command merely delays a callback instead of parking the whole stack. That's why it's immune regardless of which command hangs or which timer races.

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -7040,88 +7040,101 @@ function sendNetworkUpdate(force) {
 }
 
 // Called periodically to check if we need to send updates to the server
+var sendPeriodicServerUpdateInProgress = false;
 function sendPeriodicServerUpdate(flags, force) {
     if (meshServerConnectionState == 0) return; // Not connected to server, do nothing.
-    if (!flags) { flags = 0xFFFFFFFF; }
-    if (!force) { force = false; }
 
-    // If we have a connected MEI, get Intel ME information
-    if ((flags & 1) && (amt != null) && (amt.state == 2)) {
-        delete meshCoreObj.intelamt;
-        amt.getMeiState(9, function (meinfo) {
-            meshCoreObj.intelamt = meinfo;
-            meshCoreObj.intelamt.microlms = amt.lmsstate;
-            meshCoreObjChanged();
-        });
-    }
+    // Defensive guard against re-entrancy. The info collectors below are currently
+    // non-blocking (Linux av()/firewall() are async; Windows uses native WMI), so
+    // nothing here pumps the event loop and this guard never actually fires. It's
+    // kept as a safety net: if a future collector re-introduces a blocking
+    // child.waitExit() (which runs a nested event loop), a timer firing inside it
+    // could re-enter and throw 'waitExit() already in progress'. Skipping the
+    // overlapping run avoids that.
+    if (sendPeriodicServerUpdateInProgress) return;
+    sendPeriodicServerUpdateInProgress = true;
+    try {
+        if (!flags) { flags = 0xFFFFFFFF; }
+        if (!force) { force = false; }
 
-    // Update network information
-    if (flags & 2) { sendNetworkUpdateNagle(false); }
-
-    // Update anti-virus information
-    if ((flags & 4) && (process.platform == 'win32')) {
-        // Windows Command: "wmic /Namespace:\\root\SecurityCenter2 Path AntiVirusProduct get /FORMAT:CSV"
-        try { meshCoreObj.av = require('win-info').av(); meshCoreObjChanged(); } catch (ex) { av = null; } // Antivirus
-        //if (process.platform == 'win32') { try { meshCoreObj.pr = require('win-info').pendingReboot(); meshCoreObjChanged(); } catch (ex) { meshCoreObj.pr = null; } } // Pending reboot
-    }
-    // Update Linux AV/Firewall information. av()/firewall() are asynchronous (they
-    // spawn child processes), so they don't block this periodic update with a nested
-    // waitExit() loop. Results are applied in the callbacks and flushed via
-    // meshCoreObjChanged().
-    if ((flags & 4) && (process.platform == 'linux')) {
-        try {
-            require('linux-info').av(function (avResult) {
-                try {
-                    var lsc = {};
-                    if (avResult && avResult.length > 0) { meshCoreObj.av = avResult; lsc.antiVirus = 'OK'; }
-                    require('linux-info').firewall(function (fwResult) {
-                        try {
-                            if (fwResult && fwResult.installed) { lsc.firewall = fwResult.enabled ? 'OK' : 'BAD'; }
-                            if (Object.keys(lsc).length > 0) { meshCoreObj.lsc = lsc; meshCoreObjChanged(); }
-                        } catch (ex) { }
-                    });
-                } catch (ex) { }
+        // If we have a connected MEI, get Intel ME information
+        if ((flags & 1) && (amt != null) && (amt.state == 2)) {
+            delete meshCoreObj.intelamt;
+            amt.getMeiState(9, function (meinfo) {
+                meshCoreObj.intelamt = meinfo;
+                meshCoreObj.intelamt.microlms = amt.lmsstate;
+                meshCoreObjChanged();
             });
-        } catch (ex) { }
-    }
+        }
 
-    if (process.platform == 'win32') {
-        if (require('MeshAgent')._securitycenter == null) {
+        // Update network information
+        if (flags & 2) { sendNetworkUpdateNagle(false); }
+
+        // Update anti-virus information
+        if ((flags & 4) && (process.platform == 'win32')) {
+            // Windows Command: "wmic /Namespace:\\root\SecurityCenter2 Path AntiVirusProduct get /FORMAT:CSV"
+            try { meshCoreObj.av = require('win-info').av(); meshCoreObjChanged(); } catch (ex) { av = null; } // Antivirus
+            //if (process.platform == 'win32') { try { meshCoreObj.pr = require('win-info').pendingReboot(); meshCoreObjChanged(); } catch (ex) { meshCoreObj.pr = null; } } // Pending reboot
+        }
+        // Update Linux AV/Firewall information. av()/firewall() are asynchronous (they
+        // spawn child processes), so they don't block this periodic update with a nested
+        // waitExit() loop. Results are applied in the callbacks and flushed via
+        // meshCoreObjChanged().
+        if ((flags & 4) && (process.platform == 'linux')) {
             try {
-                require('MeshAgent')._securitycenter = require('win-securitycenter').status();
-                meshCoreObj['wsc'] = require('MeshAgent')._securitycenter; // Windows Security Central (WSC)
-                require('win-securitycenter').on('changed', function () {
-                    require('MeshAgent')._securitycenter = require('win-securitycenter').status();
-                    meshCoreObj['wsc'] = require('MeshAgent')._securitycenter; // Windows Security Central (WSC)
-                    require('MeshAgent').SendCommand({ action: 'coreinfo', wsc: require('MeshAgent')._securitycenter });
+                require('linux-info').av(function (avResult) {
+                    try {
+                        var lsc = {};
+                        if (avResult && avResult.length > 0) { meshCoreObj.av = avResult; lsc.antiVirus = 'OK'; }
+                        require('linux-info').firewall(function (fwResult) {
+                            try {
+                                if (fwResult && fwResult.installed) { lsc.firewall = fwResult.enabled ? 'OK' : 'BAD'; }
+                                if (Object.keys(lsc).length > 0) { meshCoreObj.lsc = lsc; meshCoreObjChanged(); }
+                            } catch (ex) { }
+                        });
+                    } catch (ex) { }
                 });
             } catch (ex) { }
         }
 
-        // Get Defender Information
-        try {
-            meshCoreObj.defender = require('win-info').defender();
-            meshCoreObjChanged();
-        } catch (ex) { }
+        if (process.platform == 'win32') {
+            if (require('MeshAgent')._securitycenter == null) {
+                try {
+                    require('MeshAgent')._securitycenter = require('win-securitycenter').status();
+                    meshCoreObj['wsc'] = require('MeshAgent')._securitycenter; // Windows Security Central (WSC)
+                    require('win-securitycenter').on('changed', function () {
+                        require('MeshAgent')._securitycenter = require('win-securitycenter').status();
+                        meshCoreObj['wsc'] = require('MeshAgent')._securitycenter; // Windows Security Central (WSC)
+                        require('MeshAgent').SendCommand({ action: 'coreinfo', wsc: require('MeshAgent')._securitycenter });
+                    });
+                } catch (ex) { }
+            }
 
-        // Calculate Windows Idle Time
-        try {
-            require('win-deskutils').idle.getSecondsAllSessions().then(function (seconds) {
-                meshCoreObj.idletime = seconds;
+            // Get Defender Information
+            try {
+                meshCoreObj.defender = require('win-info').defender();
                 meshCoreObjChanged();
-            });
-        } catch (ex) { sendConsoleText('Error getting idle time: ' + ex.toString());}
-    }
+            } catch (ex) { }
 
-    // Send available data right now
-    if (force) {
-        meshCoreObj = sortObjRec(meshCoreObj);
-        var x = JSON.stringify(meshCoreObj);
-        if (x != LastPeriodicServerUpdate) {
-            LastPeriodicServerUpdate = x;
-            mesh.SendCommand(meshCoreObj);
+            // Calculate Windows Idle Time
+            try {
+                require('win-deskutils').idle.getSecondsAllSessions().then(function (seconds) {
+                    meshCoreObj.idletime = seconds;
+                    meshCoreObjChanged();
+                });
+            } catch (ex) { sendConsoleText('Error getting idle time: ' + ex.toString());}
         }
-    }
+
+        // Send available data right now
+        if (force) {
+            meshCoreObj = sortObjRec(meshCoreObj);
+            var x = JSON.stringify(meshCoreObj);
+            if (x != LastPeriodicServerUpdate) {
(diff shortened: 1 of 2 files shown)


PTR-inc:
Verified through a test that the options 'timeout' & 'cwd' aren't implemented in `ILibDuktape_ChildProcess_execFile` even though it is documented in the source. A setTimeout on the childprocess is the way to go:

```
var child = require('child_process').execFile(file, args, options);
var t = setTimeout(function () { child.kill(); }, 5000);
child.on('exit', function () { clearTimeout(t); });
```
There are more options available: type, uid, detached and env. Undocumented, but useful.

### Summary by the C-boy:

<html>
<body>
<p>Here's what each pre-existing option does, traced from the <code>execFile</code> parser (<a href="https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L519-L563">ILibDuktape_ChildProcess.c:519–563</a>) down into <code>SpawnProcessEx4</code>.</p>
<h2><code>type</code> — <em>how</em> and <em>in what context</em> the process is spawned</h2>
<p><a href="https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L522">Line 522</a>: read straight into <code>spawnType</code> (default <code>0</code>). It's the <code>ILibProcessPipe_SpawnTypes</code> enum (<a href="https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.h#L36">ILibProcessPipe.h:36</a>):</p>

type | Meaning
-- | --
0 DEFAULT | Normal child in the agent's own security context. Windows CreateProcessW; POSIX vfork+execv (real fork on macOS).
1 USER | (Windows) run as the logged-in console user — duplicates the process token, retargets it to the active console session.
2 WINLOGON | (Windows) run on the secure Winsta0\Winlogon desktop (ILibProcessPipe.c:625).
3 TERM | (POSIX) allocate a PTY via forkpty — gives the child a pseudo-terminal (used for interactive shells; exposes pty/tcsetsize).
4 DETACHED | Spawn with no stdio pipes — no stdin/stdout/stderr capture (fire-and-forget). This is what makes IsDetached() true.
5 SPECIFIED_USER | (Windows) run in a specific session (id taken from uid).


<h2><code>uid</code> — means different things per platform</h2>
<p><a href="https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L523">Line 523</a>, default <code>-1</code>:</p>
<ul>
<li><strong>POSIX</strong>: the numeric user id to drop to. Passed as <code>sid</code> → in the fork-child, if <code>UID != -1 &amp;&amp; != 0</code> it calls <code>setuid(UID)</code> (<a href="https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L924">ILibProcessPipe.c:924</a>) so the child runs as that user.</li>
<li><strong>Windows</strong>: not a POSIX uid — it's a <strong>session id</strong>. If <code>uid &gt;= 0</code> <em>and</em> <code>type == USER</code>, the type is upgraded to <code>SPECIFIED_USER</code> (<a href="https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L525">line 525</a>) and <code>uid</code> becomes the target session for the token (<a href="https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L623">ILibProcessPipe.c:623</a>). With other types it's effectively ignored (passed as <code>0</code>).</li>
</ul>
<h2><code>detached</code> — new session / process group (NOT the same as <code>type:4</code>)</h2>
<p><a href="https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L527">Line 527</a>: if truthy, OR-s the <code>POSIX_DETACHED</code> (<code>0x8000</code>) <strong>bitflag</strong> onto <code>spawnType</code>. In the spawn it sets <code>needSetSid</code>, which:</p>
<ul>
<li><strong>POSIX</strong>: calls <code>setsid()</code> in the child (<a href="https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L928">ILibProcessPipe.c:928</a>) → new session, decoupled from the parent's controlling terminal/process group, survives the parent.</li>
<li><strong>Windows</strong>: adds <code>DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP</code> to the creation flags.</li>
</ul>
<p>⚠️ <strong>Gotcha — <code>detached:true</code> ≠ <code>type:4 (DETACHED)</code>:</strong></p>
<ul>
<li><code>detached:true</code> still wires up stdio pipes (you keep getting <code>stdout</code>/<code>stderr</code>); it only changes the session/process-group.</li>
<li><code>type:4</code> creates <strong>no</strong> pipes at all.
They're orthogonal and can combine.</li>
</ul>
<h2><code>env</code> — child environment (full replacement, and overloaded for TERM)</h2>
<p><a href="https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L528-L562">Lines 528–562</a>: enumerates the <code>env</code> object's own keys into a <code>key,value,...</code> array → passed as <code>envvars</code>. Windows builds a <code>CREATE_UNICODE_ENVIRONMENT</code> block; POSIX builds <code>KEY=VAL</code> strings for <code>execve</code>. If <code>env</code> is omitted, <code>envargs</code> stays <code>NULL</code> and the child <strong>inherits the agent's environment</strong>.</p>
<p>Two important nuances:</p>
<ol>
<li><strong>It's a full replacement, not a merge.</strong> Unlike the <code>execve</code> binding (which explicitly appends <code>process.env</code>, <a href="https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L632-L633">lines 632–633</a>), <code>execFile</code> passes <em>only</em> the keys you provide. <code>{ env: { FOO: 'bar' } }</code> gives the child essentially just <code>FOO</code>.</li>
<li><strong>For <code>type:3</code> (TERM), <code>env</code> is reinterpreted as terminal config, not environment.</strong> In the TERM branch the same buffer is parsed for <code>LINES</code>, <code>COLUMNS</code>, <code>c_iflag/c_oflag/c_cflag/c_lflag</code> (<a href="https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L772-L819">ILibProcessPipe.c:772–819</a>) to size the PTY and set termios flags — and <code>TERM=xterm-256color</code> is force-injected (<a href="https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L535">line 535</a> and again in the child).</li>
</ol>
<hr>
<p>So the four split into: <strong><code>type</code></strong> (spawn mechanism/context), <strong><code>uid</code></strong> (POSIX setuid / Windows session), <strong><code>detached</code></strong> (session-detach flag, distinct from <code>type:4</code>), and <strong><code>env</code></strong> (replacement environment, repurposed as PTY/termios settings under <code>type:3</code>).</p>
<p>One caveat on the links: the line numbers come from your <strong>local</strong> <code>master</code>. If the GitHub remote's <code>master</code> has drifted from your local copy, the anchors could be off by a few lines — worth a quick check against the rendered files if precision matters.</p></body></html>
</body>
</html>Here's the answer with every reference now linking to the files on GitHub `master` (https://github.com/Ylianst/MeshAgent).

Here's what each pre-existing option does, traced from the `execFile` parser ([[ILibDuktape_ChildProcess.c:519–563](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L519-L563)](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L519-L563)) down into `SpawnProcessEx4`.

## `type` — *how* and *in what context* the process is spawned
[[Line 522](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L522)](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L522): read straight into `spawnType` (default `0`). It's the `ILibProcessPipe_SpawnTypes` enum ([[ILibProcessPipe.h:36](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.h#L36)](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.h#L36)):

| `type` | Meaning |
|---|---|
| `0` DEFAULT | Normal child in the agent's own security context. Windows `CreateProcessW`; POSIX `vfork`+`execv` (real `fork` on macOS). |
| `1` USER | (Windows) run as the **logged-in console user** — duplicates the process token, retargets it to the active console session. |
| `2` WINLOGON | (Windows) run on the secure `Winsta0\Winlogon` desktop ([[ILibProcessPipe.c:625](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L625)](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L625)). |
| `3` TERM | (POSIX) allocate a **PTY** via `forkpty` — gives the child a pseudo-terminal (used for interactive shells; exposes `pty`/`tcsetsize`). |
| `4` DETACHED | Spawn with **no stdio pipes** — no stdin/stdout/stderr capture (fire-and-forget). This is what makes `IsDetached()` true. |
| `5` SPECIFIED_USER | (Windows) run in a **specific session** (id taken from `uid`). |

## `uid` — means different things per platform
[[Line 523](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L523)](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L523), default `-1`:
- **POSIX**: the numeric user id to drop to. Passed as `sid` → in the fork-child, if `UID != -1 && != 0` it calls `setuid(UID)` ([[ILibProcessPipe.c:924](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L924)](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L924)) so the child runs as that user.
- **Windows**: not a POSIX uid — it's a **session id**. If `uid >= 0` *and* `type == USER`, the type is upgraded to `SPECIFIED_USER` ([[line 525](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L525)](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L525)) and `uid` becomes the target session for the token ([[ILibProcessPipe.c:623](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L623)](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L623)). With other types it's effectively ignored (passed as `0`).

## `detached` — new session / process group (NOT the same as `type:4`)
[[Line 527](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L527)](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L527): if truthy, OR-s the `POSIX_DETACHED` (`0x8000`) **bitflag** onto `spawnType`. In the spawn it sets `needSetSid`, which:
- **POSIX**: calls `setsid()` in the child ([[ILibProcessPipe.c:928](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L928)](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L928)) → new session, decoupled from the parent's controlling terminal/process group, survives the parent.
- **Windows**: adds `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` to the creation flags.

⚠️ **Gotcha — `detached:true` ≠ `type:4 (DETACHED)`:**
- `detached:true` still wires up stdio pipes (you keep getting `stdout`/`stderr`); it only changes the session/process-group.
- `type:4` creates **no** pipes at all.
They're orthogonal and can combine.

## `env` — child environment (full replacement, and overloaded for TERM)
[[Lines 528–562](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L528-L562)](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L528-L562): enumerates the `env` object's own keys into a `key,value,...` array → passed as `envvars`. Windows builds a `CREATE_UNICODE_ENVIRONMENT` block; POSIX builds `KEY=VAL` strings for `execve`. If `env` is omitted, `envargs` stays `NULL` and the child **inherits the agent's environment**.

Two important nuances:
1. **It's a full replacement, not a merge.** Unlike the `execve` binding (which explicitly appends `process.env`, [[lines 632–633](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L632-L633)](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L632-L633)), `execFile` passes *only* the keys you provide. `{ env: { FOO: 'bar' } }` gives the child essentially just `FOO`.
2. **For `type:3` (TERM), `env` is reinterpreted as terminal config, not environment.** In the TERM branch the same buffer is parsed for `LINES`, `COLUMNS`, `c_iflag/c_oflag/c_cflag/c_lflag` ([[ILibProcessPipe.c:772–819](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L772-L819)](https://github.com/Ylianst/MeshAgent/blob/master/microstack/ILibProcessPipe.c#L772-L819)) to size the PTY and set termios flags — and `TERM=xterm-256color` is force-injected ([[line 535](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L535)](https://github.com/Ylianst/MeshAgent/blob/master/microscript/ILibDuktape_ChildProcess.c#L535) and again in the child).

---

So the four split into: **`type`** (spawn mechanism/context), **`uid`** (POSIX setuid / Windows session), **`detached`** (session-detach flag, distinct from `type:4`), and **`env`** (replacement environment, repurposed as PTY/termios settings under `type:3`).

PR #7910, opened 2026-06-18
Win-wmi-fixed: add wmi exec method

PTR-inc:
Adds the function 'execMethod' to call wmi methods on classes and instances. Function description in the source.
Add windows_volumes_wmi to win-volumes.js as an example usage of the new wmi methods to get the bitlocker info instead of launching manage-bde.
Also windows_volumes_wmi adds windows 7 bitlocker support.
@si458 The windows_volumes_wmi function is made to fit into the current db volume info definition, but can be easily adepted/added to #7887 if you want

### DIFF
--- a/agents/modules_meshcore/win-volumes.js
+++ b/agents/modules_meshcore/win-volumes.js
@@ -111,7 +111,98 @@ function windows_volumes()
     return (p1);
 }
 
-module.exports = { 
+// Same as windows_volumes() but get the BitLocker recovery password through wmi. And ψ7 added
+function windows_volumes_wmi()
+{
+    var promise = require('promise');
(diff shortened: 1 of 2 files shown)


agents/modules_meshcore/win-volumes.js
```diff
@@ -111,7 +111,98 @@ function windows_volumes()
     return (p1);
 }
 
-module.exports = { 
+// Same as windows_volumes() but get the BitLocker recovery password through wmi. And ψ7 added
```
si458:
```suggestion
// Same as windows_volumes() but get the BitLocker recovery password through wmi. And win7 added
```

PR #7925, opened 2026-06-22
Disable 'minifycore' option for now

PTR-inc:
to prevent mistakenly using old minified versions until there is made a discission to update them or remove this


PTR-inc:
Just made a testminify of the modules_meshcore dir through 'translate.js minifydir' and the size reduction is from 410k to 202k.
Is that worth it?

si458:
not going to merge because its OFF BY DEFAULT!
you have to actually enable the option in the first place.
also its not shown in the schema file (as it shouldnt be)

PTR-inc:
Yeah, but if someone's got it enabled from way back when it was a documented option it's going to fail. It's more a better safe than sorry thing.

Issue #7918, opened 2026-06-20
Dependency cleanup: No need to have archiver and @zip.js


panreyes:
**Describe the bug**

archiver is used for:
- Backups
- A ZIP file with the Windows agents (although I don't know where it's currently being used)

But at the same time, @zip.js is also being used in other places.

I like that @zip.js is only one dependency, while archiver has [66 dependencies](https://github.com/user-attachments/files/29165666/archiver_dependencies.txt) (+ the ones related to archiver-zip-encrypted, not counted here), so it would be a good idea to reimplement the functions using archiver to use @zip.js instead.

I already have it working and tested in this branch:
https://github.com/panreyes/MeshCentral/tree/remove_archiver

si458:
It's on the to-do list! Just not had time, will be next months release probably now

https://github.com/Ylianst/MeshCentral/pull/7916
https://github.com/Ylianst/MeshCentral/pull/7914
https://github.com/Ylianst/MeshCentral/pull/7661

panreyes:
Perfect, thanks for letting me know about those!

In any case, this implementation is already a 1:1 archiver to @zip.js replacement and, AFAIK, can be integrated as it is until @PTR-inc and you have the newer methods ready.

PTR-inc:
Indeed, the idea is to move the backup functions now in db.js to the new backup.js module, use zip.js and extend the current restore function to allow for other db engines as it now only allows nedb for online restore.

si458:
not stale just ran out of time!

PR #7927, opened 2026-06-22
Replace archiver with @zip.js and remove archiver dependency

panreyes:
Replace archiver with @zip.js (which is already being used in MeshCentral), and remove archiver dependency.
It's a 1:1 replacement, can be used/integrated as it is until the new backup methods are ready.

It removes 66 dependencies.

(Furtherly explained in #7918 )


panreyes:
Also added fixes related to showing probable main LAN IP.
Feel free to accept the PR with everything together or to cherry pick what you consider best :)

si458:
@PTR-inc would you like to test this out when you have chance
wont merge for this release
will do it for next months release instead

panreyes:
If it's possible, I would like to request to fit it in this month's release.
No special reason, aside from requiring 🙌 65 less dependencies! 🥳

For people like me, that are really worried about security, the many very random names in packages with just a simple function in the npm system and possible chain supply attacks, it's an important step towards making MeshCentral more reliable :)

si458:
@panreyes no chance, I cant work on anymore mesh stuff at moment as I'm snowed under with ozzie appointments and normal work, so will be in next months now

Also I need to test it all out etc first too

panreyes:
Of course, no rush then. It also contains a big pile of code for replacing glob() function (which is not available in @zip.js) that definitely will benefit from extra reviews.

Thanks!

PTR-inc:
@panreyes Could you split this PR into separate PRs based on which part the change is? So all the archiver removal commits are separate PRs, the mainlanip commits toghether are a separate PR and the battery one also. This keeps everything clean and easier to review.
First quickcheck of the archiver/zip part seems ok, will check the glob part later.

panreyes:
Yes, I'll do it later today. I tried to do multiple PRs before with branches, but for some reason they didn't work well and I ended up having everything in the main branch.

panreyes:
Replaced by: https://github.com/Ylianst/MeshCentral/pull/7947 https://github.com/Ylianst/MeshCentral/pull/7949 and https://github.com/Ylianst/MeshCentral/pull/7948

PR #7936, opened 2026-06-25
Generated systemd unit: add NODE_OPTIONS=--experimental-require-module

acervantes23-amc:
## Problem

On Node 22 (versions 22.0 – 22.11), users running MeshCentral 1.2.x cannot enroll a TOTP authenticator app (Google Authenticator, Authy, etc.). The QR-code dialog under *My Account → Manage authentication app* shows "Unable to load OTPLIB" and no QR is generated. Existing users with a previously-enrolled TOTP can still log in (their stored secret is reused), but no new enrollment can happen.

## Root cause

`otplib` was bumped to v13.4.1 in MeshCentral 1.2.x. Internally, `otplib` pulls in `@otplib/plugin-base32-scure`, which `require()`s `@scure/base`. `@scure/base` is an ESM-only package, and on Node 22 (<22.12) `require()`-ing an ESM throws `ERR_REQUIRE_ESM`:

```
Error: require() of ES Module .../@scure/base/index.js from .../@otplib/plugin-base32-scure/dist/index.cjs not supported.
```

`meshuser.js` catches this silently:

```js
var otplib = null;
try { otplib = require('otplib'); } catch (ex) { }
if (otplib == null) { ws.send(JSON.stringify({ action: 'otpauth-request', err: 6 })); return; }
```

…and replies with `err: 6`, which the UI renders as "Unable to load OTPLIB".

Node 22.12+ enables ES-modules-in-`require()` by default, so this only affects the older 22.x line.

## Fix

Add `Environment=NODE_OPTIONS=--experimental-require-module` to the systemd unit file MeshCentral writes during `meshcentral --install`. With the flag set, `require('otplib')` succeeds and TOTP enrollment works again.

The flag is harmless on Node 22.12+ (it produces a one-time deprecation warning and is otherwise a no-op) and will only become invalid in some future Node major where ESM/CJS interop is reworked again, at which point this line can be revisited.

Existing installs will need to add this line manually to their already-deployed unit, or fall back to running with `NODE_OPTIONS` set in the environment.

## Reproduce

1. Fresh install: `npm install -g meshcentral`, then `sudo meshcentral --install`.
2. With Node v22.0–22.11, open a fresh account in the web UI, go to *My Account → Manage authentication app*.
3. **Before the patch:** a small dialog says "Unable to load OTPLIB". No QR is shown.
4. **After the patch (or after manually adding the env var):** the QR appears and can be enrolled in any TOTP app.

## Tested on

- Node v22.11.0
- Ubuntu 22.04.5 LTS
- MeshCentral 1.2.1
- Google Authenticator (Android)


si458:
no, this is not the solution!

this wont fix things like windows server running meshcentral
or people who run meshcentral with the likes of PM2 on linux

instead file a bug report and we can look into your setup properly

edit: also tested here and it works with node 22 without any issues? so not sure why your having problems?

PR #7947, opened 2026-06-26
Replace archiver with @zip.js and remove archiver dependency

panreyes:
# Summary

Replace archiver with @zip.js (which is already being used in MeshCentral), and remove archiver dependency.
It's a 1:1 replacement, can be used/integrated as it is until the new backup methods are ready.

It removes 66 dependencies.

(Furtherly explained in https://github.com/Ylianst/MeshCentral/issues/7918 )

- [X] 🧰 Dependency updates are listed and explained.
- [X] ⚠️ CI passes and is green.


si458:
@PTR-inc when you have a minute (no rush) please can you verify this PR and see if the is any changes needed?

PTR-inc:
Yes, will do, as promised.

PTR-inc:
: commit fc23132 (package(-lock).json)
- don't manually edit package(& -lock).json, use 'npm uninstall archiver'. This updates both files the proper way and checks all dependencies.

: commit [c29642d](https://github.com/Ylianst/MeshCentral/pull/7947/commits/c29642dafbb8a8d4f4cd54b5cd2248d7ed3066a7) (webserver.js)
- Redundant 'setContentDispositionHeader', set twice in the same function., remove one.

: commit 2bf0b86 (meshcentral.js)
- as long as the current zip-reader.js is used in meshagent, ziplevel 9 must not be changed and also add `zip64: false` to the options as zip-reader doesn't handle zip64, just to be sure. Add this as a comment next to the two zipfunctions.
- And here also a double set `level 9` option for zipwriter, redundant, remove one
- Error handling: Use `console.error()` in both catches

: commit 30ce06e (db.js)
- minimatch require is outside of the try/catch, move it up next to the zip.js, because:
  - minimatch should be added to package.json. It works now because others have it as a dependency. Make it explicit if you want to use it on its own.
  - your treewalk follows symlinks, archiver didn't. And following symlinks shouldn't be done in my opinion, don't go outside the meshcentral-data dir. (@si458 opinion?)


- minimatch is a matcher, not a globber, don't write your own.
- check the package `readdir-glob`, as this is the package used by archiver. Easier & same result. Does all the treewalking, globbing, excluding etc. for you. And only one dependency...minimatch!
- since node v24 there is fs.promises.glob, which does almost exactly the same thing. No dependency! But we're not at that minimal version yet. Readdir-glob can be replaced eventually almost in-place by this function.

Commits c29642d & 2bf0b86 are fine after the suggested changes. Make them separate pr's as suggested in the previous pr so they can be independently accepted, no cherry-picking.
Replace minimatch with readdir-glob (and again a separate pr). As this is the final pr to remove archiver usage, add the removal to this one as a separate, final commit.

panreyes:
Thanks for the comments.
Node, Javascript and the npm ecosystems are far from my comfort zone, to be honest, and I have used AI more than I should in those changes.

I have followed your suggestions and decided to replace those custom glob functions with `fs.promise.glob`, which seems available and working from Node v22.

I have tested those changes and, as far as I can see, they work fine.

PTR-inc:
Alright, I'll look at it tonight.

PTR-inc:
First observations:

- Subdirs aren't globbed
- The override/files/record path aren't globbed.
- Minimatch still isn't declared properly, but I think the path.matchesGlob function in node can replace minimatch if we're going the minimal v22 path and less dependencies.
- Should add a warning if using a node version < v22.17 (<v22.20 if minimatch is replaced)

PTR-inc:
If you want, I can amend the pr with the changes needed

Tech-Werkz:
im just curious, may i ask why we need to change archiver to zip.js ? 
is it way better for compression ?

si458:
@Tech-Werkz is gives us loads of possibilities
the new download backup with a password from the web ui and this uses zip.js under the hood,
this wasnt possible with archiver.
zip.js also lets u restore from a zip file too where as archiver is just an archiver tool, not an unzipper

edit: also belows reason too

PTR-inc:
@Tech-Werkz zip.js is better maintained and has less (0) dependencies. More futureproof.

panreyes:
> If you want, I can amend the pr with the changes needed

Yes, please do that if you prefer, I'm out for a few days

PTR-inc:
@si458 What will be the minimum node version for the next release? Is the nodejs EOL map going to be followed? Just so I know if the new node functions can be used

si458:
@PTR-inc i wanted to force node 22.13 because `require(esm) is now enabled by default` and this is NEEDED for the `otplib` library to not break at the moment
but at the same time node 22 is EOL in like 8 months time! https://endoflife.date/nodejs

PTR-inc:
Would be nice if you could push it to v22.20 so I can eliminate the minimatch dependency :-)

panreyes:
Just a quick observation: @si458 . if I'm not wrong, didn't we push the minimal version to Node 24 recently because of otplib?

PR #7949, opened 2026-06-26
Add file existence check before reading battery data in Linux

panreyes:
# Summary

Certain Linux devices might expose that they have a battery (like Nintendo Joy-Cons), but they might not have the expected files to retrieve the capacity, for example.

This small fix in the agent's code just checks that the file exists before trying to read it.

Fixes this: 
https://discord.com/channels/891895284005752893/891895284005752896/1519506362441273436


si458:
ok ill merge but its inside a try already so it should catch the error? 
not sure why it wouldnt catch the error and instead error out?
i hate duktape haha

PR #7950, opened 2026-06-26
Safer helper functions

panreyes:
# Summary

Some helper functions in default*.handlebars don't filter the input fields as expected.
It will be best to update them to prevent possible HTML injections.

(I would even go further and even do the same to every other helper function unless a new optional parameter is true, like "is_html")


si458:
can you explain why you would need to EscapreHTML the date or time?

si458:
slap on the wrists (both me and u)
you got the modern ui functions the wrong way round  in the code and i hadnt spotted it haha
thankfully @Ndaboom spotted it while fixing the themes!

PR #7957, opened 2026-06-29
Improve robustness of the Intel AMT MPS server

mlinton:
Hi - a few small input-handling improvements to the Intel AMT MPS server (mpsserver.js):

- Treat the certificate subject "O" field as a string before parsing it (it can come through as an array when a certificate has repeated O attributes).
- Wait for the protocol-version handshake before processing an auth request.
- Use the correct length value when validating CHANNEL_DATA frames.

These just tighten up a few edge cases in how incoming data is handled - there is no behaviour change for normal Intel AMT devices. Happy to split this into separate PRs or reword anything if you'd prefer.

Thanks for all your work on MeshCentral!

### DIFF
--- a/mpsserver.js
+++ b/mpsserver.js
@@ -439,7 +439,9 @@ module.exports.CreateMpsServer = function (parent, db, args, certificates) {
                 socket.tag.first = false;
 
                 // Setup this node with certificate authentication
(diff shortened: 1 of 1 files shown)


mlinton:
Tidied up the branch name and re-opened this as #7958 - please use that one instead. Sorry for the small bit of churn!

PR #7963, opened 2026-07-01
Add wsl check and remove throws

PTR-inc:
# Summary

- Adds wsl check to 'computer-identifiers.js'
- Adds try/catch to the bios part
- Removed the throws in the bios part so the function continues on regardless with gathering the rest of the linux info and classifies with 'Uknown' if no bios info found.
- Adds check to isVM ('vm' console command)

<details>

</details>

### DIFF
--- a/agents/modules_meshcore/computer-identifiers.js
+++ b/agents/modules_meshcore/computer-identifiers.js
@@ -1078,6 +1078,7 @@ module.exports.isVM = function isVM()
         }
     }
 
+    if (id.identifiers.bios_vendor == 'Microsoft' && id.identifiers.bios_version.slice(0,3) === 'WSL') { ret = true; }
 
     if (!ret) { ret = this.isDocker(); }
     return (ret);
--- a/agents/modules_meshcore/computer-identifiers.js
+++ b/agents/modules_meshcore/computer-identifiers.js
@@ -70,55 +70,58 @@ function linux_identifiers()
     var ret = {};
     var values = {};
     var child = null;
-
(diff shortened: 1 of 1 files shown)


agents/modules_meshcore/computer-identifiers.js
```diff
@@ -70,55 +70,58 @@ function linux_identifiers()
...
-        identifiers['bios_version'] = ret['bios_version'];
-        identifiers['bios_serial'] = ret['product_serial'];
-        identifiers['board_name'] = ret['board_name'];
-        identifiers['board_serial'] = ret['board_serial'];
-        identifiers['board_vendor'] = ret['board_vendor'];
-        identifiers['board_version'] = ret['board_version'];
-        identifiers['product_uuid'] = ret['product_uuid'];
-        identifiers['product_name'] = ret['product_name'];
-    }
+    } catch (e) { console.log(e && e.message ? e.message : e); }
```
si458, line 124: Don't need the console log messages :)

PTR-inc: Could make it console.info1, but it's an error so I thought it should appear in the local log if enabled this way

si458: But console.log does nothing when running as a service, it will only output to the command window if u run the app from the command line

PTR-inc: You can enable the output redirection with the setdebug console command. Default is off, so it does nothing normally, but 'setdebug 2' outputs to the webconsole and 'setdebug 8' outputs to the local log.

PR #7966, opened 2026-07-02
Only send Slack messages to hooks.slack.com

arkha10:
Hi! Small fix. The Slack handle was used to build the request URL directly, so it could point anywhere. Slack incoming webhooks are always on hooks.slack.com, so this checks the host before sending. Real Slack webhooks keep working exactly as before.


si458:
thanks for this!
this also could be an extra feature in the future!
so we could allow ANY hook urls using say `hooks:https://...` etc

PR #7973, opened 2026-07-08
fix maxrecordings

fernandobayit:
this is fix parameter maxrecordings set in config.json

### DIFF
--- a/db.js
+++ b/db.js
@@ -28,7 +28,19 @@
 module.exports.CreateDB = function (parent, func) {
     var obj = {};
     var Datastore = null;
-    var expireEventsSeconds = (60 * 60 * 24 * 20);              // By default, expire events after 20 days (1728000). (Seconds * Minutes * Hours * Days)
+	var expireEventsSeconds = (60 * 60 * 24 * 20);              // By default, expire events after 20 days (1728000). (Seconds * Minutes * Hours * Days)
+	// If sessionrecording.maxrecordingdays set, use this is value
+	if (parent.args && parent.args.domains) {
+	  for (var d in parent.args.domains) {
+	    var domain = parent.args.domains[d];
+	    if (domain.sessionrecording && typeof domain.sessionrecording.maxrecordingdays == 'number') {
+	      var recordingDays = domain.sessionrecording.maxrecordingdays;
+	      if (recordingDays * 86400 > expireEventsSeconds) {
+	        expireEventsSeconds = recordingDays * 86400;
+	      }
+	    }
+	  }
+	}
     var expirePowerEventsSeconds = (60 * 60 * 24 * 10);         // By default, expire power events after 10 days (864000). (Seconds * Minutes * Hours * Days)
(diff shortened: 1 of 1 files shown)


si458:
can u explain your fix?
as you havent followed the pr template so we dont know what its meant to do etc?

fernandobayit:
This pull request fixes an issue in MeshCentral where the `maxrecordingdays` setting is not being properly enforced.

Currently, the server only lists recordings from the last 20 days, regardless of the value configured in `maxrecordingdays`. This behavior prevents administrators from accessing recordings that should still be available according to the configured retention period.

Changes included:

* Correct the logic used to filter and display recordings.

* Ensure that the `maxrecordingdays` parameter is respected when retrieving the recording list.

* Allow recordings older than 20 days to be listed when they are still within the configured retention period.

* Preserve the existing behavior for recordings that exceed the configured retention limit.

The goal of this PR is to make the recording listing behavior consistent with the `maxrecordingdays` configuration and ensure that all valid recordings are displayed as expected.

si458:
erm, the expireEventsSeconds already gets set further down in the code, so im confused?
`if (typeof parent.args.dbexpire.events == 'number') { expireEventsSeconds = parent.args.dbexpire.events; }`
```
{
  "$schema": "https://raw.githubusercontent.com/Ylianst/MeshCentral/master/meshcentral-config-schema.json",
  "settings": {
    "dbExpire": {
      "events": 86400
    }
  },
  "domains": {
    "": {}
  }
}
```
so if you use `dbExpire -> events` then `domain.sessionrecording.maxrecordingdays` would be irrelevant?

fernandobayit:
I tried this method before, but it still didn't work. The change I proposed was the solution I found.However, if it's not suitable, we can revert and remove the pull request.On 12 Jul 2026, at 12:02, Simon Smith ***@***.***> wrote:﻿si458 left a comment (Ylianst/MeshCentral#7973)
erm, the expireEventsSeconds already gets set further down in the code, so im confused?
if (typeof parent.args.dbexpire.events == 'number') { expireEventsSeconds = parent.args.dbexpire.events; }
{
  "$schema": "https://raw.githubusercontent.com/Ylianst/MeshCentral/master/meshcentral-config-schema.json",
  "settings": {
    "dbExpire": {
      "events": 86400
    }
  },
  "domains": {
    "": {}
  }
}

si458:
this PR wont work. (and im still confused anyways?)
but you are looping the domains argument which can have multiple domains in.
and each domain can have its own maxrecordingdays,
and you are overwriting expireEventsSeconds with every check, 
also what if 1 domain only wants to see 2 days but another wants 5 days?
the 2 day domain would end up seeing 5 days worth

si458:
also just re-read the code, the maxrecordingdays is ONLY used for the `.mrec` files themselves, not the actual events!
the events is controlled by what i said above about `dbExpire -> events`
also if you check the schema description too
```
"maxRecordingDays": {
   "type": "integer",
   "default": null,
   "description": "Maximum number of days to keep a recording."
},
```
the description says nothing about the events, but the recordings (meaning the files themselves)

PR #7978, opened 2026-07-11
Fix 2FA on Node 22 by preloading otplib via dynamic import()

acervantes23-amc:
### Problem

`otplib` v13+ ships as an ES module. On Node.js 22 before 22.12, `require()` of an ES module throws `ERR_REQUIRE_ESM` unless the process is started with `--experimental-require-module`.

`meshuser.js` and `webserver.js` load otplib with `require('otplib')`, so on a default Node 22 install **TOTP two-factor auth is silently broken**:

- authenticator enrollment (`otpauth-request` / `otpauth-setup`) returns error `6` because `require('otplib')` throws and `otplib` stays `null`
- `checkUserOneTimePassword()` in `webserver.js` throws on login because that `require` is not wrapped in a guard

### Why not the Node flag

Requiring `--experimental-require-module` is not portable — it has to be set differently for systemd, Docker, PM2, launchd, `pkg` builds, Windows service wrappers, etc. That is why the generated‑systemd‑unit approach in #7936 was incomplete and closed.

### Fix

Preload otplib once at module scope with **dynamic `import()`** (which is allowed inside CommonJS) and reuse the module at the call sites. No Node flag is needed on any runtime. The existing `otplib == null` guards are kept, and the authenticator check in `webserver.js` now also guards on otplib being loaded.

Net change: +15 / −8 across `meshuser.js` and `webserver.js`, no new dependency.

### Testing (Node 22.11.0)

- `require('otplib')` fails with `ERR_REQUIRE_ESM` **without** the flag (reproduces the bug); `import('otplib')` resolves and a full `generateSecret` → `generateURI` → `generateSync` → `verifySync` (with `createGuardrails`) round‑trip succeeds.
- Applied the patch to a running MeshCentral 1.2.1 server, **removed** `NODE_OPTIONS=--experimental-require-module`, restarted: server starts cleanly, no `ERR_REQUIRE_ESM`, and the ES‑module `require()` experimental warning is gone.
- Confirmed in the live process that both modules populate `otplib` via `import()` (`verifySync` and `generateSecret` present) after startup.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/meshuser.js
+++ b/meshuser.js
@@ -19,6 +19,12 @@ const driveType = { 0: "Unknown", 1: "No Root Directory", 2: "Removable Disk", 3
 const conversionStatus = { "-1": "Unknown", 0: "Fully Decrypted", 1: "Fully Encrypted", 2: "Encryption In Progress", 3: "Decryption In Progress", 4: "Encryption Paused", 5: "Decryption Paused" };
 const protectionStatus = { 0: "Off", 1: "On", 2: "Locked"};
 
+// otplib v13+ ships as an ES module. On Node.js 22 before 22.12, require() of an ES module throws
+// ERR_REQUIRE_ESM unless started with --experimental-require-module. Preload it once via dynamic
+// import() so 2FA works on every runtime without a Node flag (systemd, Docker, PM2, launchd, pkg, ...).
+var otplib = null;
+import('otplib').then(function (mod) { otplib = mod; }, function () { otplib = null; });
+
 // Construct a MeshAgent object, called upon connection
 module.exports.CreateMeshUser = function (parent, db, ws, req, args, domain, user) {
     const fs = require('fs');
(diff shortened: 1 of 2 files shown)


si458:
nodejs 20 is now EOL and nodejs 22 will be EOL in 9 months anyways - https://endoflife.date/nodejs
please just upgrade to node 22.13+ OR better yet, node 24
https://github.com/Ylianst/MeshCentral/issues/7956
```
# Download and install nvm:
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.5/install.sh | bash

# in lieu of restarting the shell
\. "$HOME/.nvm/nvm.sh"

# Download and install Node.js:
nvm install 22

# Verify the Node.js version:
node -v # Should print "v22.23.1".

# Verify npm version:
npm -v # Should print "10.9.8".
```

acervantes23-amc:
Thanks @si458, that's fair — this only matters on Node < 22.12, and since
Node 20 is EOL there's no reason to carry a workaround for it in the code.
I'll upgrade our servers to Node 22.13+ (where require() of the ESM otplib
works natively) and drop the flag on our side. Closing this since upgrading
Node is the proper fix. Appreciate the review and the nvm steps!

Issue #7938, opened 2026-06-25
upnusers not cleaned up in database (postgres)


stephannn:
Hi,

I just saw in my postgres SQL main table that the "upnusers" in the docs of a node is full with the same UPN user. It is like that each time the UPN is added but never cleaned, or checked before adding. "users" on the other hand is fine and only has one record.

**To Reproduce**
Open the JSON in the docs column of a node in the main table.

```
SELECT doc, id, type, domain, extra, extraex
	FROM public.main
	WHERE type = 'node';
```

**Expected behavior**
Should not be overflowed

**Server Software (please complete the following information):**
 - OS: Ubuntu
 - Virtualization: Podman
 - Network: LAN
 - Version: 1.2.0 (have also not seen a recent PR)
 - Node: docker image
 - Postgres DB

I don't know how it behaves in other DBs

Edit:
Not sure if this is correct:
```
                if ((command.users != null) && (Array.isArray(command.users)) && (device.users != command.users)) { device.users = command.users; change = 1; } // Don't save this to the db.
                if ((command.lusers != null) && (Array.isArray(command.lusers)) && (device.lusers != command.lusers)) { device.lusers = command.lusers; change = 1; } // Don't save this to the db.
                if ((command.upnusers != null) && (Array.isArray(command.upnusers)) && (device.upnusers != command.upnusers)) { device.upnusers = command.upnusers; change = 1; } // Don't save this to the db.
```

device.upnusers is an array but maybe compared only with a string:
`if ((command.upnusers != null) && (Array.isArray(command.upnusers)) && (command.upnusers.includes(device.upnusers) === false)) { device.upnusers = command.upnusers; change = 1; } `
?

Edit2:
It seems not always be the case and maybe related to the function getLogonCacheKeys()

si458:
This was caused by a bug the other month when we introduced the upnusers.
So we just need to fix it so it removed duplicates before it stored the info etc into the db

stephannn:
Fixed by #7990

PR #7990, opened 2026-07-16
UPN Duplicates

stephannn:
As described in: https://github.com/Ylianst/MeshCentral/issues/7938

Added an if clause like used for:
`if (meshCoreObj.lusers.indexOf(un) == -1) { meshCoreObj.lusers.push(un); } }`
some lines above

Edit:
At first I thought this is flooding the DB, but it is temporally only flooding it every time the machines gets locked/unlocked

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -824,7 +824,7 @@ function onUserSessionChanged(user, locked) {
                 if(userobj && userobj.length > 0){
                     for (var j = 0; j < userobj.length; j++) {
                         if (userobj[j] && userobj[j].SAM && userobj[j].SAM[0].trim() === a[i].Username) {
-                            meshCoreObj.upnusers.push(userobj[j].UPN);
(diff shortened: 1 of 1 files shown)


agents/meshcore.js
```diff
@@ -5751,7 +5751,7 @@ function processConsoleCommand(cmd, args, rights, sessionid) {
                                 v.push({ tsid: i, type: u[i].StationName, user: u[i].Username, domain: u[i].Domain, upn: u[i].UPN });
                             }
                         }
-                        sendConsoleText(JSON.stringify(v, null, 1), this.sessionid);
+                        //sendConsoleText(JSON.stringify(v, null, 1), this.sessionid);
```
si458: I think this was commented out by mistake???

si458 (changes requested):

stephannn:
yes and no, I rolled it back

PR #8017, opened 2026-07-21
Escape guest name in the device Sessions dialog

Isorgcom:
# Summary

- Escaped the guest name in the device **Sessions** dialog. `getUserName()` (in `views/default.handlebars`) decodes the guest name from the user id and returned it unescaped, and the Sessions dialog places that result into the page as HTML — so a guest name containing markup was rendered as-is instead of shown as text. The decoded name now goes through `EscapeHtml`, so it always displays literally.
- This matches how the same guest name is already handled elsewhere in the file, so it's just making the Sessions dialog consistent with the rest.

Small, self-contained change — one line, no UI/behavior change for normal names.

<details>

- [x] 🧠 I used LLMs/AI in this contribution and reviewed all generated content. I understand that I am responsible for and able to explain every line of code I submit.

</details>

### DIFF
--- a/views/default.handlebars
+++ b/views/default.handlebars
@@ -21421,7 +21421,7 @@
         function encodeURIComponentEx(txt) { return encodeURIComponent(txt).replace(/'/g,'%27'); };
         function getUserName(userid) {
             var useridsplit = userid.split('/'), userid2 = useridsplit[0] + '/' + useridsplit[1] + '/' + useridsplit[2], guestname = '';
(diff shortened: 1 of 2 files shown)


si458:
you need to also check default3.handlebars as this is the modern ui
where as default.handlebars is the old classic ui

Isorgcom:
Good catch, thanks! I've pushed the same fix to `default3.handlebars` — `getUserName()` is identical there and feeds the Sessions dialog the same way, so the modern UI is now covered too. Both files are in the PR.

PR #8024, opened 2026-07-22
Fix RDP keyboard layout selection from browser locale (#5097)

### LINKED ISSUE #5097: RDP defaults to en-US keyboard layout despite configuration settings
I am using MeshCentral in combination with Tactical RMM for remote desktop connections, and I have encountered an issue where the keyboard layout defaults to en-US despite my attempts to change it to de-DE (German).
It will stick to de-DE when I change it, however the moment I reconnect to the machine, it changes back to en-US.

I have tried removing the English keyboard layout from the remote computer. However, this has not resolved the issue. Every time I connect to a remote computer using MeshCentral's RDP feature, the keyboard layout reverts to en-US.

Here is my config.json:
```
{
  "settings": {
    "Cert": "rmm-mesh.mydomain.com",
    "MongoDb": "mongodb://127.0.0.1:27017",
    "MongoDbName": "meshcentral",
    "WANonly": true,
    "Minify": 1,
    "Port": 4430,
    "AliasPort": 443,
    "RedirPort": 800,
    "AllowLoginToken": true,
    "AllowFraming": true,
    "_AgentPing": 60,
    "AgentPong": 300,
    "AllowHighQualityDesktop": true,
    "TlsOffload": "127.0.0.1",
    "agentCoreDump": false,
    "Compression": true,
    "WsCompression": true,
    "AgentWsCompression": true,
    "MaxInvalidLogin": { "time": 5, "count": 5, "coolofftime": 30 }
  },
  "domains": {
    "": {
      "Title": "Tactical RMM",
      "Title2": "Tactical RMM",
      "NewAccounts": false,
      "CertUrl": "https://rmm-mesh.mydomain.com:443/",
      "GeoLocation": true,
      "CookieIpCheck": false,
      "mstsc": true,
      "remoteDesktop": {
        "disableWallpaper": false
      }
    }
  }
}
```

I would appreciate any assistance or guidance on how to resolve this issue and set the default keyboard layout to de-DE for remote desktop connections in MeshCentral. Please let me know if additional information is required.

Thank you.

ChiliApple:
# Summary

- `rdp/protocol/rdp.js` now derives the RDP keyboard layout from the primary
  language subtag of `config.locale`, instead of comparing the whole locale
  string against `'fr'` and `'en'` only.

The Web-RDP client matched `config.locale` exactly, but browsers report full
language tags (`de-AT`, `de-DE`, `fr-FR`). Anything that was not literally
`'fr'` or `'en'` fell through to the default and was announced to the RDP host
as the US layout (0x409), so non-US users got an en-US input locale added to
their session on every new Web-RDP connection.

The new mapping only uses layouts that are already defined in
`t125.gcc.KeyboardLayout`, so no new constants are introduced and the default
stays US. As a side effect this also fixes French, which previously only worked
when the browser reported a bare `fr` rather than `fr-FR`.

Region variants (`de-CH` 0x807, `fr-CH` 0x100C, `en-GB` 0x809, `pt-BR` 0x416)
still fall back to their base language, because `KeyboardLayout` has no entries
for them. Happy to extend the enum in a follow-up if that is wanted.

The root cause was identified by @GA-Leftclick in #5097; this turns that finding
into a patch.

Fixes #5097

## Testing

Built into a custom image on top of `ghcr.io/ylianst/meshcentral:1.2.2` and
deployed on a production server (Synology Docker, Cloudflare Tunnel, WAN mode).
Verified in a Web-RDP session against German-language Windows hosts: the session
now comes up with the German input locale on first connect, QWERTZ, umlauts and
AltGr combinations behave correctly. Before the change the same session came up
as en-US every time.

<details>

- [ ] 🔍 Any UI changes adhere to visual style of this project. (no UI changes)
- [ ] 🤖 I added or updated automated tests where appropriate. (no test harness for this path)
- [ ] 📄 Documentation updates are included (if applicable). (not applicable)
- [ ] 🧰 Dependency updates are listed and explained. (none)

</details>

### DIFF
--- a/rdp/protocol/rdp.js
+++ b/rdp/protocol/rdp.js
@@ -135,18 +135,25 @@ function RdpClient(config) {
     log.debug('screen ' + this.mcs.clientCoreData.obj.desktopWidth.value + 'x' + this.mcs.clientCoreData.obj.desktopHeight.value);
 
     // config keyboard layout
-    // config.locale is a browser language tag (ex: 'de-AT', 'fr-FR', 'en-US'), so only
-    // the primary language subtag is used to select the RDP keyboard layout.
+    // config.locale is a browser language tag (ex: 'en-GB', 'de-AT', 'fr-CH'). The region
+    // can select a different layout, since 'en-GB' is not 'en-US', so the full tag is
+    // looked up first and only then the language on its own.
     var localeToKeyboardLayout = {
-        'ar': 'ARABIC',    'bg': 'BULGARIAN', 'cs': 'CZECH',     'da': 'DANISH',
-        'de': 'GERMAN',    'el': 'GREEK',     'en': 'US',        'es': 'SPANISH',
-        'fi': 'FINNISH',   'fr': 'FRENCH',    'he': 'HEBREW',    'hu': 'HUNGARIAN',
-        'is': 'ICELANDIC', 'it': 'ITALIAN',   'ja': 'JAPANESE',  'ko': 'KOREAN',
-        'nb': 'NORWEGIAN', 'nl': 'DUTCH',     'nn': 'NORWEGIAN', 'no': 'NORWEGIAN',
-        'zh': 'CHINESE_US_KEYBOARD'
+        'ar':    'ARABIC',        'bg':    'BULGARIAN',        'cs':    'CZECH',
+        'da':    'DANISH',        'de':    'GERMAN',           'de-ch': 'SWISS_GERMAN',
+        'el':    'GREEK',         'en':    'US',               'en-gb': 'UNITED_KINGDOM',
+        'en-ie': 'IRISH',         'es':    'SPANISH',          'fi':    'FINNISH',
+        'fr':    'FRENCH',        'fr-be': 'BELGIAN_FRENCH',   'fr-ca': 'CANADIAN_FRENCH',
+        'fr-ch': 'SWISS_FRENCH',  'he':    'HEBREW',           'hu':    'HUNGARIAN',
+        'is':    'ICELANDIC',     'it':    'ITALIAN',          'ja':    'JAPANESE',
+        'ko':    'KOREAN',        'nb':    'NORWEGIAN',        'nl':    'DUTCH',
+        'nl-be': 'BELGIAN_DUTCH', 'nn':    'NORWEGIAN',        'no':    'NORWEGIAN',
+        'zh':    'CHINESE_US_KEYBOARD'
     };
-    var localeLanguage = String(config.locale || 'en').toLowerCase().split(/[-_]/)[0];
-    var keyboardLayoutName = localeToKeyboardLayout[localeLanguage] || 'US';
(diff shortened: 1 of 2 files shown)


si458:
this doesnt work for me :(
my browser is set as en-GB, but you are stripping the country locale out and changing my rdp to en-us

ChiliApple:
Good catch, thanks — en-GB was exactly the blind spot. Stripping the region was too coarse: en-GB is 0x809, not 0x409.

Updated in a second commit: the full tag is looked up first, then the language, then US as before. I only added region variants for languages the enum already covered, with identifiers from the Microsoft keyboard identifier reference — de-CH 0x807, en-GB 0x809, fr-BE 0x80C, nl-BE 0x813, fr-CA 0x1009, fr-CH 0x100C, en-IE 0x1809.

Regions that share the plain language layout are deliberately left out: per the same reference de-AT, de-LI, de-LU use 0x407 and fr-LU, fr-MC use 0x40C.

Could you re-test with your en-GB browser?

si458:
its being detected as `UNITED_KINGDOM 2057` (i console.logged it for testing)
but when i connect its still setting the keyboard language to US
(i only have US enabled in the vm for testing, so it should add in the UK keyboard, unless im wrong?)

ChiliApple:
Thanks for testing. The patch does send the correct layout ID — I checked the serialised client core data, 09 08 00 00 for en-GB.

Windows only activates a layout that's present on the target. US (0x409) always works because it ships with every install; 0x809 doesn't, and you said only US is enabled in that VM. mstsc.exe has the same limitation.

Quick test: add the United Kingdom keyboard in the VM, log off (not just disconnect), reconnect. Per MS-RDPBCGR Appendix A note <6> the layout is only applied to a newly created session.

Also worth ruling out: HKLM\SYSTEM\CurrentControlSet\Control\Keyboard Layout\IgnoreRemoteKeyboardLayout — if that's 1 from earlier #5097 testing, the layout is discarded either way.

Two unrelated things I noticed, separate PR if you want them: codePage in sec.js is never set although INFO_UNICODE is (spec section 4.10 puts the language id in its low word), and global.js:367 has a typo, keyboardrFunctionKey, so that field stays 0 in the input capability set.

si458:
@ChiliApple ignore me, it works!
turns out you must SIGN OUT of your RDP session and not disconnect

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/cbe1ed0a-d320-4ea5-be5a-f2eb6e032853?utm_source=chatgpt.com

<6> Section 2.2.1.3.2: Microsoft RDP servers apply only the active locale identifier to a newly created session. The value is ignored when connecting to an existing session.

EDIT: @ChiliApple just seen your post above! im having one of those days!
thank you so much for your help and support! ❤️

ChiliApple:
Great, thanks for digging that out — sign out vs disconnect explains it. And thanks for the quick review.

PR #8035, opened 2026-07-26
Software inventory read permissions

stephannn:
# Summary

- Added the Bootstraps switch (checkbox) to the style-bootstraps (not used currently, but maybe in the future?)
- Added View Only Permissions to the Software Tab
- I am not sure about the line: else if (url.searchParams.get('p') == '6') { requiredNonRights = MESHRIGHT_NOSOFTWARE; requiredRights = MESHRIGHT_SOFTWAREREAD; }
- Language/Translation Missing

CURRENTLY ONLY Style 3, when OK so far I will do it for the classic style 1.

Did not use AI, but I would had probably a better result.

Did a test with my Windows VM and 7zip


si458:
i dont think the permission numbering is correct?
`0x01600000` translates to `23068672` from what my browser shows?
(hex/binary is confusing to me 😕)

stephannn:
ahh, I thought I just take take the previous value and multiplicate it by 2 (8388608 x 2)

oh jesus, they don't know what they are doing. I will fix it. Is the rest OK? Can I do it for style 1?

public/styles/style-bootstrap.css
```diff
@@ -2965,6 +2965,23 @@ body[data-bs-theme="light"] {
     background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='-4 -4 8 8'%3e%3ccircle r='2' fill='%23fff'/%3e%3c/svg%3e") !important;
 }
 
+.form-switch .form-check-input {
```
si458, line 2968: was this added in by mistake or for a reason?

stephannn: Hi,
in the beginning I tried to make it with a switch (https://getbootstrap.com/docs/5.1/forms/checks-radios/#switches), but noticed that the bootstraps code did not work. With the edited CSS it worked, but I removed the idea. I still left the code there because maybe someone else can use it, but you can also remove it from the PR, it has no impact.

PR #8052, opened 2026-07-31
fix: add null check for obj.agentInfo to prevent TypeError crash

fernandobayit:
Prevents TypeError: Cannot read properties of undefined (reading capabilities) when processing devices with missing agentInfo in the database.

4 lines modified in meshagent.js to add (obj.agentInfo) && guard before accessing obj.agentInfo.capabilities & 0x20.

Summary

- Added null guard (obj.agentInfo) && before accessing obj.agentInfo.capabilities & 0x20 in 4 locations in meshagent.js to prevent a TypeError crash when devices in the database have missing or corrupted agentInfo.
- Lines 92 and 2002 already had this guard — this PR extends the same pattern to the 4 remaining unguarded locations.
- The crash occurs during db.js iteration over device records: when obj.agentInfo is undefined, accessing obj.agentInfo.capabilities throws TypeError, crashing the server process.

Modified lines:

|Line | Before         | After         |
-----|------------------|-------------
|829  |  INLINECODE10  |  INLINECODE11 |
|876  |  INLINECODE12  |  INLINECODE13 |
|2050 |  INLINECODE14  |  INLINECODE15 |
|2159 |  INLINECODE16  |  INLINECODE17 |

**Error trace**:
`meshagent.js:2045
    if (obj.agentInfo.capabilities & 0x20) { event.nolog = 1; }
                      ^
TypeError: Cannot read properties of undefined (reading 'capabilities')
    at meshagent.js:2045:39
    at db.js:1598:21
    at Statement.<anonymous> (db.js:1457:29)`
    
- [x] 🧠 I used LLMs/AI in this contribution and reviewed all generated content. I understand that I am responsible for and able to explain every line of code I submit.
- [x] 🔍 Any UI changes adhere to visual style of this project. (N/A — no UI changes)
- [x] 🛠️ I have self-reviewed and self-tested my code to ensure it works as expected. (Syntax validated with node -c; tested in production server — crash stopped after applying the fix)
- [ ] 🧰 Dependency updates are listed and explained. (N/A — no dependency changes)

### DIFF
--- a/meshagent.js
+++ b/meshagent.js
@@ -826,7 +826,7 @@ module.exports.CreateMeshAgent = function (parent, db, ws, req, args, domain) {
                     db.Set(device);
 
                     // If this is a temporary device, don't log changes
-                    if (obj.agentInfo.capabilities & 0x20) { log = 0; }
+                    if ((obj.agentInfo) && (obj.agentInfo.capabilities & 0x20)) { log = 0; }
 
                     // Event the node change
                     var event = { etype: 'node', action: 'changenode', nodeid: obj.dbNodeKey, domain: domain.id, node: parent.CloneSafeNode(device) };
@@ -873,7 +873,7 @@ module.exports.CreateMeshAgent = function (parent, db, ws, req, args, domain) {
         db.Set(device);
 
         // Event the new node
-        if (obj.agentInfo.capabilities & 0x20) {
+        if ((obj.agentInfo) && (obj.agentInfo.capabilities & 0x20)) {
             // This is a temporary agent, don't log.
             parent.parent.DispatchEvent(parent.CreateMeshDispatchTargets(obj.dbMeshKey, [obj.dbNodeKey]), obj, { etype: 'node', action: 'addnode', node: device, domain: domain.id, nolog: 1 });
         } else {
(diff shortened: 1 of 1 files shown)


panreyes:
Why do we need those guards? It feels like we're fixing the symptom, and probably enabling an insecure behaviour.
I think we should reject any connections that do not include `obj.agentInfo` instead.

fernandobayit:
It’s not a symptom, but rather a fix for the crash occurring in MeshCentral
that causes the service to restart. Here is an explanation of what
currently happens:
What happens:
1. Agent reports an IP change → server initiates a DB callback
(asynchronous)
2. Agent disconnects before the callback completes → `delete obj.agentInfo`
is executed
3. DB callback attempts to read `obj.agentInfo.capabilities` → `undefined`
→ crash

si458:
@panreyes im happy with these changes!
i started to write code so it would always make sure `obj.agentInfo` was available on startup as i believed this was the issue
but as @fernandobayit points out the issue is actually the fact the agents are closing/quiting/exiting/disconnecting SO QUICK
the systems actually calling the disconnect command and then `delete obj.agentInfo` so when the other command hits their point where its checking `agentInfo` its crashing cuz the value doesnt exist!
please keep these fixes coming! ❤️ 
thank you community ❤️

panreyes:
I understand the idea of these changes, but they are also changing the behaviour of those functions and, without further checking, they could be used for bypassing existing protections.

Without more information, my review of these changes is not positive.

We might be converting a DoS bug into a RCE.
Exceptions should be dealt as such, not just ignored.

si458:
@panreyes more testing will be done! but as @fernandobayit said, we are trying to stop the servers crashing at the moment!
i hate this new AI that keeps finding more bugs!
ive already had 3 emails again today!
they just keep coming! we cant keep on top!

PR #8057, opened 2026-08-04
Fixes for certain view only permissions (deprecated)

panreyes:
# Summary
Added some checks related to view only permissions.
Also improved openUrl URL validation to prevent malformed strings.

- [X] 🔍 Any UI changes adhere to visual style of this project.

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -1798,6 +1798,12 @@ function handleServerCommand(data) {
                             break;
                         }
                     case 'openUrl': {
+						// Check if URL is valid (http or https only)
+						if (!isValidHttpUrl(data.url)) {
+							sendConsoleText("OpenURL failed: Invalid URL.");
+							break;
+						}
+						
                         // Open a local web browser and return success/fail
                         MeshServerLogEx(20, [data.url], "Opening: " + data.url, data);
                         sendConsoleText("OpenURL: " + data.url);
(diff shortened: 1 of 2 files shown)


PTR-inc:
The idea is good, but I don't think this is the proper way to do these permissions. Do the check in the commands, in meshcore.js, where I think the validation needs to be anyway. And all the bits are there. Like it's done now with for example the 'bitlocker' command.
Another thought is, why does a viewonly account have console rights anyway? But I don't know the status regarding permissions at the moment and there probably is a case somewhere for console rights and viewonly.
Another, a stance could be 'deny-unless' for every console command. Or a central map of commands and roles/rights. Just thinking.
But I would propose to do it at the command itself, that way you also prevent other avenues of triggering the command, bypassing the serverside permissions check.

panreyes:
Those commands are being sent from the server. There's no other way to submit them into the agents.

So I think it will be best to prevent sending them from the server, as if an attacker can simulate being the server and submit commands to the agents, they will also probably be able to impersonate an administrator.

I don't understand why is it a bad idea to start filtering the commands from the server instead of in the agents directly. If anything, it should be filtered in both.

About the role "view-only", I dislike how it's currently designed. I think it should definitely mean "view-only" and not be compatible with Files, Terminal, etc.

PTR-inc:
There is no other way, as far as you know at this moment. So at least at the last guard, the startingpoint of execution. But an extra guard at the serverside can also be done. But maybe a bit more allround, not just this role. Maybe a kind of a map? So it can be expanded?
And how do the other commits relate to this?

panreyes:
Moved to https://github.com/Ylianst/MeshCentral/pull/8069

PR #8072, opened 2026-08-11
Extract registry functions from meshcore.js and fix reg export

### LINKED ISSUE #8067: Registry export fails in Spanish Windows
**Describe the bug**
Using Registry export fails in Spanish Windows:
Image
(Incorrect syntax)

MeshCentral v1.2.4

panreyes:
# Summary

To avoid meshcore.js being a bit more bloated with platform specific implementations, I have moved the registry functions to its own file, `win-registry-remote.js`. It works just the same.

I also fixes the reg export command call
Fixes #8067

- [X] 🔍 Any UI changes adhere to visual style of this project.

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -881,209 +881,6 @@ var nextTunnelIndex = 1;
 var apftunnel = null;
 var tunnelUserCount = { terminal: {}, files: {}, registry: {}, tcp: {}, udp: {}, msg: {} }; // List of userid->count sessions for terminal, files, registry and TCP/UDP routing
 
-function getRegistryRoots() {
-    return ['HKEY_LOCAL_MACHINE', 'HKEY_CURRENT_USER', 'HKEY_USERS', 'HKEY_CLASSES_ROOT', 'HKEY_CURRENT_CONFIG'];
-}
-
-function getRegistryHiveEnum(hiveName) {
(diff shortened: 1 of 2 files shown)


agents/meshcore.js
```diff
@@ -4103,14 +3900,15 @@ function onTunnelData(data)
             if (cmd == null) { return; }
             if ((cmd.ctrlChannel == '102938') || ((cmd.type == 'offer') && (cmd.sdp != null))) { onTunnelControlData(cmd, this); return; } // If this is control data, handle it now.
             if (cmd.action == undefined) { return; }
+            var remoteRegistry = (process.platform == 'win32') ? require('win-registry-remote') : null;
```
si458: this should be wrapped in a try
because some devices might have a problem loading the module with `require`
this would then cause the meshagent to crash if it couldnt load it for whatever reason

si458 (changes requested):
thanks for fixing and also good idea separating into its own module!
1 change required plz :)
also what was the issue with the reg export?

panreyes:
There were two issues:
1. The argument 0 has to be reg.exe for some reason related to how node works, not export.
2. The arguments with spaces weren't escaped with quotes.

+ Added try catch to the require

si458 (approved):

PR #8075, opened 2026-08-12
fix: add null guard to deviceShares in createDeviceShareLink (#8044)

### LINKED ISSUE #8044: TypeError: Cannot read properties of null (reading 'userid') in HandleEvent — createDeviceShareLink dispatches a holey deviceShares array
**Version:** observed on 1.1.58; the code is still present on current master (`meshuser.js` ~L4613/L4621 as of 2026-07-28).

**What happens:** When a device share is created (web UI or `meshctrl DeviceSharing`) and the node has at least one **expired** share, `createDeviceShareLink` purges the expired rows with `delete docs[i]` — which leaves a **hole** in the array — and then dispatches that same array:

```js
if (doc.expireTime < now) { parent.db.Remove(doc._id, function () { }); delete docs[i]; } else { ... }
...
parent.parent.DispatchEvent(targets, obj, { etype: 'node', nodeid: node._id, action: 'deviceShareUpdate', domain: domain.id, deviceShares: docs, nolog: 1 });
```

Every receiving session's `HandleEvent` clones the event (`common.Clone` = JSON round-trip, which turns the hole into a literal `null`) and then iterates:

```js
for (var i in event.deviceShares) { if (event.deviceShares[i].userid != user._id) { delete event.deviceShares[i].url; } }
```

→ `TypeError: Cannot read properties of null (reading 'userid')` on **every connected user session**, and sessions later in the dispatch chain never receive the event.

```
TypeError: Cannot read properties of null (reading 'userid')
    at ws.HandleEvent (meshuser.js:439:99)
    at CreateMeshCentralServer.obj.DispatchEvent (meshcentral.js:2482:49)
    at meshuser.js:4574:39
    at db.js:3089:57
```

*(line numbers from 1.1.58)*

**Note:** the sister dispatch sites (share removal/update paths, master ~L4305/L4388/L4460) already build a filtered `okDocs` before dispatching — only the `createDeviceShareLink` site ships the raw array.

**Suggested fix:** replace `delete docs[i];` with `docs.splice(i--, 1);` (or build `okDocs` like the sister sites). One-line change; happy to open a PR if you'd like.

On our deployment (a rental fleet that creates a time-limited share per rental, so an expired share is almost always present) this fired on virtually every share creation; patched locally with the splice and confirmed clean.

claytontavaresdan:
Fixes #8044 by adding a null guard to deviceShares in createDeviceShareLink.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,32 @@
+# Contributors
+
+We welcome contributions from the community! If you'd like to help improve MeshCentral, here's how:
+
+## How to Contribute
+
+1. ** Fork** the repository
+2. **Clone** your fork locally
+3. **Create a branch** for your fix/feature
+4. **Write tests** for your changes
+5. **Submit a Pull Request** to the upstream repository
+
+## Contributing Issues & Feature Requests
+
+Found a bug or have a feature idea? Open an issue on the [issue tracker](https://github.com/Ylianst/MeshCentral/issues).
+
+When reporting a bug, please include:
+- MeshCentral version
+- Operating system
+- Node.js version
+- Steps to reproduce
+- Expected vs actual behavior
+
+## Contact
+
+- **GitHub Issues:** [Ylianst/MeshCentral/issues](https://github.com/Ylianst/MeshCentral/issues)
+- **Community:** [meshcentral-community](https://github.com/meshcentral-community)
+
+### Active Contributors
+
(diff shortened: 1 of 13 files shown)


claytontavaresdan:
Closing: this issue is already fixed on current master. The createDeviceShareLink path now uses docs.splice(i--, 1) (meshuser.js ~L4679) and HandleEvent has a null guard (L484). The PR branch contained only unrelated batch-1 changes. No action needed upstream.

Issue #8059, opened 2026-08-05
LDAP user identity is mixed up after another user logs in


KBCACheia:
**Describe the bug**
Only the last digit of my objectSid on ldap server is different of another user. Then when i or him log in, the session is changed to the last one who logged in.

**To Reproduce**
1. Go to 'Using ldap auth'
2. Click on 'Have two objectSid with only the last digits diffent'
3. Log in user A and then user B
4. The user name/displayName change (meshcentral thinks it's the same user)

**Expected behavior**
They are differents users.

**Screenshots**

**Server Software (please complete the following information):**
 - Docker Image

**Additional context**
Changing ldapUserBinaryKey to sAMAccountName solve the problem (before was objectSid)

si458:
When u say docker image what image? What tag?

KBCACheia:
> When u say docker image what image? What tag?

Sorry: ghcr.io/ylianst/meshcentral:latest

si458:
@KBCACheia what LDAP server are you using?
can you share your config.json (like the bug report TELLS you to do!)

KBCACheia:
Its a Windows AD.

```
{
  "$schema": "https://raw.githubusercontent.com/Ylianst/MeshCentral/master/meshcentral-config-schema.json",
  "settings": {
    "plugins":{
      "enabled": false
    },
    "debug": "ldap",
    "cert": "",
    "_WANonly": true,
    "_LANonly": true,
    "_sessionKey": "MyReallySecretPassword1",
    "port": 443,
    "_aliasPort": 443,
    "redirPort": 80,
    "_redirAliasPort": 80,
    "AgentPong": 300,
    "tlsOffload": false,
    "trustedProxy": "",
    "SelfUpdate": false,
    "AllowFraming": false,
    "WebRTC": true
  },
  "domains": {
    "": {
      "auth": "ldap",
       "ldapOptions": {
         "url": "ldap://myserver:389/",
         "bindDN": "CN=user1,DC=local",
         "bindCredentials": "",
         "searchBase": "DC=local",
         "searchFilter": "(sAMAccountName={{username}})"
       },
      "ldapUserKey": null,
      "ldapUserBinaryKey": "sAMAccountName",
      "ldapUserName": "displayName",
      "ldapUserEmail": "mail",
      "ldapSiteAdminGroups": "CN=group1,DC=local",
      "ldapUserRequiredGroupMembership": ["CN=group1,DC=local"],
[...]

Issue #8049, opened 2026-07-30
Agent do not reconnect to local server when client is restarted. After Server-Restart Client is online again.


Cupiii:
**Describe the bug**

When a Windows 11 client is restarted, the MeshAgent does not reconnect to the MeshCentral server. The device remains offline in the MeshCentral web interface even though the agent service is running on the client.

The issue can be resolved temporarily by restarting the MeshCentral server. For example, selecting "Check for server updates" and choosing the current Stable version (which restarts the server) immediately causes the previously offline client to reconnect and appear online again.

Both the MeshCentral server and the client are located on the same local LAN.

**To Reproduce**

- Install MeshCentral on a Windows 11 machine.
- Install the MeshAgent on another Windows 11 machine in the same LAN.
- Verify that the client is connected and shown as Online.
- Restart the client computer.
- Wait several minutes for the MeshAgent to reconnect.
- Observe that the device remains Offline in MeshCentral.
- Restart the MeshCentral server (for example by selecting "Check for server updates" and choosing the current Stable version).
- Observe that the client immediately reconnects and is shown as Online.

**Expected behavior**

The MeshAgent should automatically reconnect to the MeshCentral server after the client has restarted, without requiring a restart of the MeshCentral server.

**Screenshots**

Not available.

**Server Software:**
OS: Windows 11
Virtualization: None (native installation)
Network: Local LAN
Version: 1.2.4
Node: ?

**Client Device:**
Device: PC
OS: Windows 11
Network: Local to MeshCentral on same switch (LAN)
Browser: Chrome
MeshCentralRouter Version: ?

**Remote Device:**
Device: PC
OS: Windows 11
Network: Local to MeshCentral (LAN)

**Additional context**

The problem appears consistently after every reboot of the client.

The MeshAgent service appears to be running after Windows starts, but the connection to the server is never re-established until the MeshCentral server itself is restarted.

Since restarting only the server immediately restores the connection, it seems that the server may not correctly detect or accept the client's reconnection after a reboot.

No reverse proxy, Cloudflare, or WAN access is involved. Also no session recording. The setup is a simple local LAN installation with two computers inviolved. Firewall is set to incoming for all apps, and IPs on TCP Ports 80, 433, 4433. 

**Your sanitized config.json file**

{

   "settings": {

      "lanonly": true,

      "nousers": true,

      "_minify": true

   },

   "domains": {

     

   }

}

Cupiii:
[...]

Issue #7969, opened 2026-07-07
[Feature Request] Granular Port Forwarding restrictions (Allow/Deny lists) via config.json and Web UI


fdefilippo:
Is your feature request related to a problem? Please describe.

Currently, the Port Forward permission in MeshCentral is binary: if an operator or a B2B user is granted port forwarding capabilities, they can map any TCP port (e.g., databases, internal web services, SMB) using MeshCentral Router or meshcmd. In strict bastion host environments, this poses a significant security risk, as it allows unauthorized lateral movement or data exfiltration, bypassing the tracking of the standard Web UI (Desktop/Terminal).
Describe the solution you'd like

We would like to propose a granular approach to restrict port forwarding, allowing administrators to define specific allowed or denied ports/port ranges. To make it easier to implement, we suggest splitting this feature into two phases:
PART 1: Backend Implementation via config.json (Phase 1)

In this first phase, administrators should be able to define global or domain-specific rules directly inside the config.json file.

We propose adding a portForwardRestrictions object under the domain settings. This object should support either an allow list or a deny list (but not both simultaneously to avoid conflicts):
```
"domains": {
  "": {
    "portForwardRestrictions": {
      "mode": "deny", 
      "ports": [3306, 445, "1433-1435"]
    }
  }
}
```
Or alternatively with an exclusive allowlist:
```
"portForwardRestrictions": {
  "mode": "allow",
  "ports": [22, 3389, 8080]
}
```
Expected Behavior:

    When meshcmd or MeshCentral Router attempts to initiate a RouteMap or RelayMap, the server checks the target remote port against these rules.

    If a port is denied, the server rejects the WebSocket tunnel handshake and returns an explicit error (e.g., Tunnel blocked by policy).

PART 2: Frontend Implementation via Web UI Dashboard (Phase 2)

Once the backend logic is capable of enforcing port restrictions, Phase 2 should introduce granular control inside the Web UI, moving the configuration from the global config.json down to the User / User Group permissions panel on specific Device Groups.

Proposed UI Changes:

    In the Device Group Permissions window, next to the "Port Forward" checkbox, add a "Configure" link or icon.

    Clicking it opens a modal allowing the admin to restrict ports specifically for that User/Group on that specific cluster of devices.

    The admin can type allowed/denied ports directly in the UI (e.g., 22, 3389).

Describe alternatives you've considered

[...]

PR #8076, opened 2026-08-12
fix: preserve distinct LDAP binary SID/GUID identity for userid (#8059)

claytontavaresdan:
## Issue
#8059 — LDAP user identity is mixed up after another user logs in.
Two users whose `objectSid` differ only in the last byte are treated as the same account (session switches to the last one who logged in).

## Root cause
`ldapjs` may serialize binary SID/GUID attributes to a string or a JSON-serialized Buffer shape `{ type: 'Buffer', data: [...] }` in `xxuser[...]` (this is what `ldapsaveusertofile` and the `test` LDAP config produce after a JSON round-trip). The old code did:

```js
shortname = Buffer.from(xxuser['objectSid'], 'binary').toString('hex').toLowerCase();
```

On older Node, `Buffer.from(object, 'binary')` coerces a plain object to the string `'[object Object]'`, so every user collapsed to the same shortname/userid. The result: two distinct binary SIDs map to one account.

## Fix
In `webserver.js` LDAP login path, introduce a `getBinaryKey(key)` helper that:
- Prefers the raw Buffer from `xxuser._raw[key]` (always the exact binary identity when `includeRaw: true`).
- Normalizes both real Buffers and the JSON-serialized `{ type:'Buffer', data }` shape before hex-encoding.
- Falls back to the string attribute only when no binary form is available.

This keeps the exact binary identity as the userid, so distinct SIDs stay distinct users.

## Test
`tests/test-8059-ldap-sid.js` — proves the OLD behavior collapses two distinct SIDs (`[object Object]` coercion) and the NEW behavior keeps them distinct, covering `_raw` Buffer, JSON-serialized Buffer, custom `ldapUserBinaryKey`, `objectGUID`, and name/cn fallback.

Contact: ctdan@protonmail.com

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,32 @@
+# Contributors
+
+We welcome contributions from the community! If you'd like to help improve MeshCentral, here's how:
+
+## How to Contribute
+
+1. ** Fork** the repository
+2. **Clone** your fork locally
+3. **Create a branch** for your fix/feature
+4. **Write tests** for your changes
+5. **Submit a Pull Request** to the upstream repository
+
+## Contributing Issues & Feature Requests
+
+Found a bug or have a feature idea? Open an issue on the [issue tracker](https://github.com/Ylianst/MeshCentral/issues).
+
+When reporting a bug, please include:
+- MeshCentral version
+- Operating system
+- Node.js version
+- Steps to reproduce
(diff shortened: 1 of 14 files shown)


claytontavaresdan:

Contact: ctdan@protonmail.com

claytontavaresdan:

PR #8077, opened 2026-08-12
fix: re-affirm connectivity when a new agent connection supersedes a stale duplicate (#8049)

claytontavaresdan:
## Issue
#8049 — Agent does not reconnect to local server when client is restarted. After a server restart the client reconnects immediately.

Both server and client on the same LAN. After rebooting the client, the device stays OFFLINE in MeshCentral until the **server** is restarted.

## Root cause
In  :
```js
const dupAgent = parent.wsagents[obj.dbNodeKey];
parent.wsagents[obj.dbNodeKey] = obj;
if (dupAgent) {
    // ... dupAgent.close(3);
} else {
    parent.parent.SetConnectivityState(...); // <-- only here!
}
```
When a duplicate (stale) agent connection is detected, the new agent is stored in `wsagents[dbNodeKey]` but `SetConnectivityState()` is **never** called for it.

Scenario: client reboots. The stale socket eventually dies and its `ws.on('close')` → `cleanup()` calls `ClearConnectivityState()`, marking the device OFFLINE. When the new agent connects, the server may still hold the stale entry as `dupAgent`, so the new connection lands in the duplicate branch and skips `SetConnectivityState` entirely. Result: the new agent IS connected, but the device shows OFFLINE. Restarting the server resets all state and the client reconnects *without* a duplicate, so the `else` branch runs and the device comes ONLINE — exactly what the reporter observes.

## Fix
Re-affirm `SetConnectivityState()` for the new agent inside the duplicate branch, so the active connection always owns the connectivity bit.

## Test
`tests/test-8049-agent-reconnect.js` — simulates the duplicate/supersede lifecycle and proves the OLD code leaves the device OFFLINE while the NEW code keeps it ONLINE.

Contact: ctdan@protonmail.com

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,32 @@
+# Contributors
+
+We welcome contributions from the community! If you'd like to help improve MeshCentral, here's how:
+
+## How to Contribute
+
+1. ** Fork** the repository
+2. **Clone** your fork locally
+3. **Create a branch** for your fix/feature
+4. **Write tests** for your changes
+5. **Submit a Pull Request** to the upstream repository
+
+## Contributing Issues & Feature Requests
+
+Found a bug or have a feature idea? Open an issue on the [issue tracker](https://github.com/Ylianst/MeshCentral/issues).
+
+When reporting a bug, please include:
+- MeshCentral version
+- Operating system
+- Node.js version
+- Steps to reproduce
+- Expected vs actual behavior
+
+## Contact
+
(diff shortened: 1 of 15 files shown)


claytontavaresdan:

Contact: ctdan@protonmail.com

claytontavaresdan:

si458:
@claytontavaresdan please can you fix this pr properly
you should create a pr and branch for each issue, 
you have included like 5 other issues into this pr

PR #8078, opened 2026-08-12
feat: granular port-forward restrictions via config.json (deny/allow lists) (#7969)

claytontavaresdan:
## Issue
#7969 — Granular Port Forwarding restrictions (Allow/Deny lists) via config.json and Web UI.

Port-forward permission is currently binary: any operator with the right can map ANY TCP port, allowing lateral movement in bastion-host environments. This implements **Part 1 (backend)** of the proposal.

## What this PR does
Adds a domain-level  policy enforced server-side when a port-forward relay cookie is requested (MeshCentral Router / meshcmd RouteMap/RelayMap go through ).

### Config (config.json)
```json
"domains": { "": { "portForwardRestrictions": { "mode": "deny", "ports": [3306, 445, "1433-1435"] } } }
```
or allow-list mode:
```json
"portForwardRestrictions": { "mode": "allow", "ports": [22, 3389, 8080] }
```
- `mode`: `deny` (block listed ports) or `allow` (only listed ports permitted). Not both.
- `ports`: array of numbers or `"start-end"` range strings.

### Enforcement
- `meshcentral.js`: `obj.isPortForwardAllowed(domain, port)` — the policy check.
- `meshuser.js`: `serverCommandGetCookie()` rejects the cookie with an explicit error (`Port N blocked by port-forward policy`) when the requested `tcpport` is blocked, so the tunnel handshake fails.
- `meshctrl.js`: registers `portForwardRestrictions` as a valid domain setting (settable via meshctrl config).

When no policy is configured, behavior is unchanged (all ports allowed).

## Part 2 (Web UI per user/group)
Left as a follow-up — once the backend enforces the policy, the UI can expose it in the Device Group Permissions panel next to the Port Forward checkbox.

## Test
`tests/test-7969-portforward-restrictions.js` — covers deny mode (single + range), allow mode, no policy, invalid mode (fail open).

Contact: ctdan@protonmail.com

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,32 @@
+# Contributors
+
+We welcome contributions from the community! If you'd like to help improve MeshCentral, here's how:
+
+## How to Contribute
+
+1. ** Fork** the repository
+2. **Clone** your fork locally
+3. **Create a branch** for your fix/feature
+4. **Write tests** for your changes
+5. **Submit a Pull Request** to the upstream repository
+
+## Contributing Issues & Feature Requests
+
+Found a bug or have a feature idea? Open an issue on the [issue tracker](https://github.com/Ylianst/MeshCentral/issues).
+
+When reporting a bug, please include:
+- MeshCentral version
+- Operating system
+- Node.js version
+- Steps to reproduce
+- Expected vs actual behavior
+
+## Contact
+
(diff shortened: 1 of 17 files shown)


claytontavaresdan:

Contact: ctdan@protonmail.com

claytontavaresdan:

Issue #6743, opened 2025-01-29
ListDevices returns output of command executed using RunCommand with --reply argument


kubamaruszczak:
**Describe the bug**
I have production EC2 instance running Meshcentral server with DocumentDB as its database. There are over 2500 linux agents connected to the server. I've created a bash script which allows me to execute command on each of the agents synchronously and to get the output back. It lists all of the devices using `ListDevices`, iterates over all of them and executes command on each connected agent using `RunCommand --reply`. See the screenshot below:

![Image](https://github.com/user-attachments/assets/5a55e9b9-1f86-4218-a408-4692720c4b05)

While the script is executing commands on the devices and I try to list device using `ListDevices` command, instead of the device I get output of the command executed on one of the devices like on the screenshot below:

![Image](https://github.com/user-attachments/assets/a6051610-81cf-4f63-a863-2a9609ebb980)

I can see is that this `RunCommand --reply` feature was added here -> https://github.com/Ylianst/MeshCentral/pull/5932 

**To Reproduce**
It's harder to reproduce in lower latency environment. However, I managed to do so with the below steps.

1. In one terminal window run infinite loop which will execute `echo hello` command on one device using `RunCommand` with `--reply` option
2. In a separate terminal window run infinite loop which will try to list that device using `ListDevices --filter <agent-name>` command
3. Observe the output of `ListDevices` command while `RunCommand` is being constantly executed. For some of the calls it will return 'hello' instead of the actual device info.

**Expected behavior**
I would like ListDevices command to return device instead of output from RunCommand --reply command.

kubamaruszczak:
I run my server with `trace=1` parameter and observe Console tab in my browser. What I see is that when I execute `RunCommand` using `meshctrl.js` I can see websocket response from RunCommand execution there as shown below. However, when I run ListDevices response doesn't appear in the Console.

![Image](https://github.com/user-attachments/assets/7d54e643-5221-4522-b9a3-c8ed1acb3d96)

My assumption is that the command response being returned from `meshcore.js` to the server creates some kind of conflict. I've just started learning how communication between Meshcentral components works so I might be wrong but it might be good catch at the same time. 

![Image](https://github.com/user-attachments/assets/c3fa2694-093e-4947-8410-d9b7de6ea61f)

kubamaruszczak:
[...]

PR #8080, opened 2026-08-12
fix(meshctrl): give runcommand its own responseid, stop --reply cross-talk on concurrent calls

pgorod:
## Summary

- `meshctrl.js` sends **every** action with the same hardcoded `responseid: 'meshctrl'`. The
  server broadcasts a `runcommands` reply to every currently-open control session under the same
  login, not only the one that made that specific request (root-caused in #6743). #6797 already
  fixed the case where a stray `runcommands` reply disrupted a *different* concurrent command
  (e.g. `ListDevices`) by adding a `data.type == 'runcommands' && settings.cmd != 'runcommand'`
  guard — but it doesn't cover two concurrent `RunCommand --reply` invocations disrupting **each
  other**, since both still satisfy `settings.cmd == 'runcommand'` and share the identical
  hardcoded `responseid`, so that guard can't tell them apart.
- Give `runcommand` its own per-process `responseid` (`RUNCOMMAND_RESPONSE_ID`, generated once via
  `crypto.randomUUID()` — `crypto` is already required at the top of the file) instead of reusing
  the shared `'meshctrl'` constant, and accept either value in the one reply handler that
  `runcommands` replies reach. Every other action keeps using the original shared `'meshctrl'`
  responseid, unchanged — this only touches the `runcommand` send site and its own reply check,
  so the ~40 other meshctrl commands are unaffected.

## Why this matters

`RunCommand --reply` was added in #5932 specifically so a command's output could be captured
programmatically. Running it across many devices — the natural use of `--reply` — means firing
off several concurrent invocations under the same login. Right now that silently cross-talks:
each process prints whichever concurrent request's reply happens to arrive on its socket first,
not necessarily its own. It's not an error and there's no warning, so a caller has no way to know
its device attribution is wrong without independently cross-checking every result.

## Reproduction (before this fix)

Two separate confirmations, to rule out anything specific to my own tooling:

1. `ForEach-Object -Parallel` in PowerShell 7, with and without `-UseNewRunspace` — identical
   wrong result either way, ruling out a PowerShell-side runspace artifact.
2. Fully independent OS processes via PowerShell `Start-Job` (separate process, separate login
   session per job, same account) — 3 and then 4 concurrent `RunCommand --reply` calls against
   different Windows devices behind `meshagent`, every process printed the same (fastest-to-reply)
   device's output.

## Testing

- Same reproduction steps above, re-run against this patch: 3, then 4, concurrent `RunCommand
  --reply` invocations (separate processes/sessions, same account) against different real devices
  — each now correctly returns its own device's output.
- `node --check meshctrl.js` — no syntax errors.
- Confirmed no regression on ordinary single-shot usage: `ListDevices` still returns the expected
  device list under the same patched file.

## How to try this yourself

Needs any 2+ online devices connected to a MeshCentral server (doesn't have to be my setup —
this is a login-token/session issue, not device- or platform-specific). Get their node IDs from
`ListDevices` first.

**1. Reproduce the bug on unpatched `meshctrl.js` (`master`):** open two terminals, and in each
run a `RunCommand --reply` against a *different* device, launched close enough together that
they're genuinely concurrent (a couple of `&`-backgrounded calls in one shell works too):

```bash
node meshctrl.js RunCommand --id 'DEVICE_A_ID' --run "hostname" --reply --url wss://YOUR_SERVER/control.ashx --loginuser YOU --loginpass YOUR_PASS &
node meshctrl.js RunCommand --id 'DEVICE_B_ID' --run "hostname" --reply --url wss://YOUR_SERVER/control.ashx --loginuser YOU --loginpass YOUR_PASS &
wait
```

Expect to see the **same** device's hostname printed for both calls at least some of the time —
it's timing-dependent, so if two calls don't show it, add a third or fourth concurrent call
against other devices; the more concurrent calls, the more reliably it reproduces (in my testing,
3-4 concurrent calls showed it on effectively every run).

**2. Apply this PR's diff, repeat the exact same test.** Each call should now consistently print
its own device's hostname, no matter how many run concurrently or how the timing lands.

**3. Confirm no regression on other commands** — e.g. `node meshctrl.js ListDevices ...` should
behave exactly as before, including while a concurrent `RunCommand --reply` is in flight (the
original #6743 scenario, still covered).

## Checklist

- [x] 🧠 I used LLMs/AI in this contribution and reviewed all generated content. I understand
      that I am responsible for and able to explain every line of code I submit.
      
- [x] 🔍 Any UI changes adhere to visual style of this project. *(no UI changes)*
- [ ] 🤖 I added or updated automated tests where appropriate. *(no existing test suite covers
      meshctrl.js's CLI paths, as far as I could tell — happy to add one if there's a preferred
      pattern to follow)*
- [x] 📄 Documentation updates are included (if applicable). *(none needed — no public API/CLI
      surface changed)*
- [x] 🧰 Dependency updates are listed and explained. *(none)*
- [ ] ⚠️ CI passes and is green. *(pending — will confirm once opened)*

Relates to #6743

### DIFF
--- a/meshctrl.js
+++ b/meshctrl.js
@@ -14,6 +14,7 @@ try { require('ws'); } catch (ex) { console.log('Missing module "ws", type "npm 
 
 var settings = {};
 const crypto = require('crypto');
+const RUNCOMMAND_RESPONSE_ID = crypto.randomUUID(); // unique per process, see PR description: avoids concurrent 'runcommand' invocations under the same login cross-talking on each other's --reply output
 const args = require('minimist')(process.argv.slice(2));
 const path = require('path');
 const possibleCommands = ['edituser', 'listusers', 'listusersessions', 'listdevicegroups', 'listdevices', 'listusersofdevicegroup', 'listevents', 'logintokens', 'serverinfo', 'serverversion', 'userinfo', 'adduser', 'removeuser', 'adddevicegroup', 'removedevicegroup', 'editdevicegroup', 'broadcast', 'showevents', 'addusertodevicegroup', 'removeuserfromdevicegroup', 'addusertodevice', 'removeuserfromdevice', 'sendinviteemail', 'generateinvitelink', 'config', 'movetodevicegroup', 'deviceinfo', 'removedevice', 'editdevice', 'addlocaldevice', 'addamtdevice', 'addusergroup', 'listusergroups', 'removeusergroup', 'runcommand', 'shell', 'upload', 'download', 'deviceopenurl', 'devicemessage', 'devicetoast', 'addtousergroup', 'removefromusergroup', 'removeallusersfromusergroup', 'devicesharing', 'devicepower', 'indexagenterrorlog', 'agentdownload', 'report', 'grouptoast', 'groupmessage', 'webrelay'];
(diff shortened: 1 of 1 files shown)


pgorod:
Hi everyone - the text above (and the patch) was made by Claude, as you might have suspected. I reviewed it carefully and I think it makes sense and solves a real bug. 

I stumbled across this when asking Claude to make a script to call each host in my fleet via meshctrl, in parallel. The responses were coming out all the same, and with this fix it's working great. Claude suddenly said "I'm making changes in meshctrl.js" to make it work, so I followed up with "Incidentally, is this something that would make sense to contribute to meshctrl repo as a PR?" and... here we are. 

Thanks for all the (brilliant) work done here in this amazing project 🎉

PR #8088, opened 2026-08-16
Multiple fixes in handlebars and validation of input

panreyes:
# Summary

- Dynamic values are encoded consistently before being displayed in the UI.
- Agent session and inventory data are validated before processing.
- File, messaging, plugin, and device dialogs handle unexpected input more reliably.
- Existing behavior and visual presentation are preserved.


meshagent.js
```diff
@@ -1446,7 +1446,7 @@ module.exports.CreateMeshAgent = function (parent, db, ws, req, args, domain) {
                                 }
                                 // Record keys actually read this scan (refreshes the timestamp).
                                 for (const v of Object.values(volumes)) {
-                                    if (v.identifier && v.recoveryPassword) { keys[v.identifier] = { rp: v.recoveryPassword, t: command.data.time }; }
+                                    if (v && v.identifier && v.recoveryPassword) { keys[v.identifier] = { rp: v.recoveryPassword, t: command.data.time }; }
```
si458, line 1449: this doesnt make sense because v is always defined because of the for loop ?
why would you need to check v is there and if it has data? 
just check if data is there

meshuser.js
```diff
@@ -6717,6 +6729,10 @@ module.exports.CreateMeshUser = function (parent, db, ws, req, args, domain, use
                     delete doc.domain;
                     delete doc._id;
 
+                    if (doc?.hardware?.windows?.volumes) {
+                        for (var i in doc.hardware.windows.volumes) { if (doc.hardware.windows.volumes[i] == null) { delete doc.hardware.windows.volumes[i]; } }
+                    }
```
si458, line 6734: why is this being deleted?

views/messenger.handlebars
```diff
@@ -95,8 +95,12 @@
         var notificationSupport = true;
 
         // File transfer state
+        var maxMessengerFileSize = 100 * 1024 * 1024;
+        var maxMessengerFileNameLength = 1024;
+        var maxMessengerFileChunkSize = 65536;
+        var maxMessengerFileDownloads = 10;
         var fileUploads = [];
-        var fileDownloads = {};
+        var fileDownloads = Object.create(null);
```
si458, line 103: why has this changed?

si458 (commented):
few questions

meshagent.js
panreyes: A rogue agent can include `null` within `hardware.windows.volumes`. The check prevents reading properties from that element and continues processing the remaining valid volumes.

meshuser.js
panreyes: We had a check to see if each volume is null, but we also need to check if the whole volumes array is null to prevent a rogue agent to block the rendering of the dashboard.

views/messenger.handlebars
panreyes: fileDownloads is used as a map, and uses methods certain methods that are not available as an array.
A certain malformed message can reach one of those methods before fileDownloads has any elements, forcing the server to restart.

panreyes:
Responded. Let me know if you have any more concerns. Thanks!

PR #8098, opened 2026-08-22
Pr 7947 reviewed and updated

PTR-inc:
As I couldn't push to the pr, a fork. Better anyway as it had some conflicts which are resolved now.

- Relates to #7947 

the edits of pr 7947 and removal of minimatch, now using node's path.matchesGlob(), as discussed the pr
Also fixed a restorebug, everything was restored to the meshcentral-data folder, now it uses the configured paths.
adds an option to force case-sensitivity for the globs.

### DIFF
--- a/backup.js
+++ b/backup.js
@@ -40,8 +40,38 @@ module.exports.zipDirectory = async function (dirPath, outputZip, password = '',
     }
 }
 
+// Move the folders zipExtract kept at the root of destPath to where they belong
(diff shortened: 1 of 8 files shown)


PTR-inc:
@si458 A good package.json strategy would make things easier. (and is an argument to use even less dependencies). Maybe something like you doing a separate pr after package.json changes with just that, the properly updated package(-lock) files?

PR #8100, opened 2026-08-23
Fix Linux URL opening from system agent sessions

erikkll:
## Summary

Fix Linux `openUrl` and desktop-file actions when MeshAgent runs as a system service and the active user is in a graphical Wayland session.

The service environment does not include the user's D-Bus session variables. `xdg-open` is already launched as the console user, but without that environment it cannot ask the desktop session to open the URL. This is observable with Mesh Messenger on GNOME Wayland, where `xdg-open` exits with code 3.

The change builds an environment for the active user using the existing `user-sessions` module:

- set the user's `HOME`;
- import `XDG_RUNTIME_DIR`, `DBUS_SESSION_BUS_ADDRESS`, `WAYLAND_DISPLAY`, and `DISPLAY` from the user's processes;
- use standard systemd user-session paths as a fallback for the runtime directory and D-Bus address.

The existing `xdg-open` call is retained.

## Validation

- `git diff --check`
- `node --check agents/meshcore.js`
- Manual Ubuntu 26.04 GNOME/Mutter Wayland test: `xdg-open` opens a URL in the existing Chrome session with the same user-session environment passed by this patch.

Related: Ylianst/MeshAgent#115

## Development note

This change and PR description were prepared with assistance from an AI coding agent. The contributor reviewed the result, confirmed the observed behavior, and authorized the submission.

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -4544,6 +4544,28 @@ function consoleHttpResponse(response) {
 }
 
 // Open a local file on current user's desktop
+function getLinuxDesktopEnvironment(uid) {
+    var env = {};
+    for (var i in process.env) { env[i] = process.env[i]; }
+
+    // Services do not inherit the graphical user's environment. Get it from
+    // the active user's processes so xdg-open can contact their desktop.
+    var sessions = require('user-sessions');
+    env.HOME = sessions.getHomeFolder(uid);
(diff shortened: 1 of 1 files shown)


erikkll:
This also fixes chat on ubuntu 26.04 (that's the primary reason i did this)

si458:
fixed in this pr instead #8103

PR #8104, opened 2026-08-23
Fix phantom click on mobile desktop connect

Rambomst:
# Summary

Connecting to a desktop from the mobile UI injected a stray left click into the remote session, which could land on whatever was under it (a browser Back button and the like).

The connect-time wake sent a synthetic mousedown+mouseup at the canvas centre, and the is-mobile GrabMouseInput override left the DOM mouse handlers attached, so a synthetic touch-compat click could fire too.

Wake streaming without injecting input instead: SendUnPause plus SendRefresh, retried over the first ~2.5s. If no frame has arrived after 3.5s, fall back to a single pointer move. Also drop the DOM mouse handlers on mobile so input comes only from the touch handlers.

### DIFF
--- a/views/default3.handlebars
+++ b/views/default3.handlebars
@@ -11087,6 +11087,14 @@
                         desktop.m.GrabMouseInput = function() {
                             _origGrab.call(desktop.m);
                             var c = desktop.m.CanvasId;
+                            // On mobile, input comes from the dtouch* handlers below; the mouse handlers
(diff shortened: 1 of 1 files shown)


si458:
does this happen on the classic ui? `default.handlebars` ?

Rambomst:
Nah, only on the new UI, it has logic specific to handling mobile touch while the older one doesn't.

PR #8124, opened 2026-08-29
Add configurable child server restart policy

panreyes:
# Summary

- Added the `RestartOnServerError` setting, enabled by default for backward compatibility.
- When disabled, MeshCentral does not restart the child process after a critical error and exits the parent process.
- Documented the setting in the configuration schema and advanced sample configuration.

<details>

- [x] 📄 Documentation updates are included (if applicable).
- [x] 🧰 Dependency updates are listed and explained.

</details>

### DIFF
--- a/meshcentral-config-schema.json
+++ b/meshcentral-config-schema.json
@@ -645,6 +645,11 @@
           "default": null,
           "description": "Execute this when the server has to restart due to an error. The first parameter will be the server version."
         },
+        "RestartOnServerError": {
+          "type": "boolean",
+          "default": true,
+          "description": "When false, the built-in supervisor exits after an unexpected child server error instead of restarting the child. Expected restarts for configuration changes, certificate updates, server updates and administrator requests are not affected. External service managers may start MeshCentral again according to their own restart policy."
+        },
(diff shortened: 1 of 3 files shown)


si458:
Erm this feature already exists.
`launch: 1` -> add to `settings`
It's just an undocumented value in the config.json

Edit. Launch can actually be any number

Edit. If u follow the source code meshcentral runs then spawns a new process with `--launch pid` 

So hense if u set `launch: (pid)` in settings then it runs as a single process, then if it crashes it crashes, it has no parent process to restart the app

panreyes:
I didn't know about launch, good to know something like that exists.

But this setting is a bit different, as this option prevents only restarting the server automatically when an unexpected error has occurred, and it still enables the other ways/types of restarting the server (updating certs, updating MeshCentral, server admin requests, etc...)..

si458:
@panreyes yes thats a side effect of `launch` when u run server update or it crashes or even using certurl and it has ro update the ssl and restart, it wont restart it just stops. 
The idea of this is to be used with systemctl or windows servuce which would handle the restarting for you and SHOULD auto reatart the service if crashes etc

PR #8125, opened 2026-08-29
Validate agent JSON inputs

panreyes:
# Summary

- Agent JSON input validation was added to `meshagent.js` before commands are routed, stored, or passed to plugin hooks.
- JSON actions such as `sysinfo`, `netinfo`, `coreinfo`, `msg`, `software`, and `plugin` are recursively checked for executable HTML attributes, script-capable elements, dangerous URL schemes, and related patterns that could be used for JavaScript injection.
- Empty top-level JSON objects and excessively nested requests are rejected, while legitimate protocol values such as nested empty objects and ordinary uses of words like `script`, `onload`, or `onerror` remain accepted.

<details>

</details>

### DIFF
--- a/meshagent.js
+++ b/meshagent.js
@@ -14,6 +14,89 @@
 /*jshint esversion: 6 */
 "use strict";
 
+const MAX_AGENT_INPUT_DEPTH = 32;
+
+// Validate JSON commands received from an agent. This is intentionally limited
+// to executable browser markup instead of rejecting words such as "script" or
+// "onerror" in ordinary inventory, log or console text. Empty nested objects and
+// arrays are valid agent values; only an empty top-level request is rejected.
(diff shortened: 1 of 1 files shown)


si458:
Hang on wont this break plugins as they use json?
Also what about runcommands? U can do some fun json with a powershell script?

panreyes:
This PR basically heavily prevents JavaScript injections that might have been sent from an agent.

It might block certain types of communications, like an agent sending HTML with JavaScript directly to the server for some reason through a plugin, but it's main point is to prevent the many XSS that we have been noticing lately.

Also, those fixes are universal: don't depend on a certain database, which will be quite useful.
It still requires testing, but I think the idea of filtering the agent input is pretty solid.

PR #8138, opened 2026-09-03
Add desktop privacy bar width option

### LINKED ISSUE #8130: Remote Toolbar width
Hi There,  is it possible that we can have options to decrease or increase the remote toolbar that is displayed on the client's desktop when you are in a remote session on their pc?

sometimes we have to go and open or close things on the remote client session and the toolbar is in the way

panreyes:
# Summary

- Added `desktopPrivacyBarMaxWidth` to limit the remote desktop privacy bar width.
- Preserved the existing automatic width when unset.
- Updated the configuration schema.

Resolves #8130

<details>
<summary>Checklist</summary>

- [x] 🧠 AI-assisted and reviewed.
- [x] 🛠️ Self-reviewed and tested.
- [x] 📄 Documentation updated.
- [x] 🧰 No dependency changes.
- [ ] ⚠️ CI passes.

</details>

### DIFF
--- a/agents/meshcore.js
+++ b/agents/meshcore.js
@@ -1335,6 +1335,7 @@ function handleServerCommand(data) {
                                 tunnel.consent = data.consent;
                                 if (global._MSH && _MSH().LocalConsent != null) { tunnel.consent |= parseInt(_MSH().LocalConsent); }
                                 tunnel.privacybartext = data.privacybartext ? data.privacybartext : currentTranslation['privacyBar'];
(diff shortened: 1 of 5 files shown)


si458:
What about agents running the 2022 version and don't want to update? (Winpe environments for example)
Will this break those setups? As the MaxWidth would be invalid/missing?

panreyes:
Thanks for catching that. I had thought about the MeshCentral server being older than the agent, but I forgot about the other possibility. It's fixed now.

PR #8141, opened 2026-09-03
Add option to disable recovery agent

panreyes:
# Summary

- Added the `disableRecoveryAgent` server setting to reject recovery agents with capability bit `0x40`.
- This opt-in measure hardens the server because recovery agents receive elevated authentication without registering as nodes, which may facilitate attacks in some situations.
- Documented the setting in the configuration schema and advanced configuration example.

<details>

- [x] 📄 Documentation updates are included (if applicable).
- [x] 🧰 Dependency updates are listed and explained.

</details>

### DIFF
--- a/meshagent.js
+++ b/meshagent.js
@@ -677,6 +677,9 @@ module.exports.CreateMeshAgent = function (parent, db, ws, req, args, domain) {
 
         // If this is a recovery agent
         if (obj.agentInfo.capabilities & 0x40) {
+            // Recovery agents are disabled by server policy.
+            if (args.disablerecoveryagent === true) { obj.close(1); return; }
+
             // Inform mesh agent that it's authenticated.
             delete obj.pendingCompleteAgentConnection;
             obj.authenticated = 2;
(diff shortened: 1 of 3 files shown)


si458:
This is a bad option.
The recovery mode/agent is needed because its the main core that's used when connecting an agent to meshcentral in the first place then it converts to meshcore.js.
I dont think I will accept this as it will cause madness!

panreyes:
It is an option, disabled by default, that helps MeshCentral servers to be a little more secure, as recovery agents are authenticated (level 2) blindly without registering as nodes, and it's an easy way for attackers to send code injections.

If my MeshAgents fail, I prefer to fix them manually than having my server not protected enough.

stephannn:
I don't know about the recovery option and how does it work, but I agree with with @panreyes about securing the server as an option. Talking our of my engineer perspective, having other tools around for app deployment and device management, I would also prefer secure the server and on the other hand having an own remediation for the agent. But can I check the agent health on the client itself by an own remediation script?

